PR #29, opened 2024-11-21
fix(#28): warning severity instead of error for abstract-decoratee lint

### LINKED ISSUE #28: The `abstract-decoratee` check must be warning but not error
Consider the code which is get after `unphi` operation:
```
# This is the default 64+ symbols comment in front of named abstract object.
[] > a
  # This is the default 64+ symbols comment in front of named abstract object.
  [] > @
    # This is the default 64+ symbols comment in front of named abstract object.
    [] > @
      # This is the default 64+ symbols comment in front of named abstract object.
      [] > @
        for. > @
          malloc.
            eolang.
              org.
                Q
          number.:0
            eolang.
              org.
                Q
            00-00-00-00-00-00-00-00:0
```

It's valid and it works but it fails with error: Abstract object can't be used as a decoratee.
I believe this check must be a warning so it's kind of bad practice to write a code in a such way
@yegor256 WDYT?

h1alexbel:
In this pull I've changed severity of `abstract-decoratee` lint from `error` to `warning`.

closes #28


h1alexbel:
@yegor256 take a look, please

h1alexbel:
@yegor256 take a look, please

PR #57, opened 2024-11-28
feat(#53): incorrect-rt-parts lint

### LINKED ISSUE #53: The `+rt` meta may have only one part or three parts, which is wrong
Currently, this is possible to do:

```
+rt hello
+rt a b c d e f
```

Both are wrong, since the `+rt` may only have exactly _two_ parts.

h1alexbel:
In this pull I've introduced new lint: `incorrect-rt-parts` that issues `critical` defect in case `+rt` meta doesn't have 2 parts inside.

closes #53
History:
- **feat(#53): incorrect-rt-parts**
- **feat(#53): motive**
- **feat(#53): more errors**


yegor256:
@h1alexbel thanks!

PR #89, opened 2024-12-03
feat(#47): incorrect-bytes-format lint

### LINKED ISSUE #47: the format of bytes inside XMIR objects is not validated
Currently, this XMIR will lead to no errors:

```
<program>
  <objects>
    <o name="foo">
      hello, world!
    </o>
  </objects>
</program>
```

However, this is not a valid XMIR. The text part of `<o>` must either be absent or match this regexp:

```
^--|[0-9A-F]{2}(-|(-[0-9A-F]{2})+)$
```

BTW, we technically can't catch this with XSD.

h1alexbel:
In this pull I've added new lint: `incorrect-bytes-format`, that issues `critical` defect in case of wrong format of the XMIR object's bytes.

closes #47
History:
- **feat(#47): incorrect-bytes-format**
- **feat(#47): bytes**
- **feat(#47): empty is ok**
- **feat(#47): test cases**
- **feat(#47): license**
- **feat(#47): motive**
- **chore: from diff branch**
- **chore: from diff branch**


yegor256:
@h1alexbel perfect, thanks!

PR #103, opened 2024-12-05
feat(#72): test-object-is-not-verb-in-singular lint

### LINKED ISSUE #72: first word of a test object must be a verb in singular
Here is a test:

```
+tests

# Foo.
[] > this-is-test
  42 > @
```

The name of the test object is wrong, because it doesn't start with a verb in plural form. A better alternative would be one of these names: `works-as-expected`, `finds-data`, `prints-report`, etc.

h1alexbel:
In this pull I've introduced new lint: `test-object-is-not-verb-in-singular`, that issues `warning` in case of test object name doesn't start with verb in singular form:

```
this-is-test, isItTrue, run-test -> warning
```

```
runs, prints-data, works-as-expected, parses-dom -> passes
```

closes #72
History:
- **feat(#72): select first**
- **feat(#72): test packs**
- **feat(#72): in java**
- **feat(#72): opennlp**
- **feat(#72): check tags**
- **feat(#72): add defect**
- **feat(#72): test cases**
- **feat(#72): yamls off**
- **feat(#72): up**
- **feat(#72): motive**
- **feat(#72): package-info**
- **feat(#72): puzzle**
- **feat(#72): URI**
- **feat(#72): clean for qulice**


h1alexbel:
@yegor256 take a look, please

h1alexbel:
@volodya-lombrozo reminder

h1alexbel:
@volodya-lombrozo reminder

src/test/java/org/eolang/lints/misc/TestObjectIsVerbInSingularTest.java
```diff
@@ -0,0 +1,105 @@
...
+final class TestObjectIsVerbInSingularTest {
+
+    /**
+     * Model.
+     */
+    private static POSTaggerME model;
+
+    @BeforeAll
+    static void setUp() throws IOException, URISyntaxException {
+        TestObjectIsVerbInSingularTest.model = new POSTaggerME(
```
volodya-lombrozo, line 61: @h1alexbel [NIT] You might use the default constructor `public TestObjectIsVerbInSingular()` here instead, am I right?

src/test/resources/org/eolang/lints/misc/test-object-is-not-verb-in-singular/bad/itIsTrue.eo
```diff
@@ -0,0 +1,27 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
++tests
```
volodya-lombrozo, line 23: @h1alexbel [Opinionated] You have a separate file for each test case. It's more or less fine, but due to the license text, it looks a bit verbose. Is it possible (worth it) to put all the test cases into a single file? Moreover, here you test only a few of examples. What about the following test cases:

```eo
# Tricky Test Names
+tests

# Foo.
[] > breaks-hearts
  42 > @

[] > it-works
  42 > @

[] > nothing-happened
  42 > @

[] > has-been-found
  42 > @

[] > something-is-wrong
  42 > @

[] > crashes-again
  42 > @

[] > will-fail-eventually
  42 > @

[] > always-returns-true
  42 > @

[] > was-lost-forever
  42 > @

[] > is-almost-correct
  42 > @

[] > nobody-knows-why
  42 > @

[] > should-not-pass
  42 > @

[] > once-upon-a-time
  42 > @

[] > forgets-everything
  42 > @

[] > must-do-better
  42 > @

[] > is-totally-broken
  42 > @

[] > was-a-trap
  42 > @

[] > hope-it-works
  42 > @

[] > dont-look-here
  42 > @

[] > never-saw-it-coming
  42 > @

[] > this-time-for-sure
  42 > @

[] > looks-fine-to-me
  42 > @

[] > there-it-goes
  42 > @

[] > it-is-fine-probably
  42 > @

[] > returns-something-strange
  42 > @

[] > maybe-next-time
  42 > @

[] > disappears-silently
  42 > @

[] > follows-the-rules
  42 > @

[] > well-this-is-awkward
  42 > @

[] > is-never-called
  42 > @

[] > why-this-again
  42 > @

[] > finds-nothing-at-all
  42 > @

[] > sounds-legit
  42 > @

[] > here-we-go-again
  42 > @

[] > it-was-working-before
  42 > @

[] > expected-the-unexpected
  42 > @

[] > it-seems-fine
  42 > @

[] > suddenly-works
  42 > @

[] > too-late-now
  42 > @

[] > could-not-care-less
  42 > @

[] > oops-try-again
  42 > @

[] > it-just-works
  42 > @

[] > dont-push-that-button
  42 > @

[] > error-404-not-found
  42 > @

[] > is-kind-of-slow
  42 > @

[] > sleeps-forever
  42 > @

[] > who-did-this
  42 > @

[] > makes-zero-sense
  42 > @

[] > please-reboot
  42 > @

[] > it-has-a-plan
  42 > @

[] > will-never-finish
  42 > @

[] > accidentally-passed
  42 > @

[] > runs-in-circles
  42 > @

[] > i-think-its-ok
  42 > @

```

src/test/java/org/eolang/lints/misc/TestObjectIsVerbInSingularTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+
+/**
+ * Tests for {@link TestObjectIsVerbInSingular}.
+ *
+ * @since 0.0.20
+ * @todo #72:90min Cache POS model after it's first load.
```
volodya-lombrozo, line 46: @h1alexbel We also might put the model right inside the plugin, it has 4MB size (which isn't so dramatically large.) What do you think? However, I'm not sure about license issues in this case.

volodya-lombrozo (changes requested):
@h1alexbel Thank you for the contribution! The PR looks pretty well. Just a few small suggestions.

src/test/java/org/eolang/lints/misc/TestObjectIsVerbInSingularTest.java
h1alexbel: @volodya-lombrozo agree

h1alexbel: @volodya-lombrozo agree

h1alexbel:
@volodya-lombrozo updated pull request. Take a look again, please

yegor256:
@h1alexbel it's not a good idea to keep binary files in Git repository, especially if these files are not ours. Instead, we can configure Maven to download the file during the build and place into the JAR.

volodya-lombrozo (commented):
@h1alexbel Let's [remove the binary](https://github.com/objectionary/lints/pull/103#issuecomment-2530570646) then and create a separate puzzle (as you did it before). Then we will decide what to do with it later. All the rest looks good to me. Thank you.

h1alexbel:
@volodya-lombrozo updated. Take a look again, please

volodya-lombrozo (approved):
@h1alexbel Looks good to me, thank you

h1alexbel:
@yegor256 could we merge this one?

src/main/java/org/eolang/lints/Program.java
```diff
@@ -46,14 +48,31 @@ public final class Program {
...
-            new XslLints(),
-            Arrays.asList(
-                new AsciiOnly()
-            )
-        )
-    );
+    private static Iterable<Lint<XML>> lints;
+
+    static {
+        try {
```
yegor256, line 54: @h1alexbel this is not a good idea in general to use static code blocks, because they are not testable at all: you can't create a unit test for this pice of code. A better design would be to create a new class `ProgramLints` and use it like this:

```
private static Iterable<Lint<XML>> lints = new ProgramLints().discover();
```

This is a perfectly testable design.

src/main/java/org/eolang/lints/misc/TestObjectIsVerbInSingular.java
```diff
@@ -0,0 +1,143 @@
...
+ * Lint that checks test object name is a verb in singular.
+ *
+ * @since 0.0.20
+ * @todo #72:60min Configure maven to download model file during the build and place into the JAR.
+ *  Currently, we download model file each time when creating the lint, which may
+ *  be slow in the usage of this lint. Instead, let's configure maven to download
+ *  model file during the build, and place into JAR, so lint will be able to locate
+ *  file from resources faster.
+ */
+@SuppressWarnings("PMD.TestClassWithoutTestCases")
```
yegor256, line 55: @h1alexbel better rename the class, to something like `UnitTestWithoutVerb`

yegor256: @h1alexbel or better make it exactly the same as the `check` of the `Defect` that you return: `UnitTestIsNotVerb` (`unit-test-is-not-verb`)

yegor256 (changes requested):
@h1alexbel looks good, see a few comments above

h1alexbel:

yegor256:
@rultor merge

PR #131, opened 2024-12-10
feat(#129): StanfordNLP in UnitTestIsNotVerb lint

### LINKED ISSUE #129: Migrate from OpenNLP to the latest version of similar lib with POS tags
Check this one: https://github.com/objectionary/lints/pull/126#issuecomment-2531215735

h1alexbel:
In this pull request I've replaced `opennlp` with `stanford-corenlp` library, and made `UnitTestIsNotVerb` lint stricter by allowing only `VBZ` (Verb, 3rd person singular present) verbs in the test names.

closes #129

### DIFF
--- a/src/main/java/org/eolang/lints/misc/UnitTestIsNotVerb.java
+++ b/src/main/java/org/eolang/lints/misc/UnitTestIsNotVerb.java
@@ -46,11 +46,6 @@ import org.eolang.lints.Severity;
  * Lint that checks test object name is a verb in singular.
  *
  * @since 0.0.22
- * @todo #72:60min Configure maven to download model file during the build and place into the JAR.
- *  Currently, we download model file each time when creating the lint, which may
- *  be slow in the usage of this lint. Instead, let's configure maven to download
- *  model file during the build, and place into JAR, so lint will be able to locate
- *  file from resources faster.
  */
 public final class UnitTestIsNotVerb implements Lint<XML> {
 
--- a/src/main/java/org/eolang/lints/misc/UnitTestIsNotVerb.java
+++ b/src/main/java/org/eolang/lints/misc/UnitTestIsNotVerb.java
@@ -44,8 +44,20 @@ import org.eolang.lints.Severity;
 
 /**
  * Lint that checks test object name is a verb in singular.
- *
+ * This lint uses <a href="https://stanfordnlp.github.io/CoreNLP/">Stanford CoreNLP model</a>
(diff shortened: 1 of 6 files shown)


h1alexbel:
@volodya-lombrozo please check this one

src/main/java/org/eolang/lints/misc/UnitTestIsNotVerb.java
```diff
@@ -60,60 +55,52 @@ public final class UnitTestIsNotVerb implements Lint<XML> {
...
-                        ).map(s -> s.toLowerCase(Locale.ROOT))
-                        .toArray(String[]::new)
+            final CoreDocument doc = new CoreDocument(
+                Stream
+                    .concat(
+                        Stream.of("It"),
+                        Arrays.stream(UnitTestIsNotVerb.KEBAB.split(name))
+                    ).map(s -> s.toLowerCase(Locale.ROOT))
+                    .collect(Collectors.joining(" "))
+            );
```
yegor256, line 97: @h1alexbel something is wrong with this formatting. Try this instead:

```
final CoreDocument doc = new CoreDocument(
  Stream
    .concat(
      Stream.of("It"),
      Arrays.stream(UnitTestIsNotVerb.KEBAB.split(name))
    )
    .map(s -> s.toLowerCase(Locale.ROOT))
    .collect(Collectors.joining(" "))
);
```

Make sure dots are vertically aligned. Also, read this: https://www.yegor256.com/2014/10/23/paired-brackets-notation.html

volodya-lombrozo (changes requested):
@h1alexbel looks good, to be honest. One more advantage over `opennlp` is that we don't need to manage model downloading and cache management. It's awesome. However, I would add a `javadoc` comment to the `UnitTestsIsNotVerb.java` to explain why we are using this specific model instead of `opennlp` - you could also include links to the issues that prompted this choice.

Also, there is a significant concern regarding the size of these models. While `opennlp` has a size ~ `4M`, the `stanford-corenlp-4.5.7-models.jar` is `452M` which might pose a problem in some cases. For now It is worth mentioning this in the docs.

h1alexbel:

volodya-lombrozo (approved):
@h1alexbel Looks good to me, thank you for the contribution!

h1alexbel:
@yegor256 take a look, please

yegor256:
@rultor merge

yegor256:
@h1alexbel thanks!

Issue #65, opened 2024-11-29
embed list of lints into the README file on every push to master


yegor256:
Let's create a GHA job that will, on every time we push to master, instantiate all lints and collect their "motive()"-s (`.md` files). Then, it will put all these Markdown files into the `gh-pages` branch, in the `motives` directory, and push. Then, it will modify the README file, inserting a list of all lints into it. Currently, we have a list there, but it's static. This one should be removed and a dynamically generated one should take its place.

Marat-Tim:
Let me try to do it

Marat-Tim:
@yegor256 Is the task relevant? Wouldn't it be better to add a test that verifies that for each lint from the `PkMono` and `PkWpa` classes there is a corresponding motive and that for each motive there is a lint?

Marat-Tim:
@yegor256 while working on [this](https://github.com/objectionary/lints/pull/505), I got the idea that we need to change the approach to providing a motive file and provide more than just a file.md(because I have to add to the documentation exactly which rules from markdownlint we ignore, what is stored in my variable). We need a template (for example, freemarker) to which variables can be passed. And then add a job that collects all the templates in a regular ci-cd pipeline, substitutes variables in them and checks that the resulting files are valid from the point of view of markdown. And add a job that, when pushed to master, dynamically collects files in the gh-pages branch

If we want to do this, then getting all the lines in the runtime makes sense

yegor256:
@Marat-Tim it's better to propose new ideas in separate tickets

yegor256:
@Marat-Tim yes, this task is relevant. We want to see a full list of lints in the README file. It's not about testing, it's about showing our users what we have in the library.

Marat-Tim:
@yegor256 We are already showing a list of what we have in the gh-pages branch. Do you mean that we need to add this list to README.md in the master branch?

yegor256:
@Marat-Tim yes, we want to see this list in the README too

Marat-Tim:
@yegor256 I'll clarify the requirements. We want the list of lints to be added to README.md in the `master` branch? Or we want to add README.md to the `gh-pages` branch. And which files should be referenced in this README? To html pages from the `gh-pages` branch or to some other ones?

yegor256:
@yegor256 we want to update the README in the master branch. We want to list _all_ lints that we have in the repo: they can be found by simply iterating `.md` motive files.

yegor256:
@Marat-Tim I agree, let's add a link

[...]

PR #137, opened 2024-12-11
feat(#65): get lints for gh-pages in runtime

Marat-Tim:
adds all motives of active lints to the Github pages branch

### DIFF
--- a/.github/motives.groovy
+++ b/.github/motives.groovy
@@ -1,0 +1,29 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints
+
+/**
+ * This script is run using gplus in the script motives.sh.
+ * This script saves a file with the lint name and extension .md to the tmp folder.
+ * The motive of this lint is preserved as content.
+ * These files are saved for all active lints in the project.
+ * @todo #65:15min after fixing this problem with creation of groovy folder
+ * https://github.com/jcabi/jcabi-parent/issues/586,
+ * move this script to the src/main/groovy folder in package org.eolang.lints
+ * @todo #65:15min after moving motives.groovy into src/main/groovy we should add test,
+ * that verifies that this script correctly saves a list of all lint motifs to the tmp folder
+ */
+
+static def saveLint(Lint<?> lint) {
+    new File("tmp/${lint.name()}.md").text = lint.motive()
+}
+
+new PkMono().forEach {
+    saveLint(it)
+}
+
+new PkWpa().forEach {
+    saveLint(it)
+}
--- a/pom.xml
+++ b/pom.xml
@@ -622,6 +622,23 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.gmavenplus</groupId>
+        <artifactId>gmavenplus-plugin</artifactId>
+        <version>4.1.1</version>
+        <configuration>
+          <scripts>
+            <script>./.github/motives.groovy</script>
+          </scripts>
+        </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>org.apache.groovy</groupId>
+            <artifactId>groovy</artifactId>
+            <version>4.0.26</version>
+          </dependency>
+        </dependencies>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/.github/motives.sh
+++ b/.github/motives.sh
@@ -41,12 +41,15 @@ tail() {
   printf '</article></section></body></html>'
 }
 
+mkdir tmp
+mvn clean compile gplus:execute
+
 while IFS= read -r f; do
   n=$(basename "${f}" .md)
   html=gh-pages/${tag}/${n}.html
   ( head && pandoc "${f}" && tail ) > "${html}"
   echo "${n} -> $(du -b "${html}" | cut -f1) bytes"
-done < <(find src/main/resources/org/eolang/motives -name '*.md' | sort)
+done < <(find tmp -name '*.md' | sort)
 
 (
   head
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ This Java package is a collection of "lints" (aka "checkers") for
 formatting. This is about best practices and readiness of code
 for successful compilation and execution.
 
+You can find the current full list of rules [here][lints-list].
+
 We use this package as a dependency in the
 [EO-to-Java compiler][EO]:
 
@@ -171,3 +173,4 @@ If you want the code to be checked using
 [XMIR]: https://news.eolang.org/2022-11-25-xmir-guide.html
 [EO]: https://www.eolang.org
 [benchmark-gha]: https://github.com/objectionary/lints/actions/runs/14839302036
+[lints-list]: https://www.objectionary.com/lints/
--- a/src/main/java/org/eolang/lints/LtByXsl.java
+++ b/src/main/java/org/eolang/lints/LtByXsl.java
@@ -11,7 +11,9 @@ import com.jcabi.xml.XML;
 import com.jcabi.xml.XMLDocument;
 import com.jcabi.xml.XSL;
 import com.jcabi.xml.XSLDocument;
+import io.github.secretx33.resourceresolver.DefaultResourceLoader;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Optional;
@@ -40,9 +42,9 @@ final class LtByXsl implements Lint<XML> {
     private final XSL sheet;
 
     /**
-     * Motive document.
+     * Path to motive document.
      */
-    private final Input doc;
+    private final String doc;
 
     /**
      * Ctor.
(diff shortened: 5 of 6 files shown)


Marat-Tim:
I understand that there are many errors here, but I wanted to do something so that you could specifically point them out to me

yegor256:
@Marat-Tim looks like a good start! Now, please, try to make all build jobs pass.

Marat-Tim:
@yegor256 I'm ready for my code to be destroyed)

I have a few questions

1. Do we collect the jar file somewhere and can we download it from there? Then from pom.xml it will be possible to remove the build plugin in jar

2. I assume that the PrintAllLines class needs to be moved to a separate project. If I create a project in the repository, won't it break some kind of pipeline? And again, how can I connect the latest version of the main project to subproject?

3. On behalf of whom to make a push request to the gh-pages branch? Now I've done it on my own login

4. How good is the idea of JavaLint? After all, if I wrap it at the wrong moment (for example, after wrapping it in Unlint), the name will be lost. Or you'll have to recursively get the decorated object, which sounds pretty crooked

Marat-Tim:
@yegor256 I updated and corrected the pull request. An example of how the gh-pages branch will be created with it can be viewed [here](https://github.com/Marat-Tim/lints/tree/gh-pages)

yegor256:
@Marat-Tim the `SaveAll.java` file is not in the right place. It's among feature files, which are shipped to the client in the final JAR. Should it be there? Maybe we can replace it with a Bash script?

yegor256:
@h1alexbel please, take a lookt

Marat-Tim:
@yegor256 We can't replace java code with bash if we want to get a list of motives in java runtime. I also tried to move this executable file to a separate subproject (similar to lints-it), but I ran into the problem that `PkWpa` and `PkMono` are not public classes, so the only way to use them is to write a file in the main project

Would it be better to just exclude this file from the build into the final jar? I just don't really understand which plugin this assembly is configured in, if you tell me, it will be very good for me.

pom.xml
```diff
@@ -243,7 +243,9 @@
             <configuration>
               <excludes>
                 <exclude>pmd:/src/it/.*</exclude>
+                <exclude>pmd:/src/main/java/org/eolang/lints/SaveAll.java</exclude>
```
h1alexbel, line 246: @Marat-Tim Why exclude this class from quality analysis?

src/main/java/org/eolang/lints/SaveAll.java
```diff
@@ -0,0 +1,75 @@
...
+import java.nio.file.Path;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.cactoos.iterable.Joined;
+
+public class SaveAll {
```
h1alexbel, line 18: @Marat-Tim This Java main class won't fit us. Ideally, we need some script that won't be shipped to the client, but can be executed during the build and forgotten. I thought about the following setup: 
1. Create script in Groovy language. with Java semantics support (we will get all the `Lint` interface implementations and call `name()` method there). For XSL lints names, we don't need Java, since we can iterate over `src/main/resources/org/eolang/lints/**.xsl` to collect them.
2. Run the script during the build with the help of [GMaven](https://groovy.github.io/GMavenPlus) plugin
3. Place the results to the `README.md` under `## Lints` section.

@yegor256 WDYT about Groovy script?

h1alexbel (changes requested):

src/main/java/org/eolang/lints/SaveAll.java
Marat-Tim: @h1alexbel we cannot get a list of all names by searching through folders with xsl files, since not all lints are implemented on xsl. In theory, we can go through the folder with markdown files, but then we need to write a test that checks that we don't have unnecessary markdown files without implementing the appropriate lint

What's the problem with keeping my class and excluding it from the final jar build?

pom.xml
Marat-Tim: @h1alexbel I thought, since this is a script, then we don't need to check it for quality

By analogy with classes in the `lints-it` subproject

src/main/java/org/eolang/lints/SaveAll.java
Marat-Tim: @h1alexbel @yegor256 Now, as far as I understand, the `motive` method is not used at all (at least its behavior is very strange, somewhere it returns the path to the file in jar, somewhere the text of the file from jar)

Is there no such thing that it should return a full-fledged markdown document (not the path to it)? In this approach, we will be able to dynamically create documentation using data from runtime

h1alexbel: @Marat-Tim I meant we can aggregate the list of lints this way:
* find XSL lints from FS
* find all Java lints

All of these manipulations can be listed in the Groovy script. Thus, you won't need to place this Java Utility class with our production classes, and we will keep [SoC](https://en.wikipedia.org/wiki/Separation_of_concerns) in the project.

Marat-Tim: @h1alexbel We can transfer this class to another package if we need to separate it from the production code

I just don't quite get it, how do we get a list of Java lints in groovy?

yegor256: @Marat-Tim @h1alexbel indeed, Groovy is a good option for such a script

.github/motives.groovy
```diff
@@ -0,0 +1,18 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
```
h1alexbel, line 5: @Marat-Tim let's remove this empty line

pom.xml
```diff
@@ -534,6 +534,23 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.gmavenplus</groupId>
+        <artifactId>gmavenplus-plugin</artifactId>
+        <version>4.1.1</version>
+        <configuration>
+          <scripts>
+            <script>./.github/motives.groovy</script>
```
h1alexbel, line 543: @Marat-Tim let's move the script into `src/main/groovy/org/eolang/lints/motives.groovy`

.github/motives.groovy
```diff
@@ -0,0 +1,18 @@
...
+package org.eolang.lints
+
+static def saveLint(Lint<?> lint) {
```
h1alexbel, line 8: @Marat-Tim let's add JavaDoc here, explaining why this script exists

h1alexbel, line 8: @Marat-Tim maybe we can setup tests for `saveLint()` method, in the `src/test/groovy`? Otherwise, we don't and _won't_ know if we break something during the future modifications. WDYT?

h1alexbel (changes requested):
@Marat-Tim check my comments, please

.github/motives.groovy
Marat-Tim: @h1alexbel when I tried to do this, I got an error https://github.com/jcabi/jcabi-parent/issues/586

I added a puzzle to this fix

Marat-Tim:
@h1alexbel

maxonfjvipon:
@h1alexbel please check this one

.github/motives.groovy
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints
+
+/**
+ * This script is run using gplus in the script motives.sh.
+ * This script saves a file with the lint name and extension .md to the tmp folder.
+ * The motive of this lint is preserved as content.
+ * These files are saved for all active lints in the project.
+ * @todo #65:30min after fixing this problem with creation of groovy folder
```
h1alexbel, line 12: @Marat-Tim let's make two todos for these issues:
* `"Move this script to the src/main/groovy folder."`
* `"Write a test for motives.groovy"`

h1alexbel (changes requested):
@Marat-Tim just one comment, please check

h1alexbel (approved):
@Marat-Tim LGTM

h1alexbel:
@yegor256 take a look, please

Marat-Tim:
@yegor256

yegor256 (changes requested):
a new version of jcabi-parent is available

Marat-Tim:
@yegor256 Yes, I'm working on it right now - but I've got a question: https://github.com/objectionary/lints/pull/137#discussion_r2085181413

pom.xml
h1alexbel: @Marat-Tim in the `jcabi-parent`, we have this `groovy` profile:

```xml
<profile>
    <!--
    Compile Groovy sources from src/main/groovy and src/test/groovy
    directories.
    -->
    <id>groovy</id>
    <activation>
      <file>
        <exists>${basedir}/src/main/groovy</exists>
      </file>
    </activation>
    <dependencies>
      <dependency>
        <groupId>org.codehaus.groovy</groupId>
        <artifactId>groovy-all</artifactId>
        <version>3.0.24</version>
        <type>pom</type>
      </dependency>
    </dependencies>
    <build>
      <plugins>
        <plugin>
          <groupId>org.codehaus.gmavenplus</groupId>
          <artifactId>gmavenplus-plugin</artifactId>
          <version>4.2.0</version>
          <configuration>
            <verbose>true</verbose>
          </configuration>
          <executions>
            <execution>
              <id>jcabi-groovy</id>
              <goals>
                <goal>compile</goal>
                <goal>compileTests</goal>
              </goals>
            </execution>
          </executions>
        </plugin>
      </plugins>
    </build>
  </profile>
```

h1alexbel: @Marat-Tim correct. Your Groovy scripts should be located in the `target/classes`. From there you can invoke them using `gmavenplus-plugin` passing path to the script:

```xml
<scripts>
  <script>src/main/groovy/org/eolang/lints/SaveMotives.groovy</script>
</scripts>
```

volodya-lombrozo:
Closing after over a year of inactivity and a merge conflict that hasn't been resolved. Feel free to rebase and reopen if you'd like to continue.

Issue #148, opened 2024-12-13
NPE in Program.defects()


yegor256:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Iterable.iterator()" because the return value of "org.cactoos.scalar.IoChecked.value()" is null
    at org.eolang.lints.Program.defects (Program.java:80)
```

Similar to #130

yegor256:
@volodya-lombrozo maybe you can help? I have no idea why this happens :((

volodya-lombrozo:
@yegor256 Sure, I'll investigate it. Can you assign it to me then, please?

yegor256:
@volodya-lombrozo assigned

yegor256:
@volodya-lombrozo here, for example: https://github.com/objectionary/eo/actions/runs/12321135063/job/34391895857?pr=3632

yegor256:
@volodya-lombrozo because of this bug we can't use lints 0.0.25 in EO

volodya-lombrozo:
> @volodya-lombrozo because of this bug we can't use lints 0.0.25 in EO. Try to clone eo repo, upgrade lints to 0.0.25 and build: you will get the error. I suspect the problem is related to the switch of java classloaders that is done (maybe!) by the NLP library.

@yegor256 Unfortunately, It's impossible to build `eo` now. I've done what you suggested: cloned `eo`, upgraded `lints` to `0.0.25` and run `mvn clean package`. After it the entire process stacked ~ 30 min, after which I had to just cancel the build (`Ctrl-C`). 

Also I checked the logs, you sent, they don't say anything useful:

```
Caused by: java.lang.NullPointerException
    at org.eolang.lints.Program.defects (Program.java:80)
    at org.eolang.maven.LintMojo.lint (LintMojo.java:266)
    at org.eolang.maven.LintMojo.lambda$lintOne$1 (LintMojo.java:142)
    at org.cactoos.scalar.ScalarOf.lambda$new$1 (ScalarOf.java:71)
    at org.cactoos.scalar.ScalarEnvelope.value (ScalarEnvelope.java:53)
    at org.cactoos.text.TextOfScalar.asString (TextOfScalar.java:58)
    at org.cactoos.text.TextEnvelope.asString (TextEnvelope.java:51)
    at org.cactoos.bytes.BytesOf.lambda$new$2 (BytesOf.java:233)
    at org.cactoos.bytes.BytesOf.asBytes (BytesOf.java:391)
[...]

PR #150, opened 2024-12-16
feat(#148): Fix Concurrent Problem With Lints Loading

volodya-lombrozo:
In this PR I demonstrate that the current implementation contains one significant concurrency bug.
This bug is related to the usage of the `Sticky` object from the `cactoos` library in a multi-threaded environment without synchronisation.
Therefore, in this PR I:
1. Added a `Lints` abstraction that encapsulates the logic of retrieving all lints (this allows us to test `Lints` independently from `Program`)
2. Added unit tests for `Lints` that demonstrate the problem
3. Fixed the original issue by using the `Synced` object from the `cactoos` library.

Closes: #148.


volodya-lombrozo:
@yegor256 Could you have a look, please?

yegor256:
@volodya-lombrozo isn't it too many classes? We already have `ProgramLints` and `XslLints`. Now, we will also have `Lints`... Can't we put this functionality into `ProgramLints`? Also, having `Lints` class visible publicly and not implementing any interfaces and coming into `Program` as an argument -- looks a bit wrong to me. WDYT?

volodya-lombrozo:
@yegor256 Sure, I agree. What do you think if we do it in a separate PR? I might add a puzzle for it

yegor256:
@volodya-lombrozo I'm not sure this will work for end-user:

```
public Program(final XML xmir, final Lints lints) {
```

The `Lints` class local, while the ctor is public. Maybe the ctor should be made package-private too?

yegor256:
@volodya-lombrozo I suggest making this PR right, instead of making a puzzle. Here, the problem is not cosmetic, it's design-level issue. Keeping it like this in the code base is dangerous.

volodya-lombrozo:
@yegor256 could you have a look one more time, please?

src/main/java/org/eolang/lints/Defect.java
```diff
@@ -32,12 +32,6 @@
  */
 public interface Defect {
 
-    /**
-     * Rule.
-     * @return Unique name of the rule
-     */
-    String rule();
```
yegor256, line 39: @volodya-lombrozo why? we use this method, in EO, in order to let users know which lint/rule failed

src/main/java/org/eolang/lints/LintProgram.java
```diff
@@ -0,0 +1,129 @@
...
+import org.cactoos.Scalar;
+import org.cactoos.iterable.Joined;
+import org.cactoos.scalar.IoChecked;
+import org.cactoos.scalar.Sticky;
+import org.cactoos.scalar.Synced;
+import org.eolang.lints.comments.AsciiOnly;
+import org.eolang.lints.misc.UnitTestIsNotVerb;
+
+/**
+ * Lints to use.
```
yegor256, line 42: @volodya-lombrozo it's a single entity, not a collection, right? Why the name is "lints"?

src/main/java/org/eolang/lints/Program.java
```diff
@@ -43,44 +39,47 @@
...
-        () -> new ProgramLints().value()
-    );
+    private final XML xmir;
 
     /**
-     * The XMIR program to analyze.
+     * The lints to use.
      */
-    private final XML xmir;
+    private final Lint<XML> lints;
```
yegor256, line 49: @volodya-lombrozo why `lints` if it's a singular `Lint`?

src/main/java/org/eolang/lints/Defect.java
volodya-lombrozo: @yegor256 As you can see, I was able to remove this method without any failed test. It indicates that we don't need this. Ok, I will return it back with a puzzle.

volodya-lombrozo:
> @volodya-lombrozo looks like you are mixing a few design approached into one. Maybe it's better to fix the problem at hands (the concurrency) issues and do the refactoring (if necessary) in another PR?

I suggested exactly the same [here](https://github.com/objectionary/lints/pull/150#issuecomment-2547489945):

> @yegor256 Sure, I agree. What do you think if we will do it in a separate PR? I might add a puzzle for it

Exactly by the same reasons you mentioned. So, originally it's your decision.

yegor256:
@volodya-lombrozo I just don't see a motivation for the refactoring of such kind. What's wrong with the current design?

volodya-lombrozo:
> @volodya-lombrozo isn't it too many classes? We already have `ProgramLints` and `XslLints`. Now, we will also have `Lints`... Can't we put this functionality into `ProgramLints`? Also, having `Lints` class visible publicly and not implementing any interfaces and coming into `Program` as an argument -- looks a bit wrong to me. WDYT?

@yegor256 It's your words. What I've done:

1. Combined `Lints` (originally created by me in this PR) and `ProgramLints` -> `ProgramLints`
2. Made `ProgramLints` implement `Lint<XML>` for consistency with the design 

That's all.

yegor256:
@volodya-lombrozo let's not do this for now. let's just fix the concurrency issue. design changes we can do in a separate PR. would be great if you submit a ticket for that!

volodya-lombrozo:
Requirement 1
> @volodya-lombrozo isn't it too many classes? We already have `ProgramLints` and `XslLints`. Now, we will also have `Lints`... Can't we put this functionality into `ProgramLints`? Also, having `Lints` class visible publicly and not implementing any interfaces and coming into `Program` as an argument -- looks a bit wrong to me. WDYT?

Requirement 2
> @volodya-lombrozo let's not do this for now. let's just fix the concurrency issue. design changes we can do in a separate PR. would be great if you submit a ticket for that!

@yegor256 How can I do this? There are conflicting requirements. I've already wrote some code to satisfy the **Requirement 1**.   Maybe we can fix the current version of code somehow? Or at least to leave a puzzle to fix current design problems if they are exist.

volodya-lombrozo:
> @volodya-lombrozo it was not a requirement

@yegor256 How that?

> @volodya-lombrozo I suggest making this PR right, instead of making a puzzle. Here, the problem is not cosmetic, it's design-level issue. Keeping it like this in the code base is dangerous.

Actually, you insisted on this. I don't see how I can satisfy conflicting "suggestions" :(

> I believe, you can do this with one line in `Program` file

I couldn't, because I had to test a separate part (originally named `Lints`) on concurrency issues. Please, see, the description of this PR. 

> Also, the tests will be helpful.

I added them from the first commit of this PR. Why are you writing about this?

> Just don't add new classes and don't delete existing one -- this is a requirement

Well, it's probably possible, but ineffective as I've already said:

> I couldn't, because I had to test a separate part (originally named `Lints`) on concurrency issues

____

Can we progress with these changes somehow? I ask you for understanding: you suggested one changes, then ask to return them back. Of course we can do it, but let's add a separate ticket for that, please?

yegor256:
@volodya-lombrozo check this out: 9b87673 Let's extend this block of text with your suggestions, it will help me understand what you are trying to achieve.

volodya-lombrozo:
@yegor256 [Opinionated] The following text seems redundant to me:

```
There are also a few classes that implement `Iterable<Lint>`.
They are supposed to be used only by the `Program` and `Programs`,
and are not supposed to be exposed to the user of the library.
They are responsible for providing a set of lints to be executed,
building them from the information in classpath.
```

These classes are internal components or implementation details, if you prefer. I don't think we should mention them.
Moreover, it seems more appropriate to hide them with `package-private` visibility because of the following statement:

> They are supposed to be used only by the `Program` and `Programs`, and are not supposed to be exposed to the user of the library.

Additionally, we might consider avoiding the use of `Iterable<Lint>` altogether. You may wonder how we could create multiple lints in that case. For this, it would be better to use a separate `X implements Lint<XML>` (not `X implements Iterable<Lint>`). We could call it `CompositeLint`, `LintComposite`, `SeveralLints`, or even `LintProgram` (as I have done in this PR). By doing so, you would simplify the overall solution and make the architecture easier to understand.

The closes analogy is `OptTrain implements Optimization` from `eo` repository that behaves as a single `Optimization`, but actually encapsulates a few of them. 

What do you think?

volodya-lombrozo:
@yegor256 Could you take a look one more time, please?

yegor256:
@rultor merge

yegor256:
@volodya-lombrozo thanks!

Issue #176, opened 2024-12-26
`NumberFormatException` in `empty-object`


yegor256:
```
Caused by: java.lang.NumberFormatException: For input string: "60.50.301"
    at java.lang.NumberFormatException.forInputString (NumberFormatException.java:67)
    at java.lang.Integer.parseInt (Integer.java:662)
    at java.lang.Integer.parseInt (Integer.java:778)
    at org.eolang.lints.LtByXsl.lineno (LtByXsl.java:144)
    at org.eolang.lints.LtByXsl.defects (LtByXsl.java:111)
    at org.eolang.lints.LtByXsl.defects (LtByXsl.java:45)
    at org.eolang.lints.Program.defects (Program.java:102)
```

also

```
IllegalStateException: Wrong line number reported by empty-object: '60.50.301'
    at org.eolang.lints.LtByXsl.lineno (LtByXsl.java:147)
    at org.eolang.lints.LtByXsl.defects (LtByXsl.java:111)
    at org.eolang.lints.LtByXsl.defects (LtByXsl.java:45)
    at org.eolang.lints.Program.defects (Program.java:102)
```

yegor256:
@h1alexbel maybe you can help? it's a blocker, since we can't use lints 0.0.27 in EO: https://github.com/objectionary/eo/pull/3765

h1alexbel:
@yegor256 yes assign me, please. What is the EO input that causes this `NumberFormatException`?

yegor256:
@h1alexbel I believe, this happens when a number of "empty objects" are present in a single XMIR file, but full story is here: https://github.com/objectionary/eo/actions/runs/12502887142/job/34882452877?pr=3765

h1alexbel:
@yegor256 can we release new version?

yegor256:
@rultor release, tag is `0.0.29`

PR #181, opened 2024-12-27
fix(#176): set line to '0' in `empty-object` if `@line` is weird

h1alexbel:
In this pull request I fixed line number reporting in `empty-object` lint, in cases when line number is NaN.

ref #176
History:
- **fix(#176): tests**
- **fix(#176): line 0 if NaN**


h1alexbel:
@maxonfjvipon please check

src/main/resources/org/eolang/lints/errors/empty-object.xsl
```diff
@@ -29,7 +29,7 @@ SOFTWARE.
       <xsl:for-each select="//o[not(o)]">
         <xsl:element name="defect">
           <xsl:attribute name="line">
-            <xsl:value-of select="if (@line) then @line else '0'"/>
+            <xsl:value-of select="if (@line and not(number(@line) = NaN) and number(@line) = number(@line)) then @line else '0'"/>
```
maxonfjvipon, line 32: @h1alexbel as far as I know `NaN = NaN `is `false`, so if you checked that line is not `NaN`, what this `line = line` extra comparison for?

h1alexbel:

maxonfjvipon (approved):

h1alexbel:
@yegor256 take a look, please

src/main/resources/org/eolang/lints/errors/empty-object.xsl
```diff
@@ -29,7 +29,7 @@ SOFTWARE.
       <xsl:for-each select="//o[not(o)]">
         <xsl:element name="defect">
           <xsl:attribute name="line">
-            <xsl:value-of select="if (@line) then @line else '0'"/>
+            <xsl:value-of select="if (@line and number(@line) = number(@line)) then @line else '0'"/>
```
yegor256, line 32: @h1alexbel why do we check it only in this XSL? If this problem is external, we should be careful with the `@line` attribute in all XSL stylesheets. Maybe let's introduce a function `eo:lineno(x)`, which will do what you do here, but used in all XSLs?

yegor256 (changes requested):
@h1alexbel thanks! see one comment above

h1alexbel:

yegor256:
@h1alexbel since you introduced this new function, maybe it's reasonable to use it in all XSLs? Also, no need to start the name of this XSL with underscore. It was done in EO repository in order to exclude this XSL from test pack.

h1alexbel:

yegor256 (approved):
@h1alexbel perfect!

PR #184, opened 2024-12-29
feat(#144): Each Lint timings in benchmark

### LINKED ISSUE #144: ProgramTest.java:59-64: Capture all the lint timings in...
The puzzle `134-6deff79b` from #134 has to be resolved: 

https://github.com/objectionary/lints/blob/ec352ec8a9402693173483008c87a0dbf2dd6028/src/test/java/org/eolang/lints/ProgramTest.java#L59-L64

The puzzle was created by @yegor256 on 13-Dec-24. 

Estimate: 90 minutes,  role: DEV. 

h1alexbel:
In this pull I've added `BcProgram` for measuring each lint timings for XMIR-generated program during the benchmark.

In the Benchmark results we will have the following:

```md
Input: com/sun/jna/Pointer.class
Size of .class: 22Kb (22Kb bytes)
Size of .xmir after disassemble: 1Mb (1Mb bytes, 29630 lines)
Lint time: 42s (42355 ms)

alias-too-long (112 ms)
alias-without-tail (11 ms)
broken-alias-first (522 ms)
broken-alias-second (11 ms)
duplicate-aliases (37 ms)
...
```

closes #144
History:
- **feat(#144): timings**
- **feat(#144): typo**
- **feat(#144): Lint#name()**
- **feat(#144): present timings.csv**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,11 +156,13 @@ SOFTWARE.
       <groupId>com.yegor256</groupId>
       <artifactId>tojos</artifactId>
       <version>0.18.4</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.opencsv</groupId>
       <artifactId>opencsv</artifactId>
       <version>5.9</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.yegor256</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,16 @@ SOFTWARE.
     </dependency>
     <dependency>
       <groupId>com.yegor256</groupId>
+      <artifactId>tojos</artifactId>
+      <version>0.18.4</version>
+    </dependency>
+    <dependency>
+      <groupId>com.opencsv</groupId>
+      <artifactId>opencsv</artifactId>
+      <version>5.9</version>
+    </dependency>
+    <dependency>
+      <groupId>com.yegor256</groupId>
       <artifactId>mktmp</artifactId>
       <version>0.0.5</version>
       <scope>test</scope>
--- a/src/test/java/org/eolang/lints/ProgramTest.java
+++ b/src/test/java/org/eolang/lints/ProgramTest.java
@@ -66,12 +66,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
  *  other classes in size too, for instance smaller classes (standard program),
  *  large class (JNA pointer), x-large class, and xxl class. Don't forget to
  *  adjust lint-summary.txt file to capture all the measurements.
- * @todo #134:90min Capture all the lint timings in timings.csv.
- *  Currently, we just capture total time - amount of milliseconds was required
- *  to lint some XMIR. Would be helpful to get extended statistics - timings of
- *  each lint. Don't forget to include this information into lint-summary.txt.
- *  You can check how it was implemented in HONE:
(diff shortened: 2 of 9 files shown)


h1alexbel:
@volodya-lombrozo could you review this one?

pom.xml
```diff
@@ -152,6 +152,16 @@ SOFTWARE.
       <version>0.49.4</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.yegor256</groupId>
+      <artifactId>tojos</artifactId>
+      <version>0.18.4</version>
+    </dependency>
+    <dependency>
+      <groupId>com.opencsv</groupId>
```
volodya-lombrozo, line 161: @h1alexbel Do we really need these libraries? At least I haven't found any usage of `com.opencsv`. Maybe I missing something?

src/test/java/org/eolang/lints/ProgramTest.java
```diff
@@ -250,4 +251,77 @@ void lintsLargeJnaClass(@Mktmp final Path home) throws Exception {
             }
         );
     }
+
+    /**
+     * Benchmarked program.
+     * @since 0.0.29
+     */
+    final class BcProgram {
```
volodya-lombrozo, line 259: @h1alexbel Can we make this class `private static`?  Or we exact this visibility level for `BcProgram`?

src/test/java/org/eolang/lints/ProgramTest.java
```diff
@@ -250,4 +251,77 @@ void lintsLargeJnaClass(@Mktmp final Path home) throws Exception {
...
+            this.xmir = program;
+            this.lints = lnts;
+            this.timings = tmngs;
+        }
+
+        /**
+         * Defects.
+         * @return Defects
+         */
+        public Collection<Defect> defects() {
```
volodya-lombrozo, line 308: @h1alexbel This is code duplication with the `Program` class. If we change `Program` implementation, `BcProgram` won't "see" these changes. It's better to rely on the existing `Program` implementation instead. I guess you can achieve this by implementing `LtMeasured`, or something similar. I'm not insisting on a concrete implementation, but it's better to remove this code duplication - at least in a separate issue. 
What do you think?

pom.xml
volodya-lombrozo: @h1alexbel Maybe, since you are using these libraries in tests only, it's better to add the `test` scope to them?

volodya-lombrozo (changes requested):
@h1alexbel Thank you for the contribution! Just several suggestions.

pom.xml
h1alexbel: @volodya-lombrozo `com.opencsv` is required for `MnCsv` class from `tojos`

src/test/java/org/eolang/lints/ProgramTest.java
h1alexbel: @volodya-lombrozo I'm not sure about the best way to do it for now. How about creating a new puzzle?

h1alexbel:

volodya-lombrozo (approved):
@h1alexbel Looks good to me, thank you!
@yegor256 Could you merge that changes, please?

yegor256:
@h1alexbel @volodya-lombrozo thanks!

PR #188, opened 2024-12-30
feat(#149): incorrect-alias lint

### LINKED ISSUE #149: incorrectly pointing `+alias` doesn't lead to any errors now, which is wrong
Consider this file:

```
+alias foo
+package ttt

[] > bar
  foo > @
```

Here, it is expected that the `ttt/foo.xmir` file exists. If it doesn't, we will have a runtime error when trying to access `foo`. This can easily be prevented in compile time. Let's just check that all files referred by `+alias` metas do exist.

h1alexbel:
In this pull I've introduced new lint: `incorrect-alias`, that issues critical defect in case `+alias` meta pointing to the non existing file.

closes #149
History:
- **feat(#149): incorrect-alias**
- **feat(#149): no empty**
- **feat(#149): passes if not aliases**
- **feat(#149): ignores absence of package**
- **feat(#149): motive**
- **feat(#149): passes if file is found**


h1alexbel:
@yegor256 please check

yegor256:
@h1alexbel looks like a great start! However, I believe that this lint must implement `Lint<Map<String, XML>>` (similar to how `LtUnitTestMissing` does). The presence of "files" must be checked in the provided `Map`, not on disc.

h1alexbel:

h1alexbel:

h1alexbel:
@yegor256 ready for review

src/main/java/org/eolang/lints/critical/LtIncorrectAlias.java
```diff
@@ -0,0 +1,93 @@
...
+        return "incorrect-alias";
+    }
+
+    @Override
+    public Collection<Defect> defects(final Map<String, XML> pkg) {
+        final Collection<Defect> defects = new LinkedList<>();
+        pkg.values().forEach(
+            xmir -> {
+                for (final XML alias : xmir.nodes("//meta[head='alias']/tail")) {
+                    if (!"1".equals(xmir.xpath("count(//meta[head='package'])").get(0))) {
```
yegor256, line 53: @h1alexbel You can do it simpler:

```
if (xmir.nodes("//meta[head='package']").size() != 1) {
```

src/main/java/org/eolang/lints/critical/LtIncorrectAlias.java
```diff
@@ -0,0 +1,93 @@
...
+    public String name() {
+        return "incorrect-alias";
+    }
+
+    @Override
+    public Collection<Defect> defects(final Map<String, XML> pkg) {
+        final Collection<Defect> defects = new LinkedList<>();
+        pkg.values().forEach(
+            xmir -> {
+                for (final XML alias : xmir.nodes("//meta[head='alias']/tail")) {
```
yegor256, line 52: @h1alexbel it's better to avoid `//` at the beginning of XPath expressions, because such requests are rather slow (we have to search the entire document). Better write this one like this: `/program/metas/meta[head='alias']/tail`

yegor256 (changes requested):
@h1alexbel looks very good now! A few small comments above.

h1alexbel:

yegor256:
@h1alexbel thanks!

PR #195, opened 2025-01-02
feat(#120): pos-bigger-than-line

### LINKED ISSUE #120: `@pos` may not be larger than the amount of characters in the line
Let's create a new lint that will check whether all `@pos` attributes are smaller than the length of the line to which they point.

h1alexbel:
In this pull I've introduced new lint: `pos-bigger-than-line`, that issues `critical` defect if `@pos` is bigger than `@line` attribute of the object in XMIR.

closes #120

### DIFF
--- a/src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
+++ b/src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
@@ -25,13 +25,12 @@ SOFTWARE.
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="pos-bigger-than-line" version="2.0">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:variable name="lines" select="tokenize(/program/listing, '&#10;')"/>
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="//o">
         <xsl:variable name="line" select="number(@line)"/>
-        <xsl:variable name="length" select="string-length(normalize-space($lines[$line]))"/>
-        <xsl:if test="@line and @pos&gt;$length">
+        <xsl:variable name="length" select="sum(//o[@line = $line and not(descendant::o)]/string-length(normalize-space(text())))"/>
+        <xsl:if test="@line and number(@pos)&gt;$length">
           <xsl:element name="defect">
             <xsl:attribute name="line">
               <xsl:value-of select="eo:lineno(@line)"/>
--- a/src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
+++ b/src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License (MIT)
+
+Copyright (c) 2016-2024 Objectionary.com
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="pos-bigger-than-line" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[(@pos and @line) and (number(@pos)&gt;number(@line))]">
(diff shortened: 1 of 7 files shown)


h1alexbel:
@maxonfjvipon please check

src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
```diff
@@ -0,0 +1,59 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="pos-bigger-than-line" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[(@pos and @line) and (number(@pos)&gt;number(@line))]">
```
maxonfjvipon, line 30: @h1alexbel please read the original ticket more carefully. `@pos` must be less that LEGTH of the line it belongs to. When we parse EO every object has its line and position at this line, for example here object `foo` would have position 6:
```
hello foo
```
But it can't have position 20 because the line length is 9

maxonfjvipon (changes requested):

src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
h1alexbel: @maxonfjvipon Ah, got it, thanks for the clarification, I will adjust this PR

h1alexbel:

src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
```diff
@@ -0,0 +1,64 @@
...
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="pos-bigger-than-line" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:variable name="lines" select="tokenize(/program/listing, '&#10;')"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o">
+        <xsl:variable name="line" select="number(@line)"/>
+        <xsl:variable name="length" select="string-length(normalize-space($lines[$line]))"/>
```
maxonfjvipon, line 33: @h1alexbel I'm not sure if it's a right choice to take length from listing. For example, in JEO repository we create XMIR from Java bytecode which would look like binary one line text in listing. Maybe it would be better to find all the `<o>` with the same line, count and sum the lengths of their `text()` and compare with `@pos`?
@yegor256 WDYT?

h1alexbel: @maxonfjvipon Agree, I think it will be more correct way to calculate the length of the `<o/>` line

maxonfjvipon: @h1alexbel be careful because there may be a performance issue since for every `<o>` element you need to find all other elements with the same `@line`

h1alexbel:
@maxonfjvipon updated. Take a look, please.

P.S. I'm not sure that is right way to calculate line length still

h1alexbel:
@maxonfjvipon reminder

h1alexbel:
@maxonfjvipon take a look, please

h1alexbel:
@maxonfjvipon reminder

src/test/resources/org/eolang/lints/packs/pos-bigger-than-line/passes-good-pos-in-line.yaml
```diff
@@ -0,0 +1,36 @@
...
+# SOFTWARE.
+---
+sheets:
+  - /org/eolang/lints/critical/pos-bigger-than-line.xsl
+asserts:
+  - /defects[count(defect[@severity='critical'])=0]
+document: |
+  <program>
+    <objects>
+      <o line="2" pos="3" name="foo">
```
maxonfjvipon, line 30: @h1alexbel this example in EO will look like this:
```
[] ([] ([]) > xyz) > foo
```
Yes, it's legal in EO, the only part which is not allowed is most deepest abstract object. It can't contain data and it cant' be nameless

src/test/resources/org/eolang/lints/packs/pos-bigger-than-line/catches-pos-bigger-than-line.yaml
```diff
@@ -0,0 +1,37 @@
...
+---
+sheets:
+  - /org/eolang/lints/critical/pos-bigger-than-line.xsl
+asserts:
+  - /defects[count(defect[@severity='critical'])=1]
+  - /defects/defect[@line='2']
+document: |
+  <program>
+    <objects>
+      <o line="2" pos="20" name="foo">
```
maxonfjvipon, line 31: @h1alexbel as I mentioned [here](), this example in EO will look like this:
```
[] ([] ([]) > xyz) > foo
```
It almost legal in your case, BUT this EO line has more that 20 symbols which means no error should be caught

src/main/resources/org/eolang/lints/critical/pos-bigger-than-line.xsl
```diff
@@ -0,0 +1,63 @@
...
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="pos-bigger-than-line" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o">
+        <xsl:variable name="line" select="number(@line)"/>
+        <xsl:variable name="length" select="sum(//o[@line = $line and not(descendant::o)]/string-length(normalize-space(text())))"/>
```
maxonfjvipon, line 32: @h1alexbel no, it still looks not correct to me.
Here's how I would try to implement it:
1. take EO code and convert it to XMIR. You'll get "real" XMIR, not a toy one which you made up
2. look at XMIR and EO code and try to make up with an algorithm of calculating line length

In general:
1. if you have `@name` - it EO it looks like this `> foo`, which is `foo` length + 2 extra symbols
2. if you have `@base` - it just `@base` length
3. if it's long horizontal application like this `x (a b c) (y 1 2) (* (o 2) (q w e))` you may see that that there are 10 braces and some amount of spaces without them, so they should be counted as well

Maybe there're some more patterns, I'm not sure, so the task is not such trivial as it seems to be

maxonfjvipon (changes requested):

h1alexbel:
@maxonfjvipon thanks for ideas, I will adjust this PR

volodya-lombrozo:
Closing this draft after nearly a year and a half of inactivity. Thanks for starting it - feel free to reopen and pick it back up anytime.

Issue #199, opened 2025-01-03
`empty-object.xsl` is super slow


yegor256:
According to the benchmark, `empty-object.xsl` is the slowest one in the list. Let's reduce it's complexity or optimize somehow.

volodya-lombrozo:
Profiling results related to `empty-object.xsl`: 
[flamegraph-disassemble-44918 copy.html.txt](https://github.com/user-attachments/files/18360730/flamegraph-disassemble-44918.copy.html.txt)

0pdd:
@yegor256 2 puzzles [#216](https://github.com/objectionary/lints/issues/216), [#217](https://github.com/objectionary/lints/issues/217) are still not solved.

yegor256:
@rultor release, tag is `0.0.31`

0pdd:
@yegor256 all 2 puzzles are solved here: [#216](https://github.com/objectionary/lints/issues/216), [#217](https://github.com/objectionary/lints/issues/217).

PR #210, opened 2025-01-09
feat(#199): Optimize `empty-object.xsl` Check

volodya-lombrozo:
I have written the test that identifies the problem with `empty-objects` check. Actually, it’s not a problem with the check. It’s a problem with `jcabi-xmi` library. I’ve optimised some parts of `LtByXsl` and added several puzzles to return the previous code back when these parts will be fixed.

Timings (for the new added test:)
- Before optimization: 2min 23sec
- After optimization: 6sec 535ms

Related to #199.


volodya-lombrozo:
@yegor256 Could you review these changes, please?

yegor256:
@volodya-lombrozo awesome, thanks!

Issue #209, opened 2025-01-09
Missed Description for `mandatory-home` Warning


volodya-lombrozo:
I'm getting the following warning when I check my `xmir`:

```
[mandatory-home WARNING]:0 The +home meta is mandatory, but is absent
```

I don't have `home` meta and I don't know what I should put here. Please add more human-readable description to this warning and provide instruction that should be done to fix this warning.

There is how I check the `xmir`:

```java
final Collection < Defect > defects = new Program(new StrictXmir(this.doc)).defects()
    .stream()
    .collect(Collectors.toList());
if (!defects.isEmpty()) {
    throw new IllegalStateException(
        String.format(
            "XMIR is incorrect: %s, %n%s%n",
            defects,
            this.doc
        )
    );
}
```

Full `xmir`:
[MethodByte.xmir.txt](https://github.com/user-attachments/files/18358754/MethodByte.xmir.txt)

volodya-lombrozo:
Blocks: https://github.com/objectionary/jeo-maven-plugin/issues/946

volodya-lombrozo:
@yegor256 Can you have a look, please?

yegor256:
@h1alexbel can you please help here and provide a better error message. I believe, we should say:

```
The +home meta is mandatory, but is absent (must contain URL of repository where object sources may be found)
```

yegor256:
@volodya-lombrozo in can of JEO, I would put there this URL (for all objects): https://github.com/objectionary/jeo-maven-plugin

volodya-lombrozo:
> @h1alexbel can you please help here and provide a better error message. I believe, we should say:
> 
> ```
> The +home meta is mandatory, but is absent (must contain URL of repository where object sources may be found)
> ```

@yegor256 @h1alexbel  It should also contain an example of this meta, otherwise, it's impossible to say which child elements the `home` meta contains.

@yegor256 Can you provide an example, please?

h1alexbel:
@yegor256 assign me, please

h1alexbel:
> It should also contain an example of this meta, otherwise, it's impossible to say which child elements the home meta contains.

@yegor256 @volodya-lombrozo I suggest we should implement [this](https://github.com/objectionary/lints/issues/207#issuecomment-2579710093), in order to resolve this type of problems. For instance, for `mandatory-home` we will link HTML page with this document: https://github.com/objectionary/lints/blob/gh-pages/_site/0.0.30/mandatory-home-meta.md. WDYT?

[...]

PR #214, opened 2025-01-09
bug(#209): better defect message in mandatory-home

h1alexbel:
In this pull I've updated defect message reported by `mandatory-home` lint, now it says about the need of home URL.

see: #209

PR #229, opened 2025-01-10
bug(#209): XMIR links in `Program`, `Programs`, `Defect` JavaDoc

h1alexbel:
In this pull I've updated JavaDocs of `Program.java`, `Programs.java`, `Defect.java`, now they mention XMIR, and resources about it.

see #209

### DIFF
--- a/src/main/java/org/eolang/lints/Defect.java
+++ b/src/main/java/org/eolang/lints/Defect.java
@@ -27,7 +27,14 @@ import com.jcabi.manifests.Manifests;
 
 /**
  * A single defect found.
- *
+ * Defect is a node in the XMIR under `/program/defects`, that describes the
+ * issue with EO/XMIR source code. Defect contain the message that addresses
+ * source code and points to the problem in it. Some defects report the problems
+ * on XMIR format itself, consider to check resources on XMIR in order to get
+ * understanding how intermediate representation of EO is structured in XML format.
+ * @see <a href="https://news.eolang.org/2022-11-25-xmir-guide.html">XMIR guide</a>
+ * @see <a href="https://www.eolang.org/XMIR.html">XMIR specification</a>
+ * @see <a href="https://www.eolang.org/XMIR.xsd">XMIR schema</a>
  * @since 0.0.1
  */
 public interface Defect {
--- a/src/main/java/org/eolang/lints/Program.java
+++ b/src/main/java/org/eolang/lints/Program.java
@@ -94,6 +94,9 @@ public final class Program {
     /**
      * Find defects possible defects in the XMIR file.
      * @return All defects found
+     * @see <a href="https://news.eolang.org/2022-11-25-xmir-guide.html">XMIR guide</a>
+     * @see <a href="https://www.eolang.org/XMIR.html">XMIR specification</a>
+     * @see <a href="https://www.eolang.org/XMIR.xsd">XMIR schema</a>
      */
     public Collection<Defect> defects() {
         try {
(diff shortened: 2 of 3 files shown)


h1alexbel:
@volodya-lombrozo please check

volodya-lombrozo (changes requested):
@h1alexbel Thank you for the contribution! Links are really helpful for developers of `objectionary/lints`. However, I should be honest with you: for users of `objectionary/lints` these links are useless. 

Let's imagine a user that sees the following error:

> [mandatory-home WARNING]:0 The +home meta is mandatory, but is absent

In this case, the user will rise one more issue in this repository or disable the linter completely. Checking code in this repository is the last thing user will do.

We need to show the error message together with the link to the documentation, otherwise it won't work.

h1alexbel:
@volodya-lombrozo how about this: https://github.com/objectionary/lints/pull/214#issuecomment-2582356926?

volodya-lombrozo:
> @volodya-lombrozo how about this: [#214 (comment)](https://github.com/objectionary/lints/pull/214#issuecomment-2582356926)?

@h1alexbel I don't like the idea of "searching" something. I would like to see the link for each linter in a row:

```
1. Cryptic error [name-outside-of-abstract-object] How to fix: URL
2. Cryptic error [mandatory-home] How to fix: URL

Lints list is here: URL
XMIR docs is here: URL
```

"But what you have suggested is already much better than what we have now.

yegor256:
@volodya-lombrozo @h1alexbel we have two types of users:

1) writers of EO
2) users of API

The first type of users will use eo-maven-plugin (or eoc) and will see the output in text form in command line. They will see the links to "motive" pages. Now they don't, but here is the ticket for this: https://github.com/objectionary/eo/issues/3800

The second type of users will use `Program` class from this library and will receive a list of `Defect` instances. We don't want to put URLs into `defect.text()` methods, because this is not where they belong. We should provide proper API documentation to API users, but _not in runtime_.

h1alexbel:
@yegor256 @volodya-lombrozo for the second group, for the "users of API", I created this PR. I suggest to merge it, if its good enough. For first group we will implement https://github.com/objectionary/eo/issues/3800. But, as far I understand, JEO uses only lints API, so @volodya-lombrozo won't get any motives on his side, the motives only will come from `eo-maven-plugin`, and `eoc`  right? WDYT?

yegor256:
@h1alexbel yes, users of API (@volodya-lombrozo is one of them) only get a list of `Defect` objects. They (users) have to do some research on their own in order to understand what each defect means. The best we can do for them is provide proper Javadoc texts.

src/main/java/org/eolang/lints/Defect.java
```diff
@@ -27,7 +27,14 @@
 
 /**
  * A single defect found.
- *
+ * Defect is a node in the XMIR under `/program/defects`, that describes the
+ * issue with EO/XMIR source code. Defect contain the message that addresses
+ * source code and points to the problem in it. Some defects report the problems
+ * on XMIR format itself, consider to check resources on XMIR in order to get
+ * understanding how intermediate representation of EO is structured in XML format.
+ * @see <a href="https://news.eolang.org/2022-11-25-xmir-guide.html">XMIR guide</a>
```
yegor256, line 35: @h1alexbel read this: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

also, run `mvn site -Psite` and then open the files generated in `target/site/javadoc/index.html` You will see how your text is formatted.

h1alexbel: @yegor256 getting this:
image

should be good I believe, or I missing something?

yegor256:
@h1alexbel thanks!

Issue #237, opened 2025-01-14
`incorrect-version` Check Throws False Negative Error for `1.0-SNAPSHOT`


volodya-lombrozo:
XMIR might be generated from any source (not only from EO), by this reason sometimes we can generate XMIR by using `jeo-maven-plugin:1.0-SNAPSHOT`. Pay attention to the version `1.0-SNAPSHOT` - it's correct  semver version. 

Semver BNF (part):

```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>
```

You can read more about it here: https://semver.org
 
But for the version `1.0-SNAPSHOT` I'm getting the following error:

```
XMIR is incorrect: [[incorrect-version WARNING]:0 The format of the +version meta is wrong: "1.0-SNAPSHOT" (SemVer expected instead)], 
```

Please, remove this check or fix it.

volodya-lombrozo:
@yegor256 Pay attention, please. It blocks: https://github.com/objectionary/jeo-maven-plugin/pull/962

volodya-lombrozo:
Well, with `0.1.0-SNAPSHOT` version it works fine. So, actually `1.0-SNAPSHOT` doesn't follow semver according with BNF I send before. My bad. I need to change the version in my project.

However, I have some doubts about it. `1.0-SNAPSHOT` looks correct to me, although it doesn't follow semver. 

@yegor256 What do you think if we allow such versions?

yegor256:
@volodya-lombrozo I agree, we should allow `x.x-SNAPSHOT` (it's [Maven convention](https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN8855))

yegor256:
@h1alexbel can you help please?

yegor256:
@rultor release, tag is `0.0.32`

volodya-lombrozo:
@h1alexbel Thank you for the fast fix! Works like a charm.

PR #238, opened 2025-01-14
bug(#237): incorrect-version allows maven convention

h1alexbel:
In this PR I've updated `incorrect-version` lint to allow maven convention, e.g.: `1.0-SNAPSHOT` if specified in `+version` meta.

see #237


yegor256:
@h1alexbel try to test `1.1-hello` (should raise an error). Only `x.x-SNAPSHOT` must be allowed.

yegor256:
@h1alexbel thanks!

PR #275, opened 2025-01-23
bug(#257): Apache OpenNLP instead of Stanford Core NLP

### LINKED ISSUE #257: edu.stanford.nlp:stanford-corenlp is GPL
The dependency that we use (`edu.stanford.nlp:stanford-corenlp`) is licensed as GPL. It's a restrictive license. Our product is MIT, which is permissive. That's an obvious violation. Let's replace them with something else.

h1alexbel:
In this PR I've replaced [Stanford CoreNLP](https://stanfordnlp.github.io/CoreNLP/) (GPL) with [Apache OpenNLP](https://opennlp.apache.org/) (Apache-2.0) in order to comply with dependency license.

closes #257

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,17 +109,6 @@ SOFTWARE.
       <!-- version from the parent pom -->
     </dependency>
     <dependency>
-      <groupId>edu.stanford.nlp</groupId>
-      <artifactId>stanford-corenlp</artifactId>
-      <version>4.5.8</version>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


h1alexbel:
@yegor256 just found that OpenNLP 2.4.0 supports only Java 18+, should we downgrade this dep, for now?

yegor256:
@h1alexbel yes, it's better to downgrade

yegor256:
@h1alexbel thanks!

Issue #266, opened 2025-01-22
`sparse-decoration` complains too much


yegor256:
A warning is raised for this code:

```
[] > test
  if. > @
    a
    42
    88
```

It looks perfectly legit to me. The `sparse-decoration` must only complain about this situation:

```
[] > test
  [] > @
    if. > @
      a
      42
      88
```

h1alexbel:
@yegor256 as far I understood, sparse decoration is allowed in application, for instance:

```eo
[] > decorates-application
  if > @
    true
    5
    five
```

should not raise any warnings, right?

yegor256:
@h1alexbel correct

maxonfjvipon:
@yegor256 your example is actually a sparse-decoration. Should be:
```
if. > test
  a
  42
  88
```

maxonfjvipon:
@yegor256 check [this](https://github.com/objectionary/eo/issues/2109#issuecomment-1570416035)

maxonfjvipon:
@yegor256 WDYT?

yegor256:
@maxonfjvipon I agree, my mistake, thanks!

PR #285, opened 2025-01-30
bug(#266): correct `sparse-decoration` detection

h1alexbel:
In this PR, I've updated `sparse-decoration.xsl` so that it doesn't issue defects in the object application.

see #266
History:
- **bug(#266): tests**
- **bug(#266): allows in application**


h1alexbel:
@yegor256 please check

yegor256:
@maxonfjvipon please, check this one

maxonfjvipon:
@h1alexbel I suggest we put this PR on hold until we resolve [this](https://github.com/objectionary/lints/pull/252). There are some breaking changes which may affect this PR

h1alexbel:
@maxonfjvipon synced PR with EO `0.51.1`. Take a look, please

src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
```diff
@@ -27,16 +27,18 @@ SOFTWARE.
...
-          <xsl:attribute name="line">
-            <xsl:value-of select="eo:lineno(@line)"/>
-          </xsl:attribute>
-          <xsl:attribute name="severity">
-            <xsl:text>warning</xsl:text>
-          </xsl:attribute>
-          <xsl:text>Sparse decoration is prohibited</xsl:text>
-        </xsl:element>
+      <xsl:for-each select="//o">
+        <xsl:if test="(count(descendant::o[not(@base='Q.org.eolang.bytes')])=1 and o[@name='@'] and (not(@base) or @base!='^') and not(o[@base='∅'])) or count(o[@name='@' and o[@name='@']])&gt;0">
```
maxonfjvipon, line 31: @h1alexbel the idea of sparse decoration is simple - don't use abstract objects if you can.
Consider the example:
```
[] > foo
  5 > @
```
Here we can get rid of abstract object:
```
5 > foo
```

The same story with:
```
[] > foo
  x y z > @
```
It'll be
```
x y z > foo
```

But here we can't just get rid of it because of void attribute:
```
[x] > foo
  5 > @
```
The same story here:
```
[] > foo
  5 > @
  4 > x
```

So the main idea - if abstract object has only one attribute and it has name `@` - it's sparse decoration. 

Considering this let's simplify the whole condition with something like this:
```xpath
//o[eo:abstract(.) and count(o)=1 and o[@name='@']]
```

I believe it should cover all the possible cases

maxonfjvipon (changes requested):

src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
h1alexbel: @maxonfjvipon How about these cases:

```eo
# No comments.
[] > test
  # No comments.
  [] > @
    if. > @
      a
      42
      88
```

I think, `sparse-decoration` should warn about `test` only here (`@` should be good).

And this:

```eo
# No comments.
[] > test
  if. > @
    a
    42
    88
```

Here, no warnings should be issued. If this is correct, I propose the following XPath for detecting sparse decorations:

```text
//o[(count(descendant::o[not(@base='Q.org.eolang.bytes')])=1 and o[@name='@'] and (not(@base) or @base!='^') and not(o[@base='∅'])) or count(o[@name='@' and o[@name='@']])&gt;0]
```

This will handle cases above. WDYT?

maxonfjvipon: @h1alexbel check [this](https://github.com/objectionary/lints/issues/266#issuecomment-2627731439), let's wait for the answer

maxonfjvipon: @h1alexbel here's [more](https://github.com/objectionary/eo/issues/2109#issuecomment-1570416035) about sparse-decoration

maxonfjvipon: @h1alexbel let's stick with this one: //o[eo:abstract(.) and count(o)=1 and o[1][@name='@']]

h1alexbel:

maxonfjvipon (approved):

h1alexbel:
@yegor256 take a look, please

Issue #283, opened 2025-01-29
`+unlint` may use invalid name of lint


yegor256:
Let's say, this is the code:

```
+unlint foo

[] > bar
  42 > @
```

It's not legit, since there is not lint by the name `foo`. Let's create a lint that will catch such mistakes.

Marat-Tim:
Let me do this

yegor256:
@Marat-Tim sure, go ahead please

Marat-Tim:
If I understand correctly then not all lints can be unlint right now. For example, `LtTestNotVerb` and `LtTestNotVerb` are not wrapped by LtUnlint. Will this be fixed or may we have lints that cannot be ignored?

yegor256:
@Marat-Tim they are wrapped by `LtWpaUnlint`. In general, when you have a question about our code base, when something is not clear about it, don't ask -- submit a ticket, blaming the code base for the absence of documentation. Read this: https://www.yegor256.com/2024/12/15/open-source-beginner-advice.html

Marat-Tim:
I think you're wrong, I don't see where the classes I mentioned wrap in LtWpaUnlint

![Image](https://github.com/user-attachments/assets/1803953f-3837-4718-a78d-2893b9f79fd5)

So, I will create a new issue about this

yegor256:
@Marat-Tim yes, good point, make a new issue please

h1alexbel:
@yegor256 this one should be solved

PR #302, opened 2025-02-02
feat(#283): checking for incorrect use `unlint`

Marat-Tim:
In this pull request, I will add a new lint that verifies that all unlint instructions reference existing lints

To do this, a collection of all existing names is collected using Lint#name()


Marat-Tim:
@yegor256 please look, I added a lint that checks that all unlints point to an existing lint. The logic in it is quite simple, but it seems that implementing it through xsl will be much more difficult than through java code. Perhaps it’s worth somehow changing the approach to getting all lints, since if other lint collections appear in the future, they will have to be added to the AllLints class, but I don’t know how exactly this can be fixed

src/main/java/org/eolang/lints/errors/LtIncorrectUnlint.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public final class LtIncorrectUnlint implements Lint<XML> {
+    @Override
+    public String name() {
+        return "incorrect-unlint";
+    }
+
+    @Override
+    public Collection<Defect> defects(final XML entity) throws IOException {
+        final Set<String> alllints = new SetOf<>(
```
maxonfjvipon, line 55: @Marat-Tim let's make this variable a static in this class and move the code from `AllLints` to here. There's no point to have a separated class `AllLints` right now since it's used only in one place. Let's give a name `ALL` to this static variable. Also let's wrap it with some `Sticky` decorator since the variable is used in cycle for many times

src/main/java/org/eolang/lints/errors/LtIncorrectUnlint.java
```diff
@@ -0,0 +1,88 @@
...
+import org.cactoos.text.UncheckedText;
+import org.eolang.lints.AllLints;
+import org.eolang.lints.Defect;
+import org.eolang.lints.Lint;
+import org.eolang.lints.Severity;
+
+/**
+ * Lint that all unlint metas point to existing lint.
+ *
+ * @since 0.0.0
```
maxonfjvipon, line 45: @Marat-Tim it's a good practice to use the next minor version of the product here, so let's update this semver here and in other places

maxonfjvipon (changes requested):
@Marat-Tim in general it looks so well, just a few comments to be fixed and we're good to merge

yegor256:
@Marat-Tim also, would be great if you can name your pull request better and provide a description for it, check this one please: https://www.yegor256.com/2024/12/15/open-source-beginner-advice.html

src/main/java/org/eolang/lints/errors/LtIncorrectUnlint.java
Marat-Tim: why is Sticky needed here? We're collecting a set of regular strings

maxonfjvipon: @Marat-Tim because objects are lazy and used in loop. Every time you do `ALL.contains(...)` all the nested objects start their work: map, join, exclude duplicates and so on

Marat-Tim: I think I got it, please see if I used Sticky correctly

maxonfjvipon (approved):
@Marat-Tim LGTM!
@yegor256 WDYT?

maxonfjvipon:
@yegor256 please check

yegor256:
@Marat-Tim thanks!

Issue #276, opened 2025-01-23
LtTestNotVerb.java:52-56: Configure model download only...


0pdd:
The puzzle `257-9496b6b4` from #257 has to be resolved: 

https://github.com/objectionary/lints/blob/d48c39e1b58875bd18d2f03f52cd42c9e91e98ca/src/main/java/org/eolang/lints/misc/LtTestNotVerb.java#L52-L56

The puzzle was created by @yegor256 on 23-Jan-25. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

Marat-Tim:
I wasn't sure if I could handle this task, but it was interesting for me to try and it turned out to be quite easy, so I made a pull request

0pdd:
The puzzle `257-9496b6b4` has disappeared from the source code, that's why I closed this issue.

PR #303, opened 2025-02-02
feat(#276): loading the model during compilation

Marat-Tim:
In this pull request, I go through the process of getting a model file to analyze words in verbs. It used to be downloaded in runtime. I tried to fix this by downloading it to the app's resources in compile time


Marat-Tim:
I'm not sure if the downloaded file will end up in the jar being compiled, since so far I don't understand at all where and where our program builds.

Marat-Tim:
@yegor256 Please take a look

yegor256:
@Marat-Tim also, would be great if you can name your pull request better and provide a description for it, check this one please: https://www.yegor256.com/2024/12/15/open-source-beginner-advice.html

Marat-Tim:
hmm, as if testing LtTestNotVerb would test for the presence of a file
added a test, @maxonfjvipon please take a look

src/test/java/org/eolang/lints/ModelExistsTest.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @since 0.0.38
+ */
+final class ModelExistsTest {
+    @Test
+    void testModelExists() {
+        Assertions.assertDoesNotThrow(
+            () -> {
+                final InputStream input = new ResourceOf("en-pos-perceptron.bin").stream();
+                input.readAllBytes();
```
maxonfjvipon, line 44: @Marat-Tim It's not a really good idea to test file presence by reading it. The file may be huge and your OS die. There is a special mechanism and method in Java to check the file presence

maxonfjvipon (changes requested):

maxonfjvipon:
@Marat-Tim I think I was wrong, maybe we shouldn't test things like this since your changes didn't break anything in build pipeline. Let's just remove the test, you added and I think we're good to merge

maxonfjvipon (approved):
@yegor256 please check

yegor256:
@Marat-Tim thanks!

PR #325, opened 2025-02-11
feat(#324): Programs.version()

### LINKED ISSUE #324: `Programs.version` is needed
Currently, we cannot get the version of `lints` library, without creating the `Defect`, since it's implementation is package-private (`Defect.Simple`). Besides that, in [some cases](https://github.com/objectionary/eo/pull/3897) it's suboptimal to retrieve only one element from all the received defects to take its `.version()`. I suggest we introduce new method: `version()` in `Program` or `Programs` in order to do it easier:

```java
final Programs programs = ...;
final String version = programs.version();
```

This idea was initiated by this: https://github.com/objectionary/eo/pull/3897#discussion_r1946703306

h1alexbel:
In this PR I've added `version()` method to `Programs` to read lints version from `MANIFEST.MF`.

closes #324

### DIFF
--- a/src/main/java/org/eolang/lints/Programs.java
+++ b/src/main/java/org/eolang/lints/Programs.java
@@ -23,6 +23,7 @@
  */
 package org.eolang.lints;
 
+import com.jcabi.manifests.Manifests;
 import com.jcabi.xml.XML;
 import com.jcabi.xml.XMLDocument;
 import java.io.FileNotFoundException;
@@ -154,6 +155,15 @@ public final class Programs {
             }
         }
         return messages;
+    }
+
+    /**
+     * Lints version.
+     * @return Version
+     * @checkstyle NonStaticMethodCheck (3 lines)
+     */
+    public String version() {
+        return Manifests.read("Lints-Version");
     }
 
     /**
--- a/src/main/java/org/eolang/lints/Programs.java
+++ b/src/main/java/org/eolang/lints/Programs.java
@@ -23,7 +23,6 @@
  */
 package org.eolang.lints;
 
-import com.jcabi.manifests.Manifests;
 import com.jcabi.xml.XML;
 import com.jcabi.xml.XMLDocument;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


src/test/java/org/eolang/lints/ProgramsTest.java
```diff
@@ -152,6 +152,15 @@ void createsProgramsWithoutMultipleLints(@Mktmp final Path dir) throws IOExcepti
         );
     }
 
+    @Test
+    void returnsVersion() throws IOException {
+        MatcherAssert.assertThat(
+            "Version doesn't match with expected",
+            new Programs(new ListOf<>()).version(),
+            Matchers.equalTo("1.2.3")
```
maxonfjvipon, line 160: @h1alexbel is it correct value for testing?

h1alexbel: @maxonfjvipon yes, it comes from `src/test/resources/META-INF/MANIFEST.MF`:

https://github.com/objectionary/lints/blob/89a791fac307cd6cadab6dc91b3f7460e7a62ddf/src/test/resources/META-INF/MANIFEST.MF#L1

h1alexbel:
@yegor256 take a look, pleaes

src/main/java/org/eolang/lints/Programs.java
```diff
@@ -156,6 +157,15 @@ public Collection<Defect> defects() {
         return messages;
     }
 
+    /**
+     * Lints version.
+     * @return Version
+     * @checkstyle NonStaticMethodCheck (3 lines)
+     */
+    public String version() {
```
yegor256, line 165: @h1alexbel it seems that this method should be static (as checkstyle suggests). However, we don't appreciate static methods. I suggest getting rid of it at all and moving this information into Javadoc. Just explain to the user how he get get the version of the library, via `Manifests`

yegor256 (changes requested):
@h1alexbel looks cool, see one comment above

yegor256:
@h1alexbel thanks!

PR #336, opened 2025-02-12
bug(#307): check for `LtUnlint` in `PkMono`

### LINKED ISSUE #307: `LtAsciiOnly` and `LtTestNotVerb` not wrapper in any Unlint Wrapper
In [PkMono](https://github.com/objectionary/lints/blob/5f5cd402bdb82627d39c4a2212bf97baa59d4858/src/main/java/org/eolang/lints/PkMono.java) creates `LtAsciiOnly` and `LtTestNotVerb`, but they are not wrapped by any Unlint

So if I understand correctly, ignoring these warnings will not work

h1alexbel:
In this PR I've wrapped lints in `PkMono` with `LtUnlint`. Besides that, I added tests that check that each lint can be unlinted.

closes #307
History:
- **bug(#307): +LtUnlint**
- **bug(#307): can be unlinted**
- **bug(#307): filter LtIncorrectUnlint**
- **bug(#307): clean for qulice**

### DIFF
--- a/src/main/java/org/eolang/lints/PkMono.java
+++ b/src/main/java/org/eolang/lints/PkMono.java
@@ -26,6 +26,7 @@ package org.eolang.lints;
 import com.jcabi.xml.XML;
 import java.util.List;
 import javax.annotation.concurrent.ThreadSafe;
+import org.cactoos.Func;
 import org.cactoos.iterable.IterableEnvelope;
 import org.cactoos.iterable.Joined;
 import org.cactoos.iterable.Mapped;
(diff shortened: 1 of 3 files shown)


h1alexbel:
@yegor256 take a look, please

src/main/java/org/eolang/lints/PkMono.java
```diff
@@ -54,7 +54,7 @@ final class PkMono extends IterableEnvelope<Lint<XML>> {
         new Joined<Lint<XML>>(
             new PkByXsl(),
             List.of(
-                new LtAsciiOnly()
+                new LtUnlint(new LtAsciiOnly())
```
yegor256, line 57: @h1alexbel maybe it's better to do the mapping, with the help of `Mapped`, as it's done in `PkByXsl`?

yegor256 (changes requested):
@h1alexbel I think we should remove it from `PkByXsl` too

h1alexbel:

yegor256:
@h1alexbel thanks!

Issue #343, opened 2025-02-17
Add Location of a Problem if `@line` Attribute is Absent


volodya-lombrozo:
We have found, that if `@line` attribute isn't set, we can't fide the place where the issue occurred, for example:

```
[[j$Application application-duality CRITICAL]:0 The anonymous object has duality in the application, which is prohibited,
```
Instead of `0` we have to put something meaningful.

There are several suggestions how to solve this issue:

https://github.com/objectionary/lints/issues/320#issuecomment-2642853520

https://github.com/objectionary/lints/issues/320#issuecomment-2643069193

volodya-lombrozo:
@yegor256 What do you think?

h1alexbel:
@yegor256 I'm in

yegor256:
@h1alexbel sure, please, help!

h1alexbel:
@yegor256 @volodya-lombrozo how about we introduce new `<context/>` node inside the `<defect/>`? For instance, if there are no lines, `application-duality` will produce such a defect:

```xml
<defect line="0" severity="critical">The anonymous object has duality in the application, which is prohibited
  <context>
    <o base="man">
      <o as="name" base="string"/>
      <o base="number"/>
    </o>
  </context>
</defect>
```

In each XSL, where it is possible, we are able to copy the problematic XML like this `<xsl:copy-of select="." />`.

The users of EO don't need this `<context/>` to be printed, since EO-to-Java compiler places lines, while other users, who generate XMIR by themselves (like JEO) may find it useful using lints library:

```java
final Defect defect = ...;
final XML context = defect.context();
```

WDYT?

h1alexbel:
@volodya-lombrozo yes, `<context/>` can be bigger than reasonable size, but depends on specific lint scope. To resolve potential performance issues, we can try to convert `<context/>` to string, and then print only first N chars of the context. WDYT?

h1alexbel:
@volodya-lombrozo About lines I agree, but on the other hand, it can add some processing overhead, since we need to annotate original XMIR with lines, resolve them, and then pass annotated XMIR to lints. @yegor256 Should we try to go for it?

h1alexbel:
@volodya-lombrozo in latest release we added `@context` to most lints, please check

PR #349, opened 2025-02-18
feat(#343): POC: show defect context if `@line` attributes are empty

h1alexbel:
In this PR I've implemented Proof Of Concept of printing defect's context, if inspected XMIR doesn't have `@line` attributes inside.

see #343
History:
- **feat(#343): print first N chars**
- **feat(#343): poc**

### DIFF
--- a/src/main/java/org/eolang/lints/Defect.java
+++ b/src/main/java/org/eolang/lints/Defect.java
@@ -77,6 +77,12 @@ public interface Defect {
     String version();
 
     /**
+     * Defect context.
+     * @return Context of the defect
+     */
+    String context();
+
+    /**
      * Default.
      *
      * @since 0.0.1
@@ -167,6 +173,11 @@ public interface Defect {
         @Override
         public String version() {
             return Manifests.read("Lints-Version");
+        }
+
+        @Override
+        public String context() {
+            return "Context is empty";
         }
     }
 
--- a/src/main/java/org/eolang/lints/DfContext.java
+++ b/src/main/java/org/eolang/lints/DfContext.java
@@ -25,7 +25,7 @@ package org.eolang.lints;
 
 /**
  * Defect with context.
- * @since 0.0.0
+ * @since 0.0.40
  */
 final class DfContext implements Defect {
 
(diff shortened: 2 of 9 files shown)


h1alexbel:
@volodya-lombrozo take a look, please

volodya-lombrozo (changes requested):
@h1alexbel Thank you for the contribution! Just one small comment.  
Additionally, maybe we can add a puzzle that would suggest to fix the issue with [large context values](https://github.com/objectionary/lints/issues/343#issuecomment-2665531562)? What do you think?

h1alexbel:
> @h1alexbel Thank you for the contribution! Just one small comment. Additionally, maybe we can add a puzzle that would suggest to fix the issue with [large context values](https://github.com/objectionary/lints/issues/343#issuecomment-2665531562)? What do you think?

@volodya-lombrozo for lage-context values, we already show only first 300 chars from the context string. In `defect-context.xsl` we do:

```
<xsl:value-of select="substring(serialize($context, map{'method':'xml'}), 1, 300)"/>
```

volodya-lombrozo:
> > @h1alexbel Thank you for the contribution! Just one small comment. Additionally, maybe we can add a puzzle that would suggest to fix the issue with [large context values](https://github.com/objectionary/lints/issues/343#issuecomment-2665531562)? What do you think?
> 
> @volodya-lombrozo for lage-context values, we already show only first 300 chars from the context string. In `defect-context.xsl` we do:
> 
> ```
> <xsl:value-of select="substring(serialize($context, map{'method':'xml'}), 1, 300)"/>
> ```

@h1alexbel Could you add the test for this case please?

h1alexbel:

h1alexbel:

volodya-lombrozo (approved):
@h1alexbel Looks good to me, thank you!

h1alexbel:
@yegor256 take a look, please, what do you think about such feature?

h1alexbel:
@yegor256 reminder

h1alexbel:
@yegor256 how about this experimental feature?

yegor256:
@maxonfjvipon what do you think about this? Not the quality of the PR, but the presence of this feature in the lib.

maxonfjvipon:
@yegor256 it can be useful for XMIRs which are generated not from EO or PHI, where lines are present. So I'm saying rather "yes"

h1alexbel:
@yegor256 WDYT?

PR #356, opened 2025-02-21
feat(#259): `inconsistent-args` lint

### LINKED ISSUE #259: inconsistent arguments count is not detected
Imagine this EO program:

```
[] > foo
  bar 42 > x
  bar 1 2 3 > y
```

It's illegal, because two objects with the same `@base` have different number of arguments.

We should check this on the level of `Programs`, not a single XMIR file.

This should lead to a warning, not error.

h1alexbel:
In this PR I've introduced new lint: `inconsistent-args`, that reports `warning`-s for objects with the same `@base`, but with different number of arguments passed in it in the scope of all programs under analysis.

closes #259
History:
- **feat(#259): test**
- **feat(#259): report localized**
- **feat(#259): opposite refs**
- **feat(#259): allows test**
- **feat(#259): typo**
- **feat(#259): typo**
- **feat(#259): multiple programs**
- **feat(#259): whole merge, invert, report**
- **feat(#259): clean for qulice**

### DIFF
--- a/src/main/java/org/eolang/lints/LtInconsistentArgs.java
+++ b/src/main/java/org/eolang/lints/LtInconsistentArgs.java
@@ -21,6 +21,11 @@ import org.cactoos.text.UncheckedText;
  * Lint for checking arguments inconsistency provided to the objects.
  *
  * @since 0.0.41
+ * @todo #259:60min Optimize performance of inconsistent arguments finding.
+ *  Instead of re-collecting objects in nested loops, we should merge all objects
+ *  from all programs into single XMIR under '<objects/>' element. After objects
+ *  are merged, we can iterate over all the objects there only once, and find
+ *  inconsistencies.
  */
 final class LtInconsistentArgs implements Lint<Map<String, XML>> {
 
--- a/src/main/java/org/eolang/lints/LtInconsistentArgs.java
+++ b/src/main/java/org/eolang/lints/LtInconsistentArgs.java
@@ -39,27 +39,24 @@ final class LtInconsistentArgs implements Lint<Map<String, XML>> {
                 if (counts.stream().distinct().count() != 1L) {
                     final List<Xnav> programs = bases.get(base);
                     programs.forEach(
-                        program ->
-                            program.path(
-                                String.format("//o[@base='%s']", base)
-                                )
-                                .map(o -> Integer.parseInt(o.attribute("line").text().orElse("0")))
-                                .forEach(
-                                    line ->
-                                        defects.add(
-                                            new Defect.Default(
-                                                this.name(),
-                                                Severity.WARNING,
-                                                program.element("program").attribute("name")
-                                                    .text().orElse("unknown"),
-                                                line,
-                                                String.format(
-                                                    "Object '%s' has arguments inconsistency",
-                                                    base
-                                                )
+                        program -> program.path(String.format("//o[@base='%s']", base))
+                            .map(o -> Integer.parseInt(o.attribute("line").text().orElse("0")))
+                            .forEach(
+                                line ->
+                                    defects.add(
+                                        new Defect.Default(
+                                            this.name(),
+                                            Severity.WARNING,
+                                            program.element("program").attribute("name")
+                                                .text().orElse("unknown"),
+                                            line,
+                                            String.format(
+                                                "Object '%s' has arguments inconsistency",
+                                                base
                                             )
                                         )
-                                )
+                                    )
+                            )
                     );
                 }
             }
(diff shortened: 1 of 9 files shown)


h1alexbel:
@maxonfjvipon please check

src/main/java/org/eolang/lints/LtInconsistentArgs.java
```diff
@@ -0,0 +1,150 @@
...
+                new ResourceOf(
+                    String.format(
+                        "org/eolang/motives/misc/%s.md", this.name()
+                    )
+                )
+            )
+        ).asString();
+    }
+
+    private static Map<Xnav, Map<String, List<Integer>>> scanUsages(final Map<String, XML> pkg) {
```
maxonfjvipon, line 102: @h1alexbel it would be better to add documentation to the private static methods so I can understand what they do without getting deep into the code

maxonfjvipon (changes requested):

src/test/java/org/eolang/lints/LtInconsistentArgsTest.java
```diff
@@ -0,0 +1,159 @@
...
+                        ).parsed()
+                    )
+                )
+            ),
+            Matchers.hasSize(2)
+        );
+    }
+
+    @Test
+    void allowsConsistentArgumentsAcrossPrograms() throws IOException {
```
maxonfjvipon, line 127: @h1alexbel let's add one more test with several programs and object used via alias, for example `Q.org.eolang.txt.text`

maxonfjvipon (changes requested):

h1alexbel:

src/main/java/org/eolang/lints/LtInconsistentArgs.java
```diff
@@ -0,0 +1,146 @@
...
+        final Map<String, List<Xnav>> bases = LtInconsistentArgs.baseOccurrences(whole);
+        LtInconsistentArgs.mergedPrograms(whole).forEach(
+            (base, counts) -> {
+                if (counts.stream().distinct().count() != 1L) {
+                    final List<Xnav> programs = bases.get(base);
+                    programs.forEach(
+                        program ->
+                            program.path(
+                                String.format("//o[@base='%s']", base)
+                                )
```
maxonfjvipon, line 45: @h1alexbel something's wrong with indentation here

maxonfjvipon (approved):

src/main/java/org/eolang/lints/LtInconsistentArgs.java
maxonfjvipon: ...

maxonfjvipon (changes requested):

src/main/java/org/eolang/lints/LtInconsistentArgs.java
maxonfjvipon: @h1alexbel maybe you need to do
```
program -> program.path(
        path-here
    )
    .method1()
    .method2()
```

maxonfjvipon:
@h1alexbel I just wonder, wouldn't it be faster to:
1. merge all of the programs first into the single program
2. go though this single program just once and on this only one cycle you definitely can find inconsistency so you won't need to collect the objects and go through them once again

Maybe we can merge this now, but let's add a todo for it? Because it seems to me that the algorithm with this many "hash maps of hash maps" is not such trivial as the origin task sounds like. WDYT?

h1alexbel:
> @h1alexbel I just wonder, wouldn't it be faster to:
> 
> 1. merge all of the programs first into the single program
> 2. go though this single program just once and on this only one cycle you definitely can find inconsistency so you won't need to collect the objects and go through them once again
> 
> Maybe we can merge this now, but let's add a todo for it? Because it seems to me that the algorithm with this many "hash maps of hash maps" is not such trivial as the origin task sounds like. WDYT?

@maxonfjvipon totally agree that current approach of many nested loops not the best one. Let's add puzzle. BTW, what do you meant by "1. merge all the programs into the single program"? My first thought was to create new `XMLDocument` with joined `<o/>` from multiple programs. And then iterate over all `<o/>` there. Is it valid approach?

maxonfjvipon:
@h1alexbel I think all `<o>` objects can be merged under one `<objects>` element

h1alexbel:
@maxonfjvipon added puzzle. Take a look, please

maxonfjvipon (approved):

maxonfjvipon:
@h1alexbel I've just remembered - we should be careful with different packages while merging

h1alexbel:
@yegor256 take a look, please

yegor256:
@h1alexbel thanks!

PR #433, opened 2025-03-25
feat(#156): `sparse-seq` lint

### LINKED ISSUE #156: `seq` of one element must be prohibited
Consider this code:

```
[] > foo
  seq > @
    *
      true
```

This is identical to:

```
[] > foo
  true > @
```

Let's create a new lint to catch such situations: `sparse-seq` may be a good name.

h1alexbel:
In this PR I've implemented new lint: `sparse-seq`, that issues defect with `warning` severity in case of `Q.org.eolang.seq` object was supplied only with one object.

closes #156


h1alexbel:
@maxonfjvipon please check

src/main/resources/org/eolang/lints/misc/sparse-seq.xsl
```diff
@@ -0,0 +1,63 @@
...
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" id="sparse-seq" version="2.0">
+  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="o" mode="arguments" as="xs:integer">
+    <xsl:choose>
+      <xsl:when test="@base='.with' and count(o)=2">
```
maxonfjvipon, line 13: @h1alexbel in EO `0.55.0` we've change the way tree of `tuples` is built (sorry :)). You can check [this](https://github.com/objectionary/eo/pull/4002/files#diff-bf4b779e6cecaa0055fca48abbf62b4cf4bd58cf48f3a18e43779df3cba82554).

Now they are almost the same as they were before:
```
* x y z
# OR
tuple
  tuple
    tuple
      tuple.empty
      x
      1 # this is the length
    y
    2 # this is the length too
  z
  3 # this is also length
```

maxonfjvipon (changes requested):

h1alexbel:

h1alexbel:
@maxonfjvipon merged branch with master. Take a look, please

h1alexbel:
@maxonfjvipon reminder

src/main/resources/org/eolang/lints/misc/sparse-seq.xsl
```diff
@@ -0,0 +1,60 @@
...
+<!--
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" id="sparse-seq" version="2.0">
+  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="o" mode="arguments" as="xs:integer">
```
maxonfjvipon, line 11: @h1alexbel I think it can be done simpler:
```
# THIS
seq
  * x y z
# looks like this after desugaring
seq
  tuple
    tuple
      tuple
        tuple.empty
        x
        1
      y
      2
  z
  3
```

The third argument of every `tuple` is its length. The length of the top-level `tuple` is the entire length of whole "sequence". So it's enough for you to take just `o[3]`. It'll be something like:
```xml
<o base="Q.org.eolang.number">
  <o base="Q.org.eolang.bytes">
    43-43-...-00 // some bytes here
  </o>
</o>
``` 

The you take this bytes (`o[3]/o[1]/text()`) and compare it with `1` in hexadecimal representation (you don't need to unhex bytes). If it's different - the length is not equal to `1`

maxonfjvipon (changes requested):

h1alexbel:

h1alexbel:
@maxonfjvipon reminder

src/main/resources/org/eolang/lints/misc/sparse-seq.xsl
```diff
@@ -0,0 +1,34 @@
...
+            </xsl:attribute>
+            <xsl:if test="$line = '0'">
+              <xsl:attribute name="context">
+                <xsl:value-of select="eo:defect-context(.)"/>
+              </xsl:attribute>
+            </xsl:if>
+            <xsl:attribute name="severity">
+              <xsl:text>warning</xsl:text>
+            </xsl:attribute>
+            <xsl:text>The "Q.org.eolang.seq" object has only one object inside</xsl:text>
```
maxonfjvipon, line 28: @h1alexbel "has only one object inside" sounds a bit weird. Since the void attribute of `seq` is called `steps` I would suggest something like: "The object "Q.org.eolang.seq" dataizes/processes/accepts only one step"

src/main/resources/org/eolang/lints/errors/empty-object.xsl
```diff
@@ -11,7 +11,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/program/objects//o[not(@base) and not(o) and not(eo:atom(.)) and not(parent::o[@base='Q.org.eolang.bytes'])]">
+      <xsl:for-each select="/program/objects//o[not(@base) and not(o) and not(eo:atom(.)) and not(eo:has-data(.) and parent::o[@base='Q.org.eolang.bytes'])]">
```
maxonfjvipon, line 14: @h1alexbel it would be better to add a test case for this change. Or do it in a separated PR

maxonfjvipon (changes requested):

h1alexbel:

yegor256 (approved):

yegor256:
@rultor merge

PR #462, opened 2025-04-04
bug(#460): relatively `program/@source` check in `object-does-not-match-filename`

### LINKED ISSUE #460: `object-does-not-match-filename` complains when the name is almost right
I see this warning:

```
org.eolang.sys.posix:17 It is expected that the "org.eolang.sys.posix" object is
coming from either "org/eolang/sys/posix.eo" or "org/eolang/sys/posix.phi" file, 
however its "program/@source" points to ".eoc/2-pull/org/eolang/sys/posix.eo" 
(object-does-not-match-filename)
```

Looks like a false alarm to me. Maybe `object-does-not-match-filename` should only check the tail of the `@source` attribute?

h1alexbel:
In this PR I've updated `object-does-not-match-filename` to check only `program/@source` contains object name, instead of checking full path `program/@source` starts with object name.

closes #460

### DIFF
--- a/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
+++ b/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(contains(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, '.eo')) or contains(/program/@source, concat($source, '.phi')))">
       <defect>
         <xsl:variable name="line" select="eo:lineno(@line)"/>
         <xsl:attribute name="line">
--- a/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
+++ b/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(contains(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, concat('.', substring-after(substring-after(/program/@source, '.'), '.')))))">
       <defect>
         <xsl:variable name="line" select="eo:lineno(@line)"/>
         <xsl:attribute name="line">
--- a/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
+++ b/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(starts-with(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, '.')))">
       <defect>
         <xsl:variable name="line" select="eo:lineno(@line)"/>
         <xsl:attribute name="line">
--- a/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
+++ b/src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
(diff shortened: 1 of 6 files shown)


h1alexbel:
@maxonfjvipon take a look, please

src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
```diff
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(starts-with(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, '.')))">
```
maxonfjvipon, line 31: @h1alexbel I didn't get why we need `concat($source, '.')`. Maybe it should be `concat($source, '.eo')`?

maxonfjvipon (changes requested):

src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
```diff
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(starts-with(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, '.eo')) or contains(/program/@source, concat($source, '.phi')))">
```
maxonfjvipon, line 31: @h1alexbel I just realized that there are many ways we can get XMIR from (bytecode for example). So it's not really correct to specify a file extension here. Maybe variant with `.` (dot) wasn't such bad. But it's not really strong because file path may contain dots in file name: `org/eolang/seq.some.eo`. In such case `concat($source, '.')` produces false positive

maxonfjvipon (changes requested):

src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
h1alexbel: @maxonfjvipon also, how about we will check all other extensions XMIR can be get from, like we did with `.phi` and `.eo`. Thus, both lint check and its message will be synchronized and controlled. WDYT?

maxonfjvipon: @h1alexbel there are so many of them, we can't cover all... Imagine your `/program/@source` is `foo/bar/seq.some.eo`, your `$source` is `foo/bar/seq` (since it's built from package and object name). Then there's a mismatch between file name and object name

maxonfjvipon (changes requested):

src/main/resources/org/eolang/lints/names/object-does-not-match-filename.xsl
```diff
@@ -28,7 +28,7 @@
       </xsl:choose>
     </xsl:variable>
     <xsl:variable name="source" as="xs:string" select="replace($opath, '\.', '/')"/>
-    <xsl:if test="not(contains(/program/@source, concat($source, '.')))">
+    <xsl:if test="not(contains(/program/@source, concat($source, concat('.', substring-after(substring-after(/program/@source, '.'), '.')))))">
```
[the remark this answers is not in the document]
h1alexbel: @maxonfjvipon it will fail with the following message:

```text
It is expected that the "foo.bar.foo" object is coming from either "foo/bar/foo.eo" or "foo/bar/foo.phi" file, however its "program/@source" points to "foo/bar/foo.seq.bar.baz.xyz.eo"
```

added new test case

maxonfjvipon (approved):

h1alexbel:
@yegor256 take a look, please

yegor256 (approved):

0crat:
@maxonfjvipon 🌟 Great job on the review! You've earned +4 points, boosting your balance to [+112](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the excellent work!

0crat:
@yegor256 Thanks for the review! You've earned +4 points, the minimum reward as per our policy. While there weren't enough comments or hits-of-code to boost your bonus this time, remember that each review contributes to your skills and our project's quality. Your running balance is now [-848](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the good work, and aim for more detailed reviews to maximize your future bonuses! 💻👍

0crat:
@h1alexbel Thank you for your contribution! We appreciate your effort, but we need to adjust the points awarded based on our team policy. The initial +4 points have been reduced due to factors such as the number of hits-of-code, comments, and reviews. Please review our [policy](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) for more details on point calculations. Your current running balance stands at [+124](https://www.eolang.org/zerocracy/objectionary-vitals.html). We encourage you to keep contributing and aim for more comprehensive code reviews in the future. 🚀

PR #468, opened 2025-04-07
bug(#458): allow sparse decoration while docblock above it

### LINKED ISSUE #458: sparse-decoration must stay quite if docblock is present
This is my code:

```
[] > app
  # Checks whether it is TRUE.
  [] > is-true
    0.eq 0 > @
```

The `sparse-decoration` complains and forces me to turn this code into this:

```
[] > app
  # Checks whether it is TRUE.
  0.eq 0 > is-true
```

However, it doesn't compile, because the docblock is allowed only in front of formations. I suggest silencing `sparse-decoration` when docblock is present.

h1alexbel:
In this PR I've updated `sparse-decoration` lint to not issue defect for sparse decoration that have docblock above it.

closes #458
History:
- **bug(#458): test case**
- **bug(#458): check for dockblock**


h1alexbel:
@maxonfjvipon please check

maxonfjvipon:
@h1alexbel let's wait for [this](https://github.com/objectionary/lints/issues/458#issuecomment-2783713547). It seems this change is unnecessary

h1alexbel:
@maxonfjvipon closing this one. Thanks for catch!

PR #490, opened 2025-04-16
feat(#376): `redundant-object` lint

### LINKED ISSUE #376: redundant objects are not caught by lints
Consider this EO program:

```eo
[] > check-array
  list > created
    * 1 1 2
  is-empty.
    created 
```

Here, `created` is [redundant](https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html) object, it was introduced to be used only _once_. It can be inlined like this:

```eo
[] > check-array
  is-empty.
    list
      * 1 1 2
```

Of course this situation should not be reported:

```eo
[] > uses-in-loop
  list > created
    * 1 2 3
  list
    * 1 2 3
  .eachi > @
    [item index] >>
      withi.
        created
        index
        item
```

no defect here, because, `created` was introduced in order to avoid re-creation of array inside the loop (`.eachi`)

h1alexbel:
In this PR I've introduced new lint: `redundant-object`, that issues defect with warning severity if object being created and used once or unused at all.

closes #376
History:
- **feat(#376): motive**
- **feat(#376): test cases**
- **feat(#376): passes**
- **feat(#376): catches unused**
- **feat(#376): catches redundant nested**
- **feat(#376): redundant with dot invocation**
- **feat(#376): allows voids**
- **feat(#376): allows rho access**
- **feat(#376): looped check**
- **feat(#376): eachi and list**
- **feat(#376): catches**
- **feat(#376): remove basing**
- **feat(#376): passes**
- **feat(#376): ignore phi**
- **feat(#376): phi**
- **feat(#376): clean up**


h1alexbel:
blocked by https://github.com/objectionary/lints/issues/495

h1alexbel:
@maxonfjvipon @yegor256 this is my attempt to implement the check for redundant objects, and I think I failed it. Since EO is dynamically-typed language, and we can't list all iteration objects (plus users can create custom ones), we should do something else, to detect redundancy. Consider this EO snippet:

```eo
# Foo.
[] > foo
  1 > x
  y.i > @
    x
```

How about we analyze repeated calls evaluation on object `x`, performed by `y.i`. If `x` object is called only once, then its redundant, otherwise we should not complain.

However, this approach requires more information than we have now in XMIR, and we are not doing linting anymore - its about combination of static and dynamic branches of program analysis. Simply put, we need to know how `foo` behaves during the runtime, in order to detect redundancy.

WDYT?

yegor256:
@yegor256 it seems that #376 suggested to detect compile-time object "usage", not runtime usage. We don't care how many times `x` is used in runtime. We only care how many times `x` is mentioned in the code. Here, it's used once:

```
[] > foo
  "test" > x
  bar x > a
```

Here, it's used twice:

```
[] > foo
  "test" > x
  bar x > a
  bar x > b
```

The semantic of `bar` is irrelevant.

h1alexbel:
@maxonfjvipon please check

src/test/resources/org/eolang/lints/packs/single/redundant-object/catches-redundant-rho-accessed.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - /org/eolang/lints/misc/redundant-object.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=1]
+  - /defects/defect[@line='3']
+input: |
+  # Foo.
+  [] > foo
+    52 > spb
+    [] > tx
+      ^.spb.plus 2 > @
```
maxonfjvipon, line 14: @h1alexbel how is this example is different from the example in motive:
```
# Foo.
[] > foo
  list > created
    * 1 2 3
  list
    * 1 2 3
  .eachi > @
    [item index] >>
      withi.
        created
        index
        item
```

It seems it should not be reported

maxonfjvipon (changes requested):

src/test/resources/org/eolang/lints/packs/single/redundant-object/catches-redundant-rho-accessed.yaml
h1alexbel: @maxonfjvipon it should, according to [this](https://github.com/objectionary/lints/pull/490#issuecomment-2830977552). Let me remove the misleading info from the motive

maxonfjvipon: @h1alexbel also add one more test case with `>>` syntax, let's see how this lint cares of it, in general - it shouldn't complain

src/main/resources/org/eolang/lints/misc/redundant-object.xsl
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="redundant-object" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[@name and @name != '@' and @base and @base != '∅']">
```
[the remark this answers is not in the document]
h1alexbel: @maxonfjvipon defect won't be issued, since, auto named objects do not have both `@name` and `@base`, which are required for select condition in `for-each`

h1alexbel:
@yegor256 take a look, please (Maxim is unavailable for now)

yegor256:
@h1alexbel let's wait for him a little bit, he should be somewhere around soon

maxonfjvipon (approved):

yegor256 (approved):

Issue #15, opened 2024-11-19
The validity of Markdown in comments is not tested


yegor256:
Let's introduce a new `Lint`, which will check the validity of formatting of comments: must be Markdown format. Maybe we can use something like this: https://www.oxygenxml.com/InstData/Editor/SDK/javadoc/ro/sync/exml/workspace/api/markdown/MarkdownValidator.html

Marat-Tim:
Do we have a list of what exactly we want to check? As far as I know, any text is valid markdown. So we want to implement the warnings that are in oxygen editor? If so, maybe you have an example of using this class, because so far I can't use this class properly

I tried this
```java
import ro.sync.exml.workspace.api.markdown.MarkdownValidator;
import ro.sync.exml.workspace.api.markdown.MarkdownValidatorFactory;

import java.io.StringReader;

public class Main {
    public static void main(String[] args) {
        MarkdownValidator markdownValidator = MarkdownValidatorFactory.createSchematronValidator("");
        var validate = markdownValidator.validate(null, new StringReader("# Hello world"));
        System.out.println(validate);
    }
}
```

but so far I'm getting the same error

```
[[SystemID: 
Severity: error
Line: 1
Column: 1
EndLine: -1
EndColumn: -1
Length: -1
Offset: -1
Message: Content is not allowed in prolog.]]
```

On the Internet, people write that such an error usually occurs due to extra characters at the beginning (often the problem was in the BOM), but this is not the case here

Now I'm trying to figure out the code of validator, but the code is quite complicated

yegor256:
@Marat-Tim we should do something similar to what Markdownlint is doing: https://stackoverflow.com/questions/33580413/syntax-code-style-writing-checking-tool-for-markdown-files

yegor256:
@Marat-Tim it's definitely better to reuse an existing linter of Markdown.

VasilevNStas:
@yegor256 @volodya-lombrozo @maxonfjvipon, a note on this one: English grammar/style validation for `.md` files is now wired into the build via **vale** (`.vale.ini` + `vale.yml` workflow, Google/Microsoft/proselint styles, pinned to specific releases per #1180). That covers the "motive" markdown files and any repository `.md`.

[...]

PR #505, opened 2025-04-23
feat(#15): `markdownlint` for check comments

Marat-Tim:
In this pull request, I propose to solve the problem of verifying the validity of comments from the markdown point of view by connecting the markdownlint tool and creating a bridge between java and javascript. Although the running time is very long, it seems to me that this is the best option, since we cannot write and maintain our own library for checking markdown. Also, using outdated tools like oxygenxml doesn't sound good either. Using a bridge between java and javascript will allow us to use the supported markdownlint library as well as the ability to run other js tools in our code.

## Summary by CodeRabbit

* **New Features**
  * Introduced a lint check to validate EO code comments against markdownlint rules, providing warnings for non-compliant Markdown in comments.
  * Added integration with markdownlint via GraalVM and Node.js, automating setup and linting as part of the build process.
  * Automated installation of Node.js and npm, and JavaScript resource management during the build.

* **Documentation**
  * Added user documentation explaining the new markdownlint-based comment validation, with examples of violations and corrections.

* **Tests**
  * Added tests to verify detection of markdownlint violations and acceptance of valid comments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,6 +146,17 @@
       <version>3.0.2</version>
     </dependency>
     <dependency>
+      <groupId>org.graalvm.js</groupId>
+      <artifactId>js</artifactId>
+      <version>22.2.0</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.graalvm.js</groupId>
+      <artifactId>js-scriptengine</artifactId>
+      <version>22.2.0</version>
+    </dependency>
+    <dependency>
       <groupId>com.yegor256</groupId>
       <artifactId>tojos</artifactId>
       <version>0.18.5</version>
@@ -718,6 +729,29 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-resources-plugin</artifactId>
+        <!-- version from the parent pom -->
+        <executions>
+          <execution>
+            <id>copy-javascript</id>
+            <phase>process-resources</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/javascript</outputDirectory>
+              <resources>
+                <resource>
+                  <directory>src/main/javascript</directory>
+                  <filtering>false</filtering>
+                </resource>
+              </resources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-dependency-plugin</artifactId>
         <!-- version from the parent pom -->
         <executions>
@@ -736,6 +770,35 @@
                   <outputDirectory>${project.build.directory}/classes</outputDirectory>
                 </artifactItem>
               </artifactItems>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>com.github.eirslett</groupId>
+        <artifactId>frontend-maven-plugin</artifactId>
+        <version>1.15.1</version>
+        <executions>
+          <execution>
+            <id>install-node-and-npm</id>
+            <goals>
+              <goal>install-node-and-npm</goal>
+            </goals>
+            <configuration>
+              <nodeVersion>v18.17.1</nodeVersion>
+              <npmVersion>9.8.1</npmVersion>
+            </configuration>
+          </execution>
+          <execution>
+            <id>npm-install</id>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <phase>process-resources</phase>
+            <configuration>
+              <arguments>install</arguments>
+              <workingDirectory>${project.build.directory}/javascript</workingDirectory>
+              <installDirectory>.</installDirectory>
             </configuration>
           </execution>
         </executions>
--- a/src/main/java/org/eolang/lints/Defect.java
+++ b/src/main/java/org/eolang/lints/Defect.java
@@ -160,6 +160,21 @@ public interface Defect {
 
         /**
          * Ctor.
+         * @param rule Rule name
+         * @param severity Severity level
+         * @param line Line number
+         * @param text Description of the defect
+         * @checkstyle ParameterNumberCheck (5 lines)
+         */
+        Default(
+            final String rule, final Severity severity,
+            final int line, final String text
+        ) {
+            this(rule, severity, "", line, text);
+        }
+
+        /**
+         * Ctor.
          * <p>
          * Constructs a defect with all required information.
          * </p>
--- a/src/main/java/org/eolang/lints/MarkdownLinter.java
+++ b/src/main/java/org/eolang/lints/MarkdownLinter.java
@@ -1,0 +1,103 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints;
+
+import java.io.Closeable;
+import java.util.stream.Stream;
+import org.cactoos.io.ResourceOf;
+import org.cactoos.io.UncheckedInput;
+import org.cactoos.text.FormattedText;
+import org.cactoos.text.TextOf;
+import org.cactoos.text.UncheckedText;
+import org.graalvm.polyglot.Context;
+import org.graalvm.polyglot.Value;
+
+/**
+ * Markdownlint interop. It can take a very long time to create.
+ *
+ * @since 0.0.47
+ */
+final class MarkdownLinter implements Closeable {
+    /**
+     * JavaScript code of markdownlint with function, that run markdownlint on single string.
+     */
+    private static final String JSCODE = String.format(
+        "%s%s",
+        "globalThis.URL=class{constructor(input){this.href=input;}};",
+        new UncheckedText(
+            new TextOf(
+                new UncheckedInput(
+                    new ResourceOf("markdownlint.js")
+                )
+            )
+        ).asString()
+    );
+
+    /**
+     * JavaScripts context with markdownlint.
+     */
+    private final Context context;
+
+    /**
+     * Function lint from JavaScript, that accept string and return rules violations.
+     */
+    private final Value mdlint;
+
+    @SuppressWarnings("PMD.ConstructorOnlyInitializesOrCallOtherConstructors")
+    MarkdownLinter() {
+        this.context = Context.newBuilder("js")
+            .option("engine.WarnInterpreterOnly", "false")
+            .build();
+        this.context.eval("js", MarkdownLinter.JSCODE);
+        this.mdlint = this.context.getBindings("js").getMember("lint");
+    }
+
+    public Stream<Defect.Default> defects(final String text) {
+        final Stream.Builder<Defect.Default> defects = Stream.builder();
+        final Value errors = this.mdlint.execute(text);
+        for (int idx = 0; idx < errors.getArraySize(); idx += 1) {
+            final Value error = errors.getArrayElement(idx);
+            final String rule = rule(error.getMember("ruleNames"));
+            defects.add(
+                new Defect.Default(
+                    rule,
+                    Severity.WARNING,
+                    error.getMember("lineNumber").asInt(),
+                    new UncheckedText(
+                        new FormattedText(
+                            "[%s] %s. See %s",
+                            rule,
+                            error.getMember("ruleDescription").asString(),
+                            error.getMember("ruleInformation").asString()
+                        )
+                    ).asString()
+                )
+            );
+        }
+        return defects.build();
+    }
+
+    @Override
+    public void close() {
+        this.context.close();
+    }
+
+    /**
+     * MD rule.
+     *
+     * @param names Javascript array of names.
+     * @return The first name starting with MD.
+     * @throws IllegalStateException If no name starting with MD is found.
+     */
+    private static String rule(final Value names) {
+        for (int idx = 0; idx < names.getArraySize(); idx += 1) {
+            final String name = names.getArrayElement(idx).asString();
+            if (name.startsWith("MD")) {
+                return name;
+            }
+        }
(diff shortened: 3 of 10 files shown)


Marat-Tim:
@yegor256

yegor256:
@h1alexbel can you please check this one?

src/main/java/org/eolang/lints/LtMarkdownOnly.java
```diff
@@ -0,0 +1,93 @@
...
+        return new IoCheckedText(
+            new TextOf(
+                new ResourceOf("org/eolang/motives/comments/markdown-only.md")
+            )
+        ).asString();
+    }
+
+    private static Stream<Defect> defectsOfXmir(final XML xmir, final MarkdownLinter mdlinter) {
+        final Stream.Builder<Defect> defects = Stream.builder();
+        final Xnav xml = new Xnav(xmir.inner());
```
h1alexbel, line 66: @Marat-Tim the variable `xml` is redundant, so we can inline it. Also, read this: https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

src/main/java/org/eolang/lints/LtMarkdownOnly.java
```diff
@@ -0,0 +1,93 @@
...
+        final List<Xnav> comments = xml.path("/object/comments/comment")
+            .collect(Collectors.toList());
+        for (final Xnav comment : comments) {
+            final String text = comment.text().get().replace("\\n", "\n");
+            final int length = text.split("\n", -1).length;
+            final int line = Integer.parseInt(comment.attribute("line").text().orElse("0"));
+            mdlinter.defects(text)
+                .filter(defect -> !IGNORED.contains(defect.name()))
```
h1alexbel, line 74: @Marat-Tim for the sake of consistency, its better to access static members via class name: `LtMarkdownOnly.IGNORED`

src/main/java/org/eolang/lints/LtMarkdownOnly.java
```diff
@@ -0,0 +1,93 @@
...
+                .map(
+                    defect -> new Defect.Default(
+                        "lt-markdown-only",
```
h1alexbel, line 77: @Marat-Tim let's reuse lint name from `.name()`

src/main/java/org/eolang/lints/MarkdownDefect.java
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints;
+
+/**
+ * Information about a single defect from markdownlint.
+ *
+ * @since 0.57.0
+ */
+public interface MarkdownDefect {
```
h1alexbel, line 12: @Marat-Tim why new class here? Maybe we can use `Defect.Simple`?

src/main/resources/org/eolang/motives/comments/markdown-only.md
```diff
@@ -0,0 +1,35 @@
+# Comments must be valid markdownlint texts
+
+Comments in the code must be checked for correctness from the point of view of
+[markdownlint](https://github.com/DavidAnson/markdownlint)
+(excluding rules
```
h1alexbel, line 5: @Marat-Tim its good to mention rules we exclude. However, its better to move this text at the bottom of motive

src/main/resources/org/eolang/motives/comments/markdown-only.md
```diff
@@ -0,0 +1,35 @@
...
+[MD041](https://github.com/DavidAnson/markdownlint/blob/main/doc/md041.md),
+[MD047](https://github.com/DavidAnson/markdownlint/blob/main/doc/md047.md),
+[MD026](https://github.com/DavidAnson/markdownlint/blob/main/doc/md026.md)
+)
+
+Incorrect:
```
h1alexbel, line 11: @Marat-Tim let's add an output, that `mdlint` will give us for "incorrect" EO snippet

src/main/resources/org/eolang/motives/comments/markdown-only.md
```diff
@@ -0,0 +1,35 @@
...
+
+```eo
+# # Main object.
```
h1alexbel, line 14: @Marat-Tim to be honest, its not really catchy example of "incorrect" code. Let's use different example to demonstrate that markdown comments are broken. How about this:

```eo
# * Item 1
#  * Nested Item 1
#  * Nested Item 2
#   * A misaligned item
[] > foo
```

This is the violation of [MD005](https://github.com/DavidAnson/markdownlint/blob/main/doc/md005.md)

src/test/java/org/eolang/lints/LtMarkdownOnlyTest.java
```diff
@@ -0,0 +1,84 @@
...
+import matchers.DefectMatcher;
+import org.eolang.parser.EoSyntax;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for {@link LtMarkdownOnly}.
+ *
+ * @since 0.57.0
```
h1alexbel, line 19: @Marat-Tim let's place a valid @since tag here

h1alexbel (changes requested):
@Marat-Tim looks like a good start! Take a look at my comments, please

h1alexbel:
@Marat-Tim also, when you create PR to "close" some issue, [mention](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) it in the PR description, please

pom.xml
```diff
@@ -508,6 +520,22 @@
               </resources>
             </configuration>
           </execution>
+          <execution>
+            <id>copy-js</id>
```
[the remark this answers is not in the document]
Marat-Tim: @h1alexbel does this apply only to the maven id, or do I need to rename all js uses to javascript (for example, in folder names and comments)?

src/main/java/org/eolang/lints/LtMarkdownOnly.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @since 0.57.0
+ */
+final class LtMarkdownOnly implements Lint<Map<String, XML>> {
+    /**
+     * Markdownlint violation names to ignore.
+     */
+    private static final Set<String> IGNORED = Set.of(
+        "MD041",
+        "MD047",
```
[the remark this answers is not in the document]
Marat-Tim: @h1alexbel do we want comments on objects to have an empty line at the end?

It will look like this
```
# Main object.
#
[] > foo
```
Is this exactly what we want? That's not how they write it in the EO code I've seen.

src/main/java/org/eolang/lints/LtMarkdownOnly.java
```diff
@@ -0,0 +1,93 @@
...
+import org.eolang.parser.ObjectName;
+
+/**
+ * Comments should not be violated from the point of view of markdownlint.
+ * This lint works for multiple files, as creating a MarkdownLinter takes a very long time
+ * and creating it many times is a bad idea.
+ *
+ * @since 0.57.0
+ */
+final class LtMarkdownOnly implements Lint<Map<String, XML>> {
```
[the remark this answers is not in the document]
Marat-Tim: @h1alexbel I did this because it takes a lot of time to create a `MarkdownLinter` and it is very expensive to create it for each file. It is better to reuse it. We can require an instance of it in the class constructor and shift responsibility for its lifetime to the library user, but in my opinion this complicates our API

src/main/java/org/eolang/lints/MarkdownDefect.java
Marat-Tim: @h1alexbel I think not. Ideologically, `Defect` is an error in the eolang. But this class is an interop between the java class and the javascript class. I don't think it's worth mixing them up

src/main/java/org/eolang/lints/LtMarkdownOnly.java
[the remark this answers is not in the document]
h1alexbel: @Marat-Tim indeed, you are right. We should ignore this rule

src/main/java/org/eolang/lints/MarkdownDefect.java
h1alexbel: @Marat-Tim I think we just transferring them from `markdownlint` to the `lints`. Thats why its reasonable to use `Defect.Simple` and not overcomplicate the logic with new class

src/main/java/org/eolang/lints/LtMarkdownOnly.java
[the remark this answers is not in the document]
h1alexbel: @Marat-Tim let's document it in the JavaDoc then

src/main/java/org/eolang/lints/MarkdownDefect.java
Marat-Tim: @h1alexbel no, this error appears due to the analysis of plain text (not XMIR). It has no severity, it has no name, it has no version and context. This is a completely different logical entity

h1alexbel: @Marat-Tim yes, but `MarkdownDefect` is used only as intermediate class. Eventually, in the lint you map `MarkdownDefect` into `Defect.Simple`. Seems to be redundant to me. Can we directly produce `Defect.Simple` without mapping?

Marat-Tim: @h1alexbel Are you suggesting that the `MarkdownLinter` class returns a `Stream<Detect.Default>`? Then what should I specify in values that are not present in the context of markdownlint (such as severity, version, context, and others)? Should I specify `null` there?

src/main/resources/org/eolang/motives/comments/markdown-only.md
Marat-Tim: @h1alexbel This seems to be a problem, in the xmir of this program, such a comment looks like this
```xml
<comments>
    <comment line="5">* Item 1\n* Nested Item 1\n* Nested Item 2\n* A misaligned item</comment>
</comments>
```

Marat-Tim:
@h1alexbel I've corrected all the comments except deleting MarkdownDefect. 

Can you clarify exactly whether it needs to be replaced with Defect.Default? And if so, what values should I put in the severity, program, and version fields? Should I set them to null?

src/main/java/org/eolang/lints/MarkdownDefect.java
h1alexbel: @Marat-Tim severity is the same you specify here (`WARNING`):

```java
defect -> new Defect.Default(
    ...
    Severity.WARNING,
    ...
)
```

Context can be empty in `Defect.Default` (see the ctor without context param). About version - you should not set it, it will be automatically resolved in the `.version()` method of the `Defect`'s implementation.

You can omit other parameters: `line` and `program` , while creating them in `MarkdownLinter`. You should be able to create new defect without those parameters in `MarkdownLinter` first, and then, in `LtInvalidMarkdownComment` create `Defect` with those parameters being set.

All in all, my point is that instead of `MarkdownDefect`, we can use `Defect.Simple` in two-steps, but probably we should extend `Defect.Simple` ctors. WDYT?

h1alexbel (changes requested):
@Marat-Tim please check my comments

h1alexbel (changes requested):

src/main/java/org/eolang/lints/Defect.java
```diff
@@ -158,6 +158,21 @@ public Default(
...
+         * @param severity Severity level
+         * @param line Line number
+         * @param text Description of the defect
+         * @checkstyle ParameterNumberCheck (5 lines)
+         */
+        public Default(
+            final String rule, final Severity severity,
+            final int line, final String text
+        ) {
+            this(rule, severity, null, line, text);
```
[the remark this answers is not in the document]
h1alexbel: @Marat-Tim Let's make this ctor private for end-users as well

src/main/java/org/eolang/lints/MarkdownLinter.java
```diff
@@ -84,6 +84,13 @@ public void close() {
...
     }
 
+    /**
+     * Get the first name starting with MD from the name array.
+     *
+     * @param names Javascript array of names.
+     * @return The first name starting with MD.
+     * @throws IllegalStateException If no name starting with MD is found.
+     */
     private static String getName(final Value names) {
```
h1alexbel, line 94: @Marat-Tim let's simply method naming to just `.rule(Value)`

src/main/javascript/markdownlint.js
```diff
@@ -2,7 +2,10 @@
  * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
  * SPDX-License-Identifier: MIT
  */
-
+/**
+ * @todo #15:45min Add eslint to repository https://github.com/eslint/eslint.
+ * We should integrate eslint into the eslint repository to monitor the quality of code written in java script.
```
h1alexbel, line 7: @Marat-Tim `"eslint into the eslint"` typo?

h1alexbel (changes requested):
@Marat-Tim I think these comments are last

src/main/java/org/eolang/lints/MarkdownLinter.java
```diff
@@ -0,0 +1,103 @@
...
+        return defects.build();
+    }
+
+    @Override
+    public void close() {
+        this.context.close();
+    }
+
+    /**
+     * Get the first name starting with MD from the name array.
```
h1alexbel, line 88: @Marat-Tim let's update JavaDoc as well, in order to get rid of `"Get first name"`. Can be simplified just to `"MD rule"`

h1alexbel (approved):
@Marat-Tim LGTM

h1alexbel:
@yegor256 take a look, please

volodya-lombrozo:
Closing this after a very long period of inactivity and a failing pipeline. Thanks for the effort - feel free to reopen if you want to pick it back up.

Issue #514, opened 2025-04-25
`no-attribute-formation` Incorrectly Flags Atoms and `λ` Formations


maxonfjvipon:
In `number.eo` object in `eo-runtime` which is:
```
[as-bytes] > number
  ...
  [] > as-i64 ?
```

I see this:
```
[WARNING] org.eolang.number:31 The formation "as-i64" does not have any void attributes (no-attribute-formation)
[WARNING] org.eolang.number:31 The formation "λ" does not have any void attributes (no-attribute-formation)
```

I think this lint should not complain about atoms and `λ` formations (which is part of atom) at all

maxonfjvipon:
@yegor256 @h1alexbel WDYT?

h1alexbel:
@maxonfjvipon atom formation void attributes are the same attributes we pass to the normal formations, so the `no-attribute-formation` check is right, even though its experimental. About `λ`, agree, we should not complain. WDYT?

0crat:
@maxonfjvipon Hello! Thank you for reporting a new bug in our project. As per our team's policy, you've earned **8** points for this valuable contribution. Your keen eye helps improve our project's quality. If you notice anything else that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) up to par, please [don't hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your current balance stands at [+100](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the great work!

h1alexbel:
@maxonfjvipon please try new version: [0.0.48](https://github.com/objectionary/lints/releases/tag/0.0.48). It should solve the issue

h1alexbel:
@maxonfjvipon please try new version: [0.0.48](https://github.com/objectionary/lints/releases/tag/0.0.48). It should solve the issue

PR #522, opened 2025-04-27
bug(#514): `no-attribute-formation` does not complain about lambdas and nested formations

h1alexbel:
In this pull I've updated `no-attribute-formation` not to complain about lambda formations inside atoms, and allowed nested formations to be attribute-free, if its parent is a formation.

see #514
History:
- **bug(#514): lambda**
- **bug(#514): allows if parent is formation too**
- **bug(#514): polish and tests**


h1alexbel:
@maxonfjvipon please check

maxonfjvipon (commented):
@h1alexbel let's add a test case with 3 nested formations where only the top one has the a void attribute. I suppose it won't work as expected

h1alexbel:
@maxonfjvipon updated. Now we fail the nested formation, if its parent formation does not have any void attributes. Take a look, please

src/test/resources/org/eolang/lints/packs/single/no-attribute-formation/allows-recursive-nested-formation-to-be-attribute-free.yaml
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/design/no-attribute-formation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=1]
```
maxonfjvipon, line 7: @h1alexbel this is exactly what I was thinking. I think we should not fail here since `x` attribute is reachable from `z` formation. What's the difference between `z` and `bar` in such case. Why one fails when the other does not. WDYT?
@yegor256 any thoughts?

yegor256: @maxonfjvipon I think we should report both `z` and `bar`

maxonfjvipon: @yegor256 because if [this](https://github.com/objectionary/lints/issues/514#issuecomment-2830284809)

maxonfjvipon: @yegor256 I vote for the second one
@h1alexbel WDYT?

h1alexbel: @maxonfjvipon @yegor256 agree, let's skip both

h1alexbel:
@maxonfjvipon updated. Now we skip both nested formations. Take a look, please

src/test/resources/org/eolang/lints/packs/single/no-attribute-formation/catches-only-top-attribute-free-formation.yaml
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/design/no-attribute-formation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=1]
```
maxonfjvipon, line 7: @h1alexbel I think there should be 2 defects found

h1alexbel:

src/main/resources/org/eolang/lints/design/no-attribute-formation.xsl
```diff
@@ -12,7 +12,8 @@
   <xsl:template match="/">
     <defects>
       <xsl:if test="not(/object[metas/meta[head='tests']])">
-        <xsl:for-each select="//o[eo:abstract(.) and not(parent::o[@base='Q.org.eolang.bytes']) and not(o[@base='∅'])]">
+        <xsl:variable name="primary" select="/object/o[1]/@name"/>
+        <xsl:for-each select="//o[eo:abstract(.) and not(parent::o[eo:abstract(.) and @name != $primary]) and not(parent::o[o[@base='∅']]) and not(@name = 'λ' and parent::o[eo:atom(.)]) and not(parent::o[@base='Q.org.eolang.bytes']) and not(o[@base='∅'])]">
```
maxonfjvipon, line 16: @h1alexbel what about such pack:
```
[] > foo # <- wrong
  [x] > bar
    [] > baz # <- fine
```
I'm not sure it'll complain correctly because of this `$primary` which is top level object

I think you should not stick with `parent::o`, it's not flexible, use `ancestor::o[eo:abstract(.)]`. The rule is this: if current named(!!!) abstract object has at least one named abstract object at any level above - everything is fine. Which looks something like: `//o[eo:abstract(.) and @name and not(ancestor::o[eo:abstract(.) and @name and not(o[@base='∅']) ...])]`

maxonfjvipon: @h1alexbel ok, how about this:
```
[] > foo
  [] > bar
    [] > baz
      [x] > www
```
Will it catch 3 defects?

h1alexbel:

maxonfjvipon (approved):

h1alexbel:
@yegor256 take a look, please

yegor256 (approved):

yegor256:
@rultor merge

0crat:
@maxonfjvipon Thanks for your code review! Based on our policy, you've earned +4 points for your contribution. This includes factors like hits-of-code, comments, and self-review considerations. Your running balance is now [+140](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the great work! 💻👍

0crat:
@yegor256 Great job on the review! You've earned +4 points for your contribution. While it wasn't your own code (saving you from a 24-point deduction), the review might have benefited from more comments to boost your score further. Keep in mind, each [hit-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) and comment can increase your points next time. Your current balance stands at [-153](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the good work! 💻

0crat:
@h1alexbel Hey there, awesome work on your contribution! 💻 You've snagged +10 points for this one: 4 points just for showing up, plus an extra 5 for those 115 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) you cranked out. Keep it up, and you could earn even more - up to 16 points for code quantity! Just remember, we love quality too, so aim for that sweet spot of 200-800 hits. And don't forget about code reviews - they're crucial for avoiding point deductions. Your current balance is looking great at [+171](https://www.eolang.org/zerocracy/objectionary-vitals.html). [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions coming!

Issue #540, opened 2025-05-04
We don't check English grammar in .md motive files


yegor256:
We don't check them anyhow. Let's introduce some mechanism, into our CI pipeline. Maybe with the help of https://github.com/errata-ai/vale-action or something similar.

Marat-Tim:
@yegor256 Let me do this

0pdd:
@yegor256 the puzzle [#680](https://github.com/objectionary/lints/issues/680) is still not solved.

VasilevNStas:
@yegor256 @Marat-Tim @volodya-lombrozo, the main request of this issue — English grammar/style validation of the motive `.md` files in CI — is **already implemented on `master`**: the `vale` workflow (`.github/workflows/vale.yml`) lints `src/main/resources/org/eolang/motives` with the Google, Microsoft and proselint styles, pinned in `.vale.ini` (see #1180, commit `d31f6aab`). It fails the build on violations (`fail_on_error: true`).

The only remaining part is the `write-good` style package, tracked separately in #680 (still a PDD puzzle). I'd suggest closing this issue and keeping `write-good` as its own ticket.

volodya-lombrozo:
Fixed on master: we now have two automated checks in CI that catch grammar and spelling mistakes. Closing.

0crat:
@marat-tim Thanks for closing this issue! You've earned +4 points for this: +8 as a basis; -20 for 464 days of delay; +16 to give you at least something. Your running score is [+4](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #547, opened 2025-05-04
chore(#540): add vale to check md

Marat-Tim:
In this pull request, I added the Vale tool to check text in markdown files. I thought it’s better to have it not only in GitHub Actions, but also integrated into the build process - if installed locally. However, I didn’t add it to the `mvn` job on purpose, since a build shouldn’t fail because of text issues. I fixed all existing errors and added commonly used words to a custom dictionary. If you think some words shouldn’t be in the dictionary and should be fixed instead - let me know. Also curious to hear your thoughts: is it better to show errors as comments in the pull request(might get spammy), or in the job logs(harder to view)?

## Summary by CodeRabbit

* **New Features**
  * Introduced automated style and prose checking for Markdown files using Vale, integrated with CI and build scripts.
  * Added configuration files and vocabulary customization for style consistency.

* **Documentation**
  * Standardized capitalization and improved wording across all Markdown documentation for clarity and consistency.
  * Updated README with instructions for using Vale to check Markdown files.

* **Tests**
  * Adjusted test assertions to match updated documentation capitalization.

### DIFF
--- a/.github/styles/config/vocabularies/Custom/accept.txt
+++ b/.github/styles/config/vocabularies/Custom/accept.txt
@@ -14,3 +14,4 @@ JVM(s)?
 XMIR
 SPDX
 ASCII
+inlined
--- a/.github/styles/config/vocabularies/Custom/accept.txt
+++ b/.github/styles/config/vocabularies/Custom/accept.txt
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+[Aa]ttr(s)?
+[Mm]eta(s)?
+[Rr]untime(s)?
+[Dd]ecoratee
+FQN(s)?
+[Ll]int(s)?
+([Uu]n)?lint(s)?
+[Mm]eta(s)?
+[Aa]rg(s)?
+[Ff]ormatter(s)?
+JVM(s)?
+XMIR
+SPDX
+ASCII
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+name: vale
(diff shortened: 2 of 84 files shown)


yegor256:
@Marat-Tim I suggest:

* place `.vale.ini` into the `.github/` directory
* remove vale checking from `pom.xml` and let it be checked ONLY by GitHub CI (no Maven)
* use the following config (for stronger validation):

```
MinAlertLevel = suggestion
Packages = Google, Microsoft, write-good, proselint
[*]
BasedOnStyles = Google, Microsoft, write-good, proselint
```

Marat-Tim:
@yegor256 
1. The GitHub Action you suggested only runs on files changed in the pull request.
So if we add a new style package but don’t touch any markdown files, the job will still pass.
If we integrate Vale into Maven instead, the build will show issues across all files, and we can fix them right away.
Which behavior do we prefer the first or the second?

2. Also, just to clarify: if we include all those style packages, we’ll get 400+ errors in markdown files.
As far as I can tell, they can’t be auto-fixed.
And based on your config, it looks like you're proposing to lint all files in the project.
That would either require a lot of transformers for various file types
(otherwise this error occurs
```console
vale src 
E100 [lintXML] Runtime error

no XSLT transform provided

Execution stopped with code 1.
```
) or very flexible rules.

yegor256:
@Marat-Tim it looks ugly inside Maven, that's why. Not every user has `vale` installed. The pom.xml gets too complicated, for no reason. It's better to keep such checkers outside of Maven - they have nothing to do with Java build pipeline.

Marat-Tim:
@yegor256 How can I fix ort job problem?

0crat:
@marat-tim Great work on your contribution! 🎉 You've earned +4 points: +16 base, -8 for exceeding hits-of-code limit (338 ≥ 200), -16 for missing code review, +12 bonus. We appreciate your effort! [Keep the contributions coming](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your current balance is [+23](https://www.eolang.org/zerocracy/objectionary-vitals.html). Looking forward to your next submission!

Issue #215, opened 2025-01-09
junit-platform.properties:23-25: Configure default...


0pdd:
The puzzle `107-b3a5bdea` from #107 has to be resolved: 

https://github.com/objectionary/lints/blob/85722d6ad5f22ecb625fc308aff5b54f7b0ee9d6/src/test/resources/junit-platform.properties#L23-L25

The puzzle was created by @yegor256 on 09-Jan-25. 

Estimate: 45 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
I can't create github labels ``. Most likely I don't have necessary permissions to `objectionary/lints` repository. Please, make sure @0pdd user is in the [list of collaborators](https://github.com/objectionary/lints/settings/collaboration):

```Octokit::UnprocessableEntity
POST https://api.github.com/repos/objectionary/lints/issues/215/labels: 422 - Validation Failed
Error summary:
  resource: Label
  code: unprocessable
  field: data
  message: Could not resolve to a node with the global id of 'I_kwDONQmRdM6lkv1Z'. // See: https://docs.github.com/rest/issues/labels#add-labels-to-an-issue
/app/vendor/bundle/ruby/3.1.0/gems/octokit-8.1.0/lib/octokit/response/raise_error.rb:14:in `on_complete'
/app/vendor/bundle/ruby/3.1.0/gems/faraday-2.9.0/lib/faraday/middleware.rb:18:in `block in call'
/app/vendor/bundle/ruby/3.1.0/gems/faraday-2.9.0/lib/faraday/response.rb:42:in `on_complete'
/app/vendor/bundle/ruby/3.1.0/gems/faraday-2.9.0/lib/faraday/middleware.rb:17:in `call'
/app/vendor/bundle/ruby/3.1.0/gems/octokit-8.1.0/lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
/app/vendor/bundle/ruby/3.1.0/gems/octokit-8.1.0/lib/octokit/middleware/follow_redirects.rb:61:in `call'
[...]

PR #568, opened 2025-05-07
chore(#215): test timeout depends on os

Marat-Tim:
In this pull request, I set the test timeout values to 15 seconds for Unix and Mac, and 45 seconds for other os. I did this for all test runs(including benchmarks and others), and I’m not sure if this is the right approach. I also tried to make profile handling more elegant, but I didn’t understand how to do it https://stackoverflow.com/questions/79610681/how-to-create-profile-unix-or-mac-and-profile-other-os-in-maven

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@
     <skipITs/>
     <!-- This is the flag that allows to skip all tests -->
     <skipTests/>
+    <test.timeout>45s</test.timeout>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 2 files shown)


yegor256:
@Marat-Tim this looks like a hack to me. It's unclear why numbers are different in different OSs and why their values are set like this. What problem are you trying to solve with this PR?

h1alexbel:
@Marat-Tim I was wrong, we should be able just to increase timeout

Issue #566, opened 2025-05-07
`invalid-name-notation` lint is not implemented


yegor256:
Let's create a new `invalid-name-notation` lint, that will check that all names in the `@name` attribute match this regexp:

```
[a-z]+(-[a-z]+)*
```

yegor256:
@Marat-Tim yes, of course. only letters are allowed.

Marat-Tim:
@yegor256 Ok, let me do this

PR #579, opened 2025-05-09
feat(#566): check name match kebab-case

Marat-Tim:
In this pull request, I added a new linter that checks whether all object names follow kebab-case using only Latin letters.


Marat-Tim:
@yegor256

src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="invalid-name-notation">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:import href="/org/eolang/funcs/special-name.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:variable name="pattern" select="'^[a-z]+(-[a-z]+)*$'"/>
+  <xsl:template match="/">
+    <defects>
```
kreinba, line 13: The selector only filters via `eo:special(@name)`, which knows about `@`, `λ`, and cactoos-prefixed names. Parser-injected names from `[] >>` (the test pack already has one literal example: `auto-named-attr-at-4-6` in `packs/single/duplicate-names/catches-conflict-with-auto-name.yaml`) contain digits and a non-kebab segment, so this for-each will report them as defects. Either extend `eo:special` to recognise the `auto-named-attr-at-*` prefix or add a sibling predicate here that excludes that prefix before the regex check.

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-special-names.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - /org/eolang/lints/names/invalid-name-notation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+input: |
+  # App.
+  [] > app
+    x > f
+      [] >>
+    foo > @
+    [] > baz ?
```
kreinba, line 14: This input uses `[] >>` and `[] > baz ?` but the asserts only check the total defect count. Add an input case whose XMIR contains a name like `auto-named-attr-at-4-6` (the same shape the parser produces for `>>`) and assert it is not reported. Without that case the lint silently regresses on the auto-named path, which is what is taking down `acceptsCanonicalCode`.

kreinba (changes requested):
The lint XSL and motive read well and the test fixtures cover the happy path, but CI is red on every mvn matrix job and the canonical-code test almost certainly fails because the parser injects auto-generated names like `auto-named-attr-at-4-6` for the `>>` syntax. Those names contain digits and are not in `eo:special`, so the for-each selector flags them. The `allows-special-names.yaml` fixture exercises this `[] >>` shape but does not assert against the injected name, so the gap is invisible from the single-lint pack and only surfaces under the full mono run. Two inline notes pinned below.

volodya-lombrozo:
Thanks for working on this! Closing since CI is failing and it's been quiet for over a month. Feel free to reopen if you'd like to revisit.

0crat:
@kreinba Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+101](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #583, opened 2025-05-10
Why is package separation missing?


Marat-Tim:
I remember the project used to have a well-structured package layout
Now all classes are in a single package - why is that?
Is it really just for the sake of encapsulation?

Marat-Tim:
@yegor256 Looks like you’re the one who committed the change moving all classes into the main package - could you share the reasoning behind that?

yegor256:
@Marat-Tim that's for the sake of reducing the public surface of the package: https://www.javadoc.io/doc/org.eolang/lints/latest/org/eolang/lints/package-summary.html Now, it's pretty small, just like we want it to be. With packages it will be much larger, confusing our clients.

Marat-Tim:
@yegor256 But won’t it be inconvenient for us to work without proper package separation?
For example, adding all the new classes related to automatic fixes into org.eolang.lints looks messy
In that case, would it make sense to create a separate package specifically for those classes?

yegor256:
@Marat-Tim I suggest we stay in one package for now. We have class-name prefixes, to help us group files. A package is something we deliver to a client.

h1alexbel:
@Marat-Tim still an issue?

Issue #556, opened 2025-05-06
Auto-fix is not present


Marat-Tim:
I suggest adding a `fix()` method to the `Defect` interface, which will return data about a possible fix for the defect. I propose using a format similar to [CodeAction](https://javadoc.io/doc/org.eclipse.lsp4j/org.eclipse.lsp4j/0.8.1/org/eclipse/lsp4j/CodeAction.html) from the LSP protocol. This format supports both simple fixes within a single file and structural changes, such as adding or removing files. By using this format, we can easily integrate the linter into our [LSP server](https://github.com/objectionary/eo-vscode/tree/master/server). Also, I think it won't be difficult to add support for this data in the [eoc](https://github.com/objectionary/eoc)

It might make sense to directly include the [lsp4j](https://mvnrepository.com/artifact/org.eclipse.lsp4j/org.eclipse.lsp4j) library in our project since it’s lightweight(less than a megabyte) and contains all the necessary classes for working with codeAction

Example usage of lsp4j
```java
CodeAction codeAction = new CodeAction();
codeAction.setEdit(
    new WorkspaceEdit(
        Map.of(
            "./test.txt",
            List.of(
                new TextEdit(
                    new Range(
                        new Position(1, 1),
                        new Position(1, 1)
                    ),
                    "test"
                )
            )
        )
    )
);
```

h1alexbel:
@Marat-Tim I like the idea. Automated repair of lint violations is an interesting perspective. Since EO Linter is about style checking, and not about static analysis, auto fixes can be introduced almost for every rule. About implementation, I think we need to implement some XMIR re-writing rather than text editing, as in suggested `lsp4j` library (correct me if I wrong). In other words, while repairing, we need to have IR semantics, not text ones

Marat-Tim:
@h1alexbel No, we should suggest fixes for the original EO source code. Users won’t be able to apply fixes to XMIR code if their project uses `.eo` files

yegor256:
@Marat-Tim @h1alexbel I also think that we need to suggest fixes for EO, not XMIR. The application of fixes will be the job of eo-maven-plugin, not Lints.

0crat:
[...]

PR #584, opened 2025-05-10
feat(#556): auto-fix architecture

Marat-Tim:
1. I didn’t find any clear standard format for describing code fixes, so I went with the LSP `DocumentChanges` format. It’s reasonably clean, supports all kinds of edits, and integrates well with our LSP server. For `eoc`, though, we’ll likely need to parse and apply changes manually([question](https://stackoverflow.com/questions/79608933/how-to-execute-lsp-codeaction-in-local-file-system-in-java)). Hopefully, it won’t be too complex

2. I originally planned to use the `lsp4j` library, hoping it would support easy (de)serialization. But it uses a complex `GSON` setup, and `gson-xml` failed to handle it. So I decided to recreate the `DocumentChanges` structure with our own data classes

3. Added the `fix` package to hold data classes describing fixes

    3.1. Initially, I made all classes interfaces with `Default` implementations(similar to the `Defect` data class). 
    But later realized that was overkill - no one would realistically extend these. So I switched to plain Java data 
    classes. Now, creating a fix is simpler:
    ```java
    new Fix(
      new File(
          new ObjectSource(xmir).get(),
          new Edit(
              new Position(line, 0),
              new Position(line + 1, 0),
              ""
          )
      )
    )
    ```
    instead of
    ```java
    new Fix.Default(
      new File.Default(
          new ObjectSource(xmir).get(),
          new Edit.Default(
              new Position.Default(line, 0),
              new Position.Default(line + 1, 0),
              ""
          )
      )
    )
    ```

    3.2. I’m aware that we aim to keep the public API as minimal as possible, and a public fix package doesn't quite align with that. Still, putting all these classes into `org.eolang.lints` would look messy. Likewise, moving `Fix` to `org.eolang.lints` and nesting the rest as inner classes would be ugly. I considered keeping only interfaces and implementing them via anonymous classes where needed, but that would be too verbose. This might be worth a separate discussion [here](https://github.com/objectionary/lints/issues/583).

4. By default, linters still don’t offer fixes and behave as before. But to add fix support:
    4.1. For Java-based linters, just pass fix data into the constructor(like [this](https://github.com/Marat-Tim/lints/blob/407b72c8b20c59d42c5153195e248b716c7acb40/src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java#L103))
    4.2. For XSL linters, move the message to a `<message>` tag and put the fix info in an `<edit>` tag(like [this](https://github.com/Marat-Tim/lints/blob/407b72c8b20c59d42c5153195e248b716c7acb40/src/main/resources/org/eolang/lints/aliases/empty-alias.xsl#L29)). I assumed XSL linters won’t need file creation or multi-file edits, so one edit per case should be enough

### DIFF
--- a/src/main/java/org/eolang/lints/Defect.java
+++ b/src/main/java/org/eolang/lints/Defect.java
@@ -6,6 +6,7 @@ package org.eolang.lints;
 
 import com.jcabi.manifests.Manifests;
 import java.util.Objects;
+import org.eolang.lints.fix.Fix;
 
 /**
  * A single defect found.
@@ -103,6 +104,13 @@ public interface Defect {
     boolean experimental();
 
     /**
+     * Information on how to fix this defect.
+     *
+     * @return Fix info.
+     */
+    Fix fix();
+
+    /**
      * Default implementation of {@link Defect}.
      * <p>
      * Provides a standard implementation with basic functionality.
@@ -143,6 +151,11 @@ public interface Defect {
         private final boolean experiment;
 
         /**
+         * How to fix.
+         */
+        private final Fix fix;
+
+        /**
          * Ctor.
          * @param rule Rule name
          * @param severity Severity level
@@ -155,7 +168,7 @@ public interface Defect {
             final String rule, final Severity severity,
             final String object, final int line, final String text
         ) {
-            this(rule, severity, object, line, text, false);
+            this(rule, severity, object, line, text, false, new Fix());
         }
 
         /**
@@ -175,7 +188,7 @@ public interface Defect {
         public Default(
             final String rule, final Severity severity,
             final String object, final int line, final String text,
-            final boolean exprmnt
+            final boolean exprmnt, final Fix fix
         ) {
             this.rle = rule;
             this.sev = severity;
@@ -183,6 +196,7 @@ public interface Defect {
             this.lineno = line;
             this.txt = text;
             this.experiment = exprmnt;
+            this.fix = fix;
         }
 
         @Override
@@ -235,6 +249,11 @@ public interface Defect {
         @Override
         public boolean experimental() {
             return this.experiment;
+        }
+
+        @Override
+        public Fix fix() {
+            return this.fix;
         }
 
         @Override
--- a/src/main/java/org/eolang/lints/LtByXsl.java
+++ b/src/main/java/org/eolang/lints/LtByXsl.java
@@ -20,6 +20,9 @@ import org.cactoos.Input;
 import org.cactoos.io.ResourceOf;
 import org.cactoos.text.IoCheckedText;
 import org.cactoos.text.TextOf;
+import org.eolang.lints.fix.Edit;
+import org.eolang.lints.fix.File;
+import org.eolang.lints.fix.Fix;
 import org.eolang.parser.ObjectName;
 
 /**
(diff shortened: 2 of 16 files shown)


Marat-Tim:
@yegor256 @h1alexbel I know the pipeline isn’t passing right now, but could you take a look at the idea behind it?
Do you like the overall direction, or is there anything you think should be changed?
If so, I'll make the adjustments and fix the pipeline

src/main/java/org/eolang/lints/LtByXsl.java
```diff
@@ -98,15 +101,26 @@ public Collection<Defect> defects(final XML xmir) {
...
+                        message.map(
+                                ignored ->
+                                    new Fix(
+                                        new File(
+                                            xmir.xpath("/@source/text()").get(0),
+                                            new Edit(xml.element("edit"))
+                                        )
+                                    )
+                            )
+                            .orElseGet(Fix::new)
```
h1alexbel, line 123: @Marat-Tim maybe we should throw an exception in this case? `Fix::new` only will make things more complicated

src/main/java/org/eolang/lints/fix/Create.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints.fix;
+
+import com.github.lombrozo.xnav.Xnav;
+
+public class Create {
```
h1alexbel, line 9: @Marat-Tim these classes should be package private. We don't want to expose them to the client

src/main/java/org/eolang/lints/fix/File.java
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints.fix;
+
+import com.github.lombrozo.xnav.Xnav;
+import java.util.Arrays;
+import java.util.Collection;
+
+public class File {
```
h1alexbel, line 11: @Marat-Tim `File` is a name of the well-known Java class. Let's rename this class to avoid confusion

src/main/java/org/eolang/lints/ObjectSource.java
```diff
@@ -0,0 +1,20 @@
...
+
+final class ObjectSource {
+    private final Xnav xnav;
+
+    ObjectSource(XML xml) {
+        this.xnav = new Xnav(xml.inner());
+    }
+
+    String get() {
+        return xnav.attribute("source").text().orElse("");
```
h1alexbel, line 18: @Marat-Tim XMIR does not have `@source` attribute anymore, you should use `org.eolang.parser.ObjectName` instead

src/main/java/org/eolang/lints/Defect.java
```diff
@@ -102,6 +103,13 @@ public interface Defect {
      */
     boolean experimental();
 
+    /**
+     * Information on how to fix this defect.
+     *
+     * @return Fix info.
+     */
+    Fix fix();
```
h1alexbel, line 111: @Marat-Tim to be honest, it is hard to review `Fix` architecture without unit tests. For now, I don't really understand how it works. So, I suggest to create unit tests, or even test YAML stories (the way we did in `LtByXslTest#testsAllLintsByEo` and `WpaLintsTest#testsAllLintsByEo`) to see how it works in practice (EO before fix vs. EO after fix).

h1alexbel (changes requested):

volodya-lombrozo:
Closing this one since there is no activity anymore

PR #612, opened 2025-05-21
bug(#608): reserved names processing in deep profile

### LINKED ISSUE #608: Unnecessary processing of home object names in default build lifecycle
Currently, we are collecting and processing [home](https://github.com/objectionary/home) object names on every build in the default lifecycle. Its rather long, and seems unnecessary, since we run the tests for this functionality only in `deep` profile. I suggest to move all the plugins, related to reserved object names processing to our `deep` profile, that will be called only on our CI servers. Locally, we should not download and process object names

h1alexbel:
In this PR I've moved home object names processing, from default build lifecycle, to the `deep` profile.

closes #608

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -408,11 +408,10 @@
           <plugin>
             <groupId>com.googlecode.maven-download-plugin</groupId>
             <artifactId>download-maven-plugin</artifactId>
-            <version>1.13.0</version>
             <executions>
               <execution>
                 <id>download-home</id>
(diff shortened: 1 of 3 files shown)


volodya-lombrozo (approved):
@h1alexbel Looks good to me. Thank you!

h1alexbel:
@yegor256 take a look, please

yegor256 (approved):

Issue #606, opened 2025-05-20
`redundant-object` lint misidentifies non-redundant object usage with dot notation


h1alexbel:
Consider this EO:

```eo
# Test.
[] +> tests-reads-the-length-with-n2-byte-characters
  and. > @
    str.length.eq 12
    str.as-bytes.size.eq 16
  "Hello, друг!" > str
```

Here, we have a unit test, that, introduces object `str`, and uses it two times via dot notation `$.str.`. However, `redundant-object` lint still treat it as "redundant":

```text
The object "str" is redundant, consider inline it instead (redundant-object (Single))
```

We should fix it

h1alexbel:
@yegor256 new bug

h1alexbel:
closing, since resolved

PR #613, opened 2025-05-21
bug(#606): `redundant-object` usage regex supports dot chaining

h1alexbel:
In this PR I've fixed the bug with `redundant-object` lint, where usage regex was not match object chaining via 2+ `.`. For instance: `$.str.as-bytes.size.eq`. Now we count it as an object usage.

see #606


h1alexbel:
@maxonfjvipon please check

maxonfjvipon (approved):

yegor256 (approved):

PR #754, opened 2025-09-15
Added linter for warning about unnecessary usage of ^ in code

### LINKED ISSUE #690: Linter doesn't warn about unnecessary usage of `^` in code
This is my code:

```
# App.
[n] > app
  foo.plus 1 > @
  [x] > foo
    ^.n > @
```

I don't get any warnings. However, I should. The last line may be written shorter:

```
# App.
[n] > app
  foo.plus 1 > @
  [x] > foo
    n > @
```

No need to use `^` if it's not required to resolve ambiguity between names. There is only one `n` in this program, the use of `^` is not motivated.

Let's create a new lint to check this.

Snehakb1219-christ:
Fixes https://github.com/objectionary/lints/issues/690

## Summary by CodeRabbit

* **New Features**
  * Added a lint rule that warns about redundant ξ.ρ notation when an object can be accessed without it.
  * Emits clear WARNING messages indicating when a name resolves to the same target without ξ.ρ.
  * Rule identifier: “redundant-hat”.

* **Tests**
  * Added unit tests ensuring no warnings when ξ.ρ is necessary and warnings appear when it’s redundant.
  * Verified rule naming and behavior across single and multiple match scenarios.


src/test/java/org/eolang/lints/LtRedundantHatTest.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Test for {@link LtRedundantHat}.
+ *
+ * @since 0.0.1
+ */
+final class LtRedundantHatTest {
+
+    @Test
+    void reportsNoDefectsForNecessaryHat() throws IOException {
+        final XML xml = new XMLDocument(
```
h1alexbel, line 24: @Snehakb1219-christ let's use `org.eolang.parser.EoSyntax` for EO programs construction

src/test/java/org/eolang/lints/LtRedundantHatTest.java
```diff
@@ -0,0 +1,83 @@
...
+        MatcherAssert.assertThat(
+            "Motive must explain when to omit redundant '^'",
+            lint.motive(),
+            Matchers.containsString("notation is redundant and can be omitted for brevity")
+        );
+    }
+
+    @Test
+    void namesStableId() {
+        final LtRedundantHat lint = new LtRedundantHat();
```
h1alexbel, line 48: @Snehakb1219-christ `lint` variable is redundant here, please read this: https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

src/main/java/org/eolang/lints/LtRedundantHat.java
```diff
@@ -0,0 +1,82 @@
...
+ * Lint that warns if a redundant {@code ^} is used.
+ * @since 0.0.59
+ */
+final class LtRedundantHat implements Lint<XML> {
+    @Override
+    public Collection<Defect> defects(final XML xmir) throws IOException {
+        final Collection<Defect> defects = new ArrayList<>(0);
+        final Xnav xml = new Xnav(xmir.inner());
+        final List<Xnav> objs = xml
+            .path("//o[@base='^']")
```
h1alexbel, line 26: @Snehakb1219-christ in EO, we can't have `@base="^"`, the correct representation of attribute is `@base="ξ.ρ"`. After `ξ.ρ` we can call any other attribute, e.g.: `ξ.ρ.foo`.

src/main/java/org/eolang/lints/LtRedundantHat.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.eolang.parser.OnDefault;
+
+/**
+ * Lint that warns if a redundant {@code ^} is used.
+ * @since 0.0.59
+ */
+final class LtRedundantHat implements Lint<XML> {
```
h1alexbel, line 20: @Snehakb1219-christ looks like we can implement this lint in XSL, there is no need to do it in Java, I believe

h1alexbel (changes requested):
@Snehakb1219-christ please check

src/main/java/org/eolang/lints/LtRedundantHat.java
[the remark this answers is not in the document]
Snehakb1219-christ: @h1alexbel Implemented the feedback

src/test/java/org/eolang/lints/LtRedundantHatTest.java
Snehakb1219-christ: @h1alexbel Implemented the feedback

src/main/java/org/eolang/lints/LtRedundantHat.java
Snehakb1219-christ: @h1alexbel It works fine in java also. I think it's necessary to do this in XSL.
WDYT?

src/test/java/org/eolang/lints/LtRedundantRhoTest.java
```diff
@@ -0,0 +1,65 @@
...
+            Matchers.equalTo("redundant-hat")
+        );
+    }
+
+    @Test
+    void doesNotReportDefectWhenCaretIsNecessary() throws IOException {
+        MatcherAssert.assertThat(
+            "Should not report a defect when '^' is necessary due to ambiguity",
+            new LtRedundantRho().defects(
+                new EoSyntax("[] > foo\n+test > @\n^foo > bar").parsed()
```
h1alexbel: @Snehakb1219-christ this syntax will have `/object/errors` after `.parsed()`. Maybe you meant this:

```eo
[] > foo
  [foo] > abc
    test > @
      ^.foo > bar
```

h1alexbel (changes requested):
@Snehakb1219-christ please check

volodya-lombrozo:
Thanks for putting this together and for the back-and-forth on the review! This has been open a long time now without the requested changes being addressed, so I'll close it to keep the queue tidy. If you'd like to pick it back up, a fresh PR with the review feedback incorporated would be very welcome.

PR #842, opened 2026-04-22
#805 Ignore @todo and @fixme markers in comment-not-capitalized lint

idrisoffrinat-cpu:
## Summary

Closes #805.

The `comment-not-capitalized` lint previously flagged any comment that didn't begin with an uppercase letter — including conventional tracker markers such as `@todo` / `@fixme`. These are structural, not prose, and shouldn't be held to the same rule:

```eo
# @todo #4475:35min Enable this feature after resize is implemented.
[] > foo
  42 > @
```

Before this PR, the snippet above produced:
```
The comment doesn't start with a capital English letter (comment-not-capitalized)
```

## Fix

Extend the XPath predicate in [comment-not-capitalized.xsl](src/main/resources/org/eolang/lints/comments/comment-not-capitalized.xsl) so comments matching `^@(todo|fixme)\b` (case-insensitive) are skipped. The `\b` word boundary prevents `@todolist` or similar from being accidentally exempted.

## Test plan

Added three pack tests under [comment-not-capitalized/](src/test/resources/org/eolang/lints/packs/single/comment-not-capitalized/):

- `allows-todo-marker.yaml` — standard `# @todo #NNN:NNmin …` form
- `allows-fixme-marker.yaml` — `@fixme` variant
- `allows-todo-case-insensitive.yaml` — `@TODO` uppercase variant

The existing `catches-not-capitalized-comment.yaml` still passes and guards against regressions on regular prose.

The motive doc [comment-not-capitalized.md](src/main/resources/org/eolang/motives/comments/comment-not-capitalized.md) has an example of the newly-exempt form.

## Downstream impact

This unblocks [this TODO in `objectionary/eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/pom.xml) referencing #805, which currently disables the `comment-not-capitalized` lint in `eo-runtime` until this fix lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Comments starting with tracker markers (`@todo`, `@fixme`, case-insensitive) are now exempt from capitalization requirements in linting rules.

* **Documentation**
  * Updated rule documentation to clarify the exemption for tracker-marked comments.

* **Tests**
  * Added test cases verifying proper handling of tracker markers in comments.

PR #843, opened 2026-04-23
#841: fix NPE in DefectMissing when +unlint range references absent lint

### LINKED ISSUE #841: NullPointerException when +unlint uses range pattern for absent lint in DefectMissing
## What Happens

Calling `LtUnlintNonExistingDefect.defects()` on an EO file where `+unlint` uses a range pattern (`name:line-line`) and the referenced lint did not fire crashes with `NullPointerException`:

```
java.lang.NullPointerException:
    Cannot invoke "java.util.List.stream()" because "lines" is null
    at org.eolang.lints.DefectMissing.apply(DefectMissing.java:47)
    at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
    at java.base/java.util.stream.DistinctOps$1$2.accept(DistinctOps.java:174)
    ...
    at org.eolang.lints.LtUnlintNonExistingDefect.defects(LtUnlintNonExistingDefect.java:86)
```

## What Should Happen

Returns a non-empty collection reporting that the `+unlint` annotation references a non-existing defect — consistent with how non-range `+unlint nonexistent` behaves (works correctly).

## Steps to Reproduce

```java
new LtUnlintNonExistingDefect(
    new ListOf<>(new LtAsciiOnly()),
    new ListOf<>()
).defects(
    new EoSyntax(
        String.join(
            "\n",
            "+unlint ascii-only:1-5",
            "[] > main",
            "  QQ.io.stdout > @",
            "    \"Hello\""
        )
    ).parsed()
);
// throws NullPointerException instead of returning defects
```

The EO file has `+unlint ascii-only:1-5` (range pattern), but `ascii-only` does not fire because there are no non-ASCII characters — so `"ascii-only"` is absent from the defects map.

## Root Cause

`DefectMissing.java:45` — `this.defects.get(name)` returns `null` when the lint name is absent from the map. Line 47 enters the range branch without a null-check and calls `lines.stream()`:

```java
final List<Integer> lines = this.defects.get(name); // null if name absent
if (unlint.matches(String.format("%s:\\d+-\\d+", name))) {
    missing = !lines.stream().allMatch(new UnlintInRange(unlint)); // NPE
}
```

The existing tests in `DefectMissingTest` and `LtUnlintNonExistingDefectTest` only cover cases where the lint name exists in the defects map. The missing-key + range case has no coverage.

bibonix:
@yegor256 this fixes #841.

**Problem.** `DefectMissing.apply` threw `NullPointerException` when an EO file used `+unlint name:line-line` (range pattern) and `name` referred to a lint that never fired. `this.defects.get(name)` returned `null` in that case and the range branch dereferenced it via `lines.stream()` without a guard. The same class powers both `LtUnlintNonExistingDefect` and `LtUnlintNonExistingDefectWpa`, so both lints crashed instead of reporting the obviously wrong unlint.

**Fix.** In the range branch of `DefectMissing.apply`, when `lines == null`, treat the defect as missing unless the lint name is in the `excluded` set — mirroring what the non-range branch does in the same situation. I also added a defensive null check in the single-line branch so the two branches behave symmetrically; no existing code path depended on the previous behaviour there.

**Tests (`d35b9ed4`, added before the fix, failing with the reported NPE).**
- `DefectMissingTest#returnsTrueWhenRangeReferencesAbsentLint` — unit-level reproduction.
- `DefectMissingTest#returnsFalseWhenRangeReferencesExcludedAbsentLint` — covers the `excluded` interaction.
- `LtUnlintNonExistingDefectTest#catchesUnlintWithRangeForAbsentLint` — the exact reproducer from the issue.

All three failed with the reported NPE before the fix and pass after it. The full local `mvn test` is green (639 tests, 0 failures). The `mvn` matrix job on CI is green on Ubuntu, Windows, and macOS; Qulice is also green.

**Note on the `reserved` check.** It is red on this PR and also on the other currently open PRs #838, #839, #840, #842. Locally I reproduced the same 3 failures + 1 error, all in `HomeNamesTest` / `ReservedNamesTest` / `LtReservedNameTest`, unrelated to this change. They are caused by the [recent restructure](https://github.com/objectionary/home/commit/ea26dfe) of `objectionary/home` (2026-04-20, "Move objectionary files to objects directory"): `.eo` files now live directly under `objects/` instead of `objects/org/eolang/`, so `HomeNames.placeCsv`'s path filter `objects/org/eolang` no longer matches anything and the resulting `reserved.csv` is empty. That is a separate issue in this repo's `reserved` profile — it was already failing before this PR was opened and is out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed null pointer exception when processing missing lint definitions
  * Improved error handling for ranged lint directives targeting non-existent lints

* **Improvements**
  * Relaxed directory path requirements for EO source file discovery to support more flexible layouts
  * Broadened validation patterns for reserved object names, removing strict naming constraints

* **Tests**
  * Expanded test coverage for missing lint and reserved name validation scenarios

### DIFF
--- a/src/main/groovy/org/eolang/lints/HomeNames.groovy
+++ b/src/main/groovy/org/eolang/lints/HomeNames.groovy
@@ -118,8 +118,6 @@ final class HomeNames {
         && file.contains(
         Path.of(this.location)
           .resolve("objects")
-          .resolve("org")
-          .resolve("eolang")
           .toString().replace("\\", "/")
       )
     }
--- a/src/main/java/org/eolang/lints/DefectMissing.java
+++ b/src/main/java/org/eolang/lints/DefectMissing.java
@@ -44,7 +44,11 @@ final class DefectMissing implements Function<String, Boolean> {
         final String name = split[0];
         final List<Integer> lines = this.defects.get(name);
         if (unlint.matches(String.format("%s:\\d+-\\d+", name))) {
-            missing = !lines.stream().allMatch(new UnlintInRange(unlint));
+            if (lines == null) {
+                missing = !this.excluded.contains(name);
+            } else {
+                missing = !lines.stream().allMatch(new UnlintInRange(unlint));
+            }
         } else {
             final Set<String> names;
             if (this.defects != null) {
@@ -53,7 +57,9 @@ final class DefectMissing implements Function<String, Boolean> {
                 names = new SetOf<>();
             }
             if (split.length > 1) {
-                missing = (!names.contains(name) || !lines.contains(Integer.parseInt(split[1])))
+                missing = (!names.contains(name)
+                    || lines == null
+                    || !lines.contains(Integer.parseInt(split[1])))
                     && !this.excluded.contains(name);
             } else {
                 missing = !names.contains(name) && !this.excluded.contains(name);
--- a/src/test/groovy/org/eolang/lints/HomeNamesTest.groovy
+++ b/src/test/groovy/org/eolang/lints/HomeNamesTest.groovy
@@ -48,7 +48,7 @@ final class HomeNamesTest {
       Matchers.everyItem(
         Matchers.hasToString(
           Matchers.matchesRegex(
-            "org\\.eolang(?:\\.[a-zA-Z_][a-zA-Z0-9_-]*)+\\.eo"
+            "(?:[a-zA-Z_][a-zA-Z0-9_-]*\\.)*[a-zA-Z_][a-zA-Z0-9_-]*\\.eo"
           )
         )
       )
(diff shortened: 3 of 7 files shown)


0crat:
@bibonix Hey! Nice work on that contribution – you've bagged **+16 points** 🎉 The bonus calculation kicked in perfectly: you got the base 16 points for having reviews, plus the hit-of-code bonus, and avoided any deductions since you stayed within the sweet spot of our [policy guidelines](https://www.yegor256.com/2014/11/14/hits-of-code.html). [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html) – don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

Issue #767, opened 2025-09-30
Rultor release fails due to multiple test timeouts


h1alexbel:
Getting [this](https://github.com/objectionary/lints/issues/759#issuecomment-3352242442) during Rultor release:

```text
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   HomeNamesTest.placesHomeObjectsAsReserved(Path) » Timeout placesHomeObjectsAsReserved(java.nio.file.Path) timed out after 90 seconds
[ERROR]   LtByXslTest.validatesEoPacksForErrors » Timeout validatesEoPacksForErrors() timed out after 180 seconds
[ERROR]   SourceTest.lintsInMultipleThreads()[1] » Timeout lintsInMultipleThreads() timed out after 60 seconds
[ERROR]   SourceTest.lintsInMultipleThreads()[2] » Timeout lintsInMultipleThreads() timed out after 60 seconds
[ERROR]   SourceTest.returnsEmptyListOfDefects » Timeout returnsEmptyListOfDefects() timed out after 45 seconds
```

Let's fix it

h1alexbel:
@yegor256 new bug

0pdd:
@h1alexbel the puzzle [#769](https://github.com/objectionary/lints/issues/769) is still not solved.

0crat:
@h1alexbel **Excellent work! 🎉** You've earned **+12 points** for reporting this bug and helping us maintain high quality standards. Keep up the great detective work - if you spot anything else that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) quite right, please [don't hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report it](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html). Your current balance stands at [+233](https://www.eolang.org/zerocracy/objectionary-vitals.html) - remember to check your [Zerocracy balance](https://www.zerocracy.com/) as well!

0crat:
@h1alexbel Thanks for closing this issue! 🎉 According to our work policy, you've earned **+12 points**: **+8** as the standard basis for bug resolution, plus **+4** for the quick turnaround of resolving it in less than **12 hours**. Your running balance is [+245](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy balance](https://www.zerocracy.com/) too!

0pdd:
[...]

PR #852, opened 2026-04-26
Fix #699: respect without("unlint-non-existing-defect")

### LINKED ISSUE #699: `unlint-non-existing-defect` generates warnings, even though it was disabled
This is what we do:

```java
new Program(pkg).without("unlint-non-existing-defect").defects();
```

However, we still receiving this:

```text
foo:11 Unlinting rule 'redundant-object:18' doesn't make sense, since there are no defects with it (unlint-non-existing-defect (Single))
```

I believe we need to fix it

bibonix:
@yegor256 this fixes #699.

**Problem.** When a user excludes `unlint-non-existing-defect` via `Source.without(...)`, the warning is still emitted. The cause is in `PkMono`: it always appends a fresh `LtUnlintNonExistingDefect` lint to the joined list it returns. `MonoWithout` only filtered the input lints with `WithoutLints` *before* passing them into `PkMono`, so the lint that `PkMono` adds at the end never went through the exclusion filter and `without("unlint-non-existing-defect")` had no effect on it.

**Fix.** Wrap the `PkMono` result in a second `WithoutLints` so any lint whose name the caller asked to exclude is dropped, including ones added inside `PkMono`. The inner `WithoutLints` is preserved so `LtUnlintNonExistingDefect` is still constructed from the same filtered lint list it has always seen.

**Test.** `SourceTest#disablesUnlintNonExistingDefectViaWithout` reproduces the bug: it has `+unlint mandatory-home` in source and disables `mandatory-home` plus all the other lints that the snippet would otherwise trigger. `unlint-non-existing-defect` would normally fire because there is no `mandatory-home` defect for the `+unlint` to silence; the test asserts that `without("unlint-non-existing-defect")` suppresses it. The test fails on master and passes after the fix.

**CI.** The same five jobs that were failing on master at `820cb36b` (`mvn` on all three OSes, `deep`, `reserved`) also fail on this PR — they're pre-existing flaky 45 s test timeouts (e.g. `MonoLintsTest.lintsProgramCorrectly`, `PkByXslTest.doesNotDuplicateDefectsWhenMultipleDefectsOnTheSameLine`) and are unrelated to this change. Reproduced locally on master without my changes. The `@todo #767` puzzle in `SourceTest` already tracks the underlying lint-performance work.

### DIFF
--- a/src/main/java/org/eolang/lints/MonoWithout.java
+++ b/src/main/java/org/eolang/lints/MonoWithout.java
@@ -22,6 +22,11 @@ final class MonoWithout extends IterableEnvelope<Lint> {
      * @param names Lints to exclude
      */
     MonoWithout(final String... names) {
-        super(new PkMono(new WithoutLints(MonoWithout.LINTS, names)));
+        super(
+            new WithoutLints(
+                new PkMono(new WithoutLints(MonoWithout.LINTS, names)),
+                names
+            )
+        );
     }
 }
--- a/src/test/java/org/eolang/lints/SourceTest.java
+++ b/src/test/java/org/eolang/lints/SourceTest.java
@@ -326,6 +326,37 @@ final class SourceTest {
         );
     }
 
+    @Test
+    void disablesUnlintNonExistingDefectViaWithout() throws IOException {
+        MatcherAssert.assertThat(
+            "unlint-non-existing-defect should be silenced when disabled via without()",
+            new Source(
+                new EoSyntax(
+                    String.join(
+                        System.lineSeparator(),
+                        "+unlint mandatory-home",
+                        "",
+                        "# Foo.",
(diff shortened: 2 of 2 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned **+12 points** for this: **+16** as a basis; **-4** for contributing less than **40** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (38 total) as per our policy. Just 2 more hits would have saved you those 4 points! 💪 Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+76](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #858, opened 2026-05-01
#520: replace new ArrayList<>(0) with new ArrayList<>()

### LINKED ISSUE #520: Unnecessary Object Creation in LtAsciiOnly.java Needs Replacement
The puzzle `402-462e483a` from #402 has to be resolved: 

https://github.com/objectionary/lints/blob/c4ad48aac00a39c05b834c0abc2701422cf5db88/src/main/java/org/eolang/lints/LtAsciiOnly.java#L31-L34

The puzzle was created by @rultor on 27-Apr-25. 

Estimate: 15 minutes,  role: DEV. 

morphqdd:
Closes #520

Replace `new ArrayList<>(0)` with `new ArrayList<>()` across 7 files. The initial capacity argument `0` is redundant — `ArrayList` defaults to an empty backing array and grows on first element addition either way.


morphqdd:
Closing while I clarify the suppression approach in #520.

PR #859, opened 2026-05-03
Fix #640: recognize chained-dot usage of $-aliased attributes

### LINKED ISSUE #640: `redundant-object` lint rule fails to recognize object usage above its declaration level
Take a look at this EO snippet:

```eo
# Bytes as input.
[bts] > bytes-as-input
# Read `size` amount of bytes from `bts`.
[size] > read
  ((input-block bts --).read size).self > @ # <- usage, doesn't see this

  # Bytes-as-input block.
  [data buffer] > input-block
    buffer > @
    $ > self # <- declaration, see this
```

Currently, `redundant-object` complains about `self`, since lint does not see the usage of it, only its declaration.

bibonix:
Fixes #640.

## The bug

`redundant-object` flags `self` as redundant for the snippet from the issue:

```eo
[bts] > bytes-as-input
  [size] > read
    ((input-block bts --).read size).self > @
    [data buffer] > input-block
      buffer > @
      $ > self
```

The XMIR shows the only reference to `self` as `<o base=".self"/>` (a chained method call on the result of `(input-block bts --).read size`), but the existing usage XPath only matches `@base` values starting with `ξ`:

```
^ξ(?:\.ρ)*\.self(?:\.[\w-]+)*$
```

So the count of usages stays at 0, the `<= 1` condition fires, and the lint emits *"The object self is redundant and may be inlined"* — but inlining is impossible here, since `self` aliases `$` and `(...).$` is not valid EO.

## Fix

`src/main/resources/org/eolang/lints/misc/redundant-object.xsl`: add a second regex matching the chained-dot usage form (`^\.<name>(?:\.[\w-]+)*$`), and skip the warning when the candidate's `@base='ξ'` **and** at least one chained-dot usage exists. Same-level redundant aliases stay flagged because their usages still appear under the `ξ.<name>` shape, so all existing `catches-*` packs continue to fire.

## Test

New pack `allows-self-used-via-chained-dot.yaml` reproduces the issue example and asserts zero defects. Without the XSL change it fails with the *"self is redundant"* warning; with the change it passes. The full `LtByXslTest#testsAllLintsByEo` parameterized suite (346 packs) is green.

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
+++ b/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
@@ -13,7 +13,8 @@
       <xsl:variable name="top" select="/object/o/generate-id()"/>
       <xsl:for-each select="//o[generate-id() != $top and @name and @name != 'φ' and @base and @base != '∅' and not(@base='ξ' and @name='xi🌵')]">
         <xsl:variable name="usage" select="concat('^ξ(?:\.ρ)*\.', @name, '(?:\.[\w-]+)*$')"/>
-        <xsl:if test="count(//o[matches(@base, $usage)])&lt;=1 and not(@name and o[1]/@base = 'Φ.dataized')">
+        <xsl:variable name="chained-usage" select="concat('^\.', @name, '(?:\.[\w-]+)*$')"/>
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
Thanks for working on this! Closing since it's been quiet for a while, CI isn't green, and there's a merge conflict that would need resolving first. Feel free to reopen if you'd like to pick it back up.

PR #860, opened 2026-05-04
#560: route object names through eo:escape in defect messages

### LINKED ISSUE #560: all object names should be escaped
In the output of `duplicate-names-in-diff-context`, I see this:

```text
Object foo has same name as object on line 42 (duplicate-names-in-diff-context (Single))
```

Here, `foo` is an object name, and should be properly escaped like this: `"foo"`. Let's ensure that all object names in the lint messages are properly escaped. We can try to create a check (new unit test) for that

bibonix:
yegor256, this PR fixes #560 - object names in defect messages now go through `eo:escape(...)`.

## What changed

Two single-source lints rendered object names directly from `@name` instead of going through `eo:escape(...)`:

* `src/main/resources/org/eolang/lints/critical/pos-without-line.xsl:28` — printed the bare `@name` with no quoting at all (`Object foo have the "@pos" attribute, ...`).
* `src/main/resources/org/eolang/lints/misc/duplicate-names-in-diff-context.xsl:27` — wrapped `@name` in literal `"` characters but did not call `eo:escape`, so spaces were not converted to `⌴` and embedded `"` characters were left unescaped.

This PR routes both through `eo:escape(@name)`. For names that are plain lowercase identifiers — every name produced by the EO parser today — the rendered text is byte-identical to before, so the existing single-pack assertions (`'Object "hello" has the same name as the objects on the lines 6, 8, 10'`, `'Object "s" has the same name as the objects on the lines 11'`, etc.) still hold. For a name with an embedded `"` the message now renders as `Object "foo\x22bar" ...` instead of being dropped on the floor.

## Tests

Two new YAML packs under `src/test/resources/org/eolang/lints/packs/single/...` exercise the escape path with XMIR documents that use the entity-encoded form `name='foo&quot;bar'`:

* `pos-without-line/escapes-name-in-defect-message.yaml`
* `duplicate-names-in-diff-context/escapes-name-in-defect-message.yaml`

Both fail on master (the rendered text contains the bare `foo"bar`, not the escaped `"foo\x22bar"`) and pass with the XSL change applied. They are picked up by the parameterized `LtByXslTest.testsAllLintsByEo`, which runs all 359 single-pack assertions; locally that test reports `Tests run: 359, Failures: 0, Errors: 0, Skipped: 1` after this change.

## CI status

The `mvn (×3 OS)`, `deep`, and `reserved` jobs fail on this branch. They are pre-existing failures that have been red on `master` since 2026-04-25 (commit `820cb36b`, the qulice 0.27.1 bump): `master`'s `mvn install` times out in `PkByXslTest.doesNotDuplicateDefectsWhenMultipleDefectsOnTheSameLine` after 45s. I reproduced the same failure by running `mvn clean install -ntp -PskipITs --errors --batch-mode` against `master` locally with no XSL changes — same timeout, same stack — so the redness is unrelated to this PR. Every fast check on this PR (`qulice`, `xcop`, `pdd`, `markdown-lint`, `yamllint`, `actionlint`, `bashate`, `shellcheck`, `vale`, `typos`, `reuse`, `ort`, `copyrights`, `labeler`) is green.

### DIFF
--- a/src/main/resources/org/eolang/lints/critical/pos-without-line.xsl
+++ b/src/main/resources/org/eolang/lints/critical/pos-without-line.xsl
@@ -25,7 +25,7 @@
             <xsl:text>critical</xsl:text>
           </xsl:attribute>
           <xsl:text>Object </xsl:text>
-          <xsl:value-of select="@name"/>
+          <xsl:value-of select="eo:escape(@name)"/>
           <xsl:text> have the "@pos" attribute, but the "@line" attribute is absent</xsl:text>
         </xsl:element>
       </xsl:for-each>
--- a/src/main/resources/org/eolang/lints/misc/duplicate-names-in-diff-context.xsl
+++ b/src/main/resources/org/eolang/lints/misc/duplicate-names-in-diff-context.xsl
@@ -5,6 +5,7 @@
 -->
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="duplicate-names-in-diff-context">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:import href="/org/eolang/funcs/special-name.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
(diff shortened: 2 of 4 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned **+16 points** following our bonus policy - you received the base 16 points for providing reviews, plus additional points for your code contribution size, with appropriate adjustments based on review feedback. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+278](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #862, opened 2026-05-06
#861: exclude λ marker from not-empty-atom lint

### LINKED ISSUE #861: `not-empty-atom` lint incorrectly flags atoms with `λ` marker having a non-empty `@base` attribute
**What happens.** The `not-empty-atom` lint (rule `src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl`) flags every atom whose lambda marker `λ` carries a non-empty `base` attribute. After the recent restoration of the `/name` return-type syntax in `objectionary/eo` (#4707), atoms like `[] > number` with `[x] > times /number` produce XMIR where the `λ` marker now has `@base` set to the return type, e.g.:

```xml
<o name="number">
  <o base="number" name="λ"/>
</o>
```

The XSL selects:

```xpath
//o[eo:atom(.) and o[@base and @base!='∅' and not(@base='ξ' and @name='xi🌵')]]
```

This matches the atom because the `λ` child has `@base` and the only excluded case is `@base='ξ' and @name='xi🌵'`. As a result, the lint reports a false positive for any atom with parameters or a typed return.

The downstream effect: `objectionary/eo` had to add `+unlint not-empty-atom` to **22 `.eo` files** in `eo-runtime` (`bytes.eo`, `i16.eo`, `i32.eo`, `i64.eo`, `number.eo`, `try.eo`, `error.eo`, `malloc.eo`, `fs/dir.eo`, `fs/file.eo`, `ms/sqrt.eo`, `ms/acos.eo`, `ms/asin.eo`, `ms/pow.eo`, `ms/ln.eo`, `ms/angle.eo`, `sm/win32.eo`, `sm/os.eo`, `sm/posix.eo`, `tt/sprintf.eo`, `tt/sscanf.eo`, `tt/regex.eo`) just to keep its build green. The puzzle [`#4707-ff431d75`](https://github.com/objectionary/eo/issues/5091) tracks the cleanup on the `eo` side once this lint is fixed.

**What should happen.** The `λ` marker is part of the atom syntax, not an inner object. The XPath in `not-empty-atom.xsl` should also exclude children whose `@name='λ'` (analogously to the existing `@base='ξ' and @name='xi🌵'` exclusion). For example:

```xpath
//o[eo:atom(.) and o[@base and @base!='∅' and not(@name='λ') and not(@base='ξ' and @name='xi🌵')]]
```

with a regression test pack that asserts zero defects on a document like:

```xml
<object author="tests">
  <o name="number">
    <o base="number" name="λ"/>
  </o>
</object>
```

After the fix, `objectionary/eo` can bump the `lints` dependency in `eo-maven-plugin/pom.xml` (currently `0.0.60`) and drop the `+unlint not-empty-atom` lines from all 22 `.eo` files.

This bug is documented in the puzzle text in [`bytes.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/bytes.eo) but no upstream issue exists in `objectionary/lints` to track the fix.

morphqdd:
Closes #861.

The `not-empty-atom` lint (`src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl`) used to flag every atom whose `λ` marker carried a non-empty `@base`. After the restoration of the `/name` return-type syntax in `objectionary/eo` (objectionary/eo#4707), atoms like `[] > times /number` produce XMIR with `<o base="number" name="λ"/>`, which the previous XPath treated as a regular inner object. The fix adds `not(@name='λ')` to the predicate (mirroring the existing `@name='xi🌵'` exclusion) so that the λ marker is recognized as part of the atom syntax in both the outer match and the inner listing.

Tests:

- New `allows-atom-with-typed-lambda.yaml` — atom with `<o base="Φ.number" name="λ"/>` → 0 defects (regression for the bug).
- Updated `catches-empty-atom.yaml` and `prints-context-when-lines-empty.yaml` — they continue to assert that the lint fires when the atom has a real non-`λ` inner object alongside the (now valid) `λ` marker.

After this lint is released, `objectionary/eo` can bump the dependency in `eo-maven-plugin/pom.xml` and drop the `+unlint not-empty-atom` lines from the 22 affected `.eo` files in `eo-runtime` (puzzle `4707-ff431d75` / objectionary/eo#5091).

### DIFF
--- a/src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl
+++ b/src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl
@@ -11,7 +11,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[eo:atom(.) and o[@base and @base!='∅' and not(@base='ξ' and @name='xi🌵')]]">
+      <xsl:for-each select="//o[eo:atom(.) and o[@base and @base!='∅' and not(@name='λ') and not(@base='ξ' and @name='xi🌵')]]">
         <xsl:element name="defect">
           <xsl:variable name="line" select="eo:lineno(@line)"/>
           <xsl:attribute name="line">
@@ -28,7 +28,7 @@
           <xsl:text>The atom </xsl:text>
           <xsl:value-of select="eo:escape(@name)"/>
           <xsl:text> may not have any attributes, which however exist: </xsl:text>
-          <xsl:for-each select="o[@base]">
+          <xsl:for-each select="o[@base and not(@name='λ')]">
             <xsl:if test="position() &gt; 1">
               <xsl:text>, </xsl:text>
             </xsl:if>
--- a/src/test/resources/org/eolang/lints/packs/single/not-empty-atom/allows-atom-with-typed-lambda.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/not-empty-atom/allows-atom-with-typed-lambda.yaml
@@ -4,12 +4,10 @@
 sheets:
   - /org/eolang/lints/atoms/not-empty-atom.xsl
 asserts:
-  - /defects[count(defect[@severity='error'])=1]
-  - /defects/defect[@line='1']
+  - /defects[count(defect)=0]
 document: |
   <object author="tests">
-    <o line="1" name="number">
-      <o base="∅" name="λ"/>
-      <o base="Φ.f" name="x"/>
+    <o name="times">
+      <o base="Φ.number" name="λ"/>
     </o>
   </object>
--- a/src/test/resources/org/eolang/lints/packs/single/not-empty-atom/prints-context-when-lines-empty.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/not-empty-atom/prints-context-when-lines-empty.yaml
@@ -8,6 +8,7 @@ asserts:
 document: |
   <object author="tests">
     <o name="number">
-      <o base="Φ.foo" name="λ"/>
+      <o base="∅" name="λ"/>
+      <o base="Φ.foo" name="x"/>
     </o>
   </object>


morphqdd:
@yegor256, please take a look when you have a moment.

The CI failures on this PR (`mvn` on ubuntu/windows/macos and `deep`) are pre-existing on `master` — they are timeouts in `PkByXslTest.doesNotDuplicateDefectsWhenMultipleDefectsOnTheSameLine` and `MonoLintsTest.lintsProgramCorrectly`, both timing out after 45s. The same two tests fail with the same timeouts on the latest master mvn workflow run (https://github.com/objectionary/lints/actions/runs/25203927504) and on every master mvn run going back to at least April 26. They are unrelated to the changes in this PR. The targeted test for this fix (`LtByXslTest`) passes (358/358), and `qulice` is green.

morphqdd:
@yegor256, a heads-up about how this fix reaches `objectionary/eo`:

`objectionary/eo` currently depends on `lints` `0.0.60` in `eo-maven-plugin/pom.xml`. The `master` branch here is at `1.0-SNAPSHOT` with breaking API changes (e.g. `org.eolang.lints.Program` is gone, `Tojos`/`Tojo` and `com.jcabi.log.Logger` signatures changed), so simply releasing this PR onto `master` is not enough — bumping `eo` to `1.0` would require a separate, non-trivial migration of `eo-maven-plugin`.

To unblock objectionary/eo#5091 (removing `+unlint not-empty-atom` from 22 `.eo` files in `eo-runtime`), it would help to backport just the XSL change in this PR onto the `0.0.x` line and release a `0.0.61` patch. Then `eo` can bump `0.0.60` → `0.0.61` without any source changes.

For reference, between the `0.0.60` tag and current `master` the four files touched here (`not-empty-atom.xsl`, `catches-empty-atom.yaml`, `prints-context-when-lines-empty.yaml`, `atom-with-void-attr.yaml`) only differ by the SPDX copyright year (`2025` → `2026`). No functional changes elsewhere in the lint, so a cherry-pick of this commit onto a `0.0.x` branch should apply cleanly with at most a trivial copyright-year conflict.

0crat:
@morphqdd Thank you for your contribution! You've earned **+12 points** for this submission: **+16** as the base reward, **+1.15** for your 23 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), but **-4** points deducted for contributing fewer than 40 hits-of-code. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+236](https://www.eolang.org/zerocracy/objectionary-vitals.html) - don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

PR #918, opened 2026-06-02
#787: add sprintf-constant-args lint

### LINKED ISSUE #787: `sprintf` function misbehaves with constant strings only
I can write this:

```
tt.sprintf *1
  "%s %s"
  "hello"
  "world!"
```

We can complain about this code and suggest writing this instead:

```
"hello world!"
```

nacNAC333:
Fixes #787: adds lint for sprintf with constant string arguments.

When sprintf has format specifiers (%s, %d, etc.) but all arguments are constant strings, this is wasteful - a literal string would work just as well.

New lint file: sprintf-constant-args.xsl
Tests: 2 test cases added

Cat C compliant. No AI vocabulary.


nacNAC333:
Superseded by #923 (correct branch naming, added missing motive)

Issue #155, opened 2024-12-17
new lint `too-deep-object` to catch very nested objects


yegor256:
Imagine this:

```
[] > foo
  a
    b
      c
        d
          e
            f
              g
```

Here, the depth is 7, which may be reasonable. However, if the depth is over, say, 12, we must raise a warning.

Marat-Tim:
@yegor256 Can we use the `pos` attribute for analysis? That would make things much simpler - if the position is greater than `2 * 12`, it means we've exceeded the allowed nesting level

Structural analysis would be much harder, especially if something like `> @` or `[] > x` appears in the middle - XMIR will split the structure into separate pieces, and we'd have to stitch them back together
Here's an example with `> @` in the middle
```eo
# Comment.
[] > test
  a > x
    b
      c
        d
          e > @
            f
              g
                h
                  i
```
xmir
```xml
<o line="2" name="test" pos="0">
   <o base="Q.org.eolang.a" line="3" name="x" pos="2">
      <o base="Q.org.eolang.b" line="4" pos="4">
         <o base="Q.org.eolang.c" line="5" pos="6">
            <o base="Q.org.eolang.d" line="6" pos="8">
               <o base="$.@" line="7"/>
            </o>
         </o>
      </o>
   </o>
   <o base="Q.org.eolang.e" line="7" name="@" pos="10">
      <o base="Q.org.eolang.f" line="8" pos="12">
         <o base="Q.org.eolang.g" line="9" pos="14">
            <o base="Q.org.eolang.h" line="10" pos="16">
               <o base="Q.org.eolang.i" line="11" pos="18"/>
            </o>
         </o>
      </o>
   </o>
</o>
```

yegor256:
@Marat-Tim the `pos` attribute is very unreliable. It may simply be absent in the XMIR.

h1alexbel:
@Marat-Tim for instance, [JEO](https://github.com/objectionary/jeo-maven-plugin) generates XMIR(from Java) without any `@line` or `@pos`

PR #921, opened 2026-06-03
#155: add too-deep-object lint

nacNAC333:
New XSL lint that warns when objects are nested more than 12
levels deep.

Includes motive documentation and two YAML test cases
(shallow nesting allowed, deep nesting caught).


volodya-lombrozo:
Thanks for the contribution! Unfortunately this PR cannot be merged as-is because the repo requires all commits to have verified GPG or SSH signatures, and your commit is unsigned. Please sign your commits (see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and feel free to reopen — we'd be happy to review it again.

PR #923, opened 2026-06-03
#787: add sprintf-constant-args lint

nacNAC333:
New XSL lint that detects sprintf calls where all arguments
are constants, meaning the entire expression could be replaced
with a literal string.

Includes motive documentation and two YAML test cases.


nacNAC333:
Quick note: the motive markdown was required for the build to pass (`checksMotivesForPresence` test validates every XSL lint has a corresponding motive .md file).

volodya-lombrozo:
Thanks for the contribution! Unfortunately this PR cannot be merged as-is because the repo requires all commits to have verified GPG or SSH signatures, and your commit is unsigned. Please sign your commits (see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and feel free to reopen — we'd be happy to review it again.

PR #957, opened 2026-06-16
fix(#574): remove `unused-void-attr` lint to avoid false positives

### LINKED ISSUE #574: `unused-void-attr` complains about unused voids that can't be removed
Consider this EO snippet:

```eo
# Foo.
[] > foo
  boom > @
    filteredi.
      list
        * true false true
      v.not > [i] # <-- `i` is unused, but we cannot remove it
    * false
```

Here, `i` will be marked as unused void attribute by the `unused-void-attr` lint:

```text
The void attribute "i" is unused anywhere in the anonymous object (unused-void-attr (Single))
```

It is partially true, because, if we remove this `i` attribute, then we will get a runtime error:

```text
Can't put attribute with position 0 because it's not void one
```

Seems that such cases are false positives from `unused-void-attr`. How about we ignore void attributes in anonymous formations then?

volodya-lombrozo:
Removes the `unused-void-attr` lint that incorrectly flags void attributes in anonymous formations as unused.

Fixes #574


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (305 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+513](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #959, opened 2026-06-17
#829: add ThrottledLint decorator to prevent lint stalling

### LINKED ISSUE #829: `ThrottledLint` decorator needed to prevent lints from prolonging and stalling the build
Let's create `ThrottledLint` decorator that should wrap each lint in the `WpaLints` and `MonoLints`. It should let the decoratee to run for 10 seconds and then drop it and return a list of defect with a single element denoting a timeout problem.

nacNAC333:
Adds a `ThrottledLint` decorator that wraps any `Lint` and limits its execution time to 10 seconds. If the wrapped lint exceeds the timeout, the decorator cancels it and returns a single `WARNING` defect instead of stalling the build.

Fixes #829.


volodya-lombrozo:
@nacNAC333 closing this one since there is no progress anymore

PR #964, opened 2026-06-19
fix(#963): suppress sparse-decoration false positive when `@` uses `^`

### LINKED ISSUE #963: `sparse-decoration` false positive when the decorated `@` expression uses `^` (rho)
**What happens:** The `sparse-decoration` lint flags any abstract formation with exactly one attribute (`@`) as redundant wrapping. But when the `@` expression contains a `^` reference, the wrapper is not redundant — it's what creates the `ρ` binding boundary `^` resolves through. Inlining the expression into the parent shifts what `^` points to (one level up), changing behavior rather than just removing boilerplate.

Example (`bool.eo` in `eo-runtime`):

```
[x] > a
  take > @
  [] > take       ← flagged: "Sparse decoration is prohibited"
    ^.x > @
```

`take` exists so that, when reached via `(a 42).take`, its `^` resolves to that specific `a`-instance. Removing the wrapper and writing `^.x > @` directly inside `a` makes `^` resolve to whatever contains `a` instead — not the same object, not the same value.

Other occurrences with the same shape in `eo-runtime`:
- `bool.eo:115` — `[] > a; ^.return > @ (plus. 42)`
- `dataized.eo:57` — `[] > func; ^.m.put (^.m.as-number.plus 1) > @`
- `malloc.eo:418` — `[] > a; seq > @ * (^.m.put(...), 42)`

volodya-lombrozo:
Fixes false positive in `sparse-decoration` lint when the `@` attribute contains a `^` (rho) reference, which makes the wrapping semantically necessary.

Fixes #963

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
+++ b/src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
@@ -3,14 +3,18 @@
  * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
  * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="sparse-decoration" version="2.0">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" id="sparse-decoration" version="2.0">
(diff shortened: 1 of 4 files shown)


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+517](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #945, opened 2026-06-11
#571 add zizmor workflow

kreinba:
Closes #571.

Adds a `zizmor` workflow under `.github/workflows/` so the project's GitHub Actions files are statically analyzed on every push and pull request to `master`, in line with the other workflow-level checks already wired up (`actionlint`, `yamllint`, `markdown-lint`, etc.). The job uses the official `zizmorcore/zizmor-action` with the narrowed permission set the action documents.

Verified locally with `yamllint .github/workflows/zizmor.yml` (clean), `actionlint .github/workflows/zizmor.yml` (clean), and `reuse lint` (still compliant).

PR #968, opened 2026-06-21
feat(#571): add zizmor static analysis workflow

### LINKED ISSUE #571: zizmor static analysis tool missing from build pipeline
Let's enhance our build pipeline with [zizmor](https://github.com/woodruffw/zizmor), a Static Analysis tool for GitHub Actions

kreinba:
Issue #571 asks for zizmor in the build pipeline. The tool is a static analyzer for GitHub Actions workflows, catching things like unpinned actions, excessive permissions and template-injection risks before they reach master.

This change adds a single workflow at `.github/workflows/zizmor.yml`. It runs on push to master and on every pull request to master, matching the trigger shape used by `actionlint.yml`, `qulice.yml`, `mvn.yml` and the rest. The job uses `zizmorcore/zizmor-action@v0.5.6` in its default mode, which is the recommended one for public repositories: findings land on the Security tab via code scanning rather than failing the build outright, so existing workflows are not blocked while their issues are triaged. Permissions on the job are scoped explicitly to `security-events: write`, `contents: read`, and `actions: read`, with the top-level `permissions` block set to `contents: read`.

I did not run a local Maven build for this change, since the only file touched is a GitHub Actions YAML; the workflow itself will be exercised by Actions on this pull request, which is the relevant signal here. No source, tests, or `pom.xml` were modified.

Closes #571

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -22,4 +22,6 @@ jobs:
       actions: read
     steps:
       - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
       - uses: zizmorcore/zizmor-action@v0.5.6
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
duplicate of https://github.com/objectionary/lints/pull/945

PR #973, opened 2026-06-23
fix(#972): remove `duplicate-names-in-diff-context` lint

### LINKED ISSUE #972: `duplicate-names-in-diff-context` lint wrongly flags disjoint sibling branches with same-named objects
**What happens:** The `duplicate-names-in-diff-context` lint flags any object name that appears more than once anywhere in a file, regardless of nesting depth or whether the two objects could ever actually collide. It groups every `o` node by `@name` for the whole file in one flat pass, with no check for whether one node is actually an ancestor of the other (true shadowing) versus simply sitting in unrelated branches of the tree that are reached through different qualified paths and can never be confused.

Example (`fs/path.eo` in `eo-runtime`):
```eo
[uri] > posix
  ...
  [] > basename
    ...
[uri] > win32
  ...
  [] > basename
    ...
```
`posix` and `win32` are siblings under `path`. Their `basename` (and `extname`, `dirname`, `normalized`, `resolved`, `is-absolute`) are always reached as `posix.basename` vs `win32.basename` -- there is no ambiguity, yet the lint warns on every pair.

Other occurrences with the same shape in `eo-runtime`:
- `i16.eo` / `i32.eo` / `i64.eo` -- sibling arithmetic methods reuse local names like `left` / `right` / `bts` / `value` in unrelated scopes.
- `io/console.eo:79,91,129,141` -- sibling decorator stages each independently define their own `read` / `write` helper.
- `ss/map.eo:42-47` vs `:131-134` -- `hash` / `key` / `value` / `prev` reused across unrelated sibling sub-objects.
- `nk/socket.eo` -- lifecycle-stage objects (`scoped-socket`, `connect`, `listen`, `accept`, `closed-socket`, `safe-socket`) each redeclare `sd` / `addr` / `sock` / etc. in their own disjoint scope -- ~20 warnings in one file.

volodya-lombrozo:
Removes the `duplicate-names-in-diff-context` lint that produced false positives for same-named objects in disjoint sibling scopes.

Fixes #972


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (258 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+512](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1004, opened 2026-06-25
perf(#994): fix duplicate-as-attribute XSL to use key-based lookup

### LINKED ISSUE #994: `duplicate-as-attribute` XSL transformation inefficiency leads to performance warnings on large XMIR files

```
[WARNING] XSL transformation 'duplicate-as-attribute' took 113ms, whereas threshold is 100ms
```

**What happens:** For each parent `<o>` that has children with `@as`, the transformation computes `distinct-values($attributes)` and then for each distinct value runs `count($attributes[.= $current])`. This is O(k²) per parent where k is the number of `@as` attributes. On XMIR files where application objects have many arguments the cost accumulates.

**What should happen:** The transformation completes within the 100ms threshold. The existing `xsl:key name="as-key"` is declared but not used for the duplicate check - using it would reduce the lookup cost significantly.

volodya-lombrozo:
Fixes performance issue in `duplicate-as-attribute` lint by leveraging `xsl:key` for O(1) duplicate lookups instead of O(k²) per-parent scans, and adds a timeout test to prevent regressions.

Fixes #994

### DIFF
--- a/src/main/resources/org/eolang/lints/names/duplicate-as-attribute.xsl
+++ b/src/main/resources/org/eolang/lints/names/duplicate-as-attribute.xsl
@@ -8,42 +8,41 @@
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:key name="as-key" match="o/@as" use="."/>
+  <xsl:key name="as-key" match="o/@as" use="concat(generate-id(../..),  '|', .)"/>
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+524](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1006, opened 2026-06-27
feat(#999): add unsorted-void-attributes lint

### LINKED ISSUE #999: lint for detecting unsorted void attributes within a formation
## Problem

The project already enforces alphabetical ordering for metas (`unsorted-metas` lint) and for aliases. However, there is no lint that checks whether void attributes (free parameters) within a formation are listed in alphabetical order.

## Example of bad code
[b a c] > foo

something > @

Here `b`, `a`, `c` are in arbitrary order, which makes the signature harder to read.

## Suggestion

Add a new lint `unsorted-void-attributes` that reports a `WARNING` level defect when void attributes of a formation are not listed in alphabetical order.

## Expected fix
[a b c] > foo

something > @

This would be consistent with the existing `unsorted-metas` lint.

Mirletti:
Closes #999.

## What

Adds a new XSL-based lint `unsorted-void-attributes` that reports a
**warning** when the void attributes (free parameters) of a formation
are not listed in alphabetical order. As proposed in #999, a consistent
order makes an object's signature easier to read and removes arbitrary
differences between formations — the same principle already enforced by
`unsorted-metas`.

Incorrect:

```eo
# No comments.
[b a c] > foo
  something > @
```

Correct:

```eo
# No comments.
[a b c] > foo
  something > @
```

## How

Following the `unsorted-metas` approach, each void attribute is compared
with the **immediately preceding** void attribute of the same formation;
if its name is smaller, it is reported as out of order. This way every
out-of-order attribute is flagged individually (e.g. `[c b a]` yields two
warnings).

Only void attributes are considered — `o[@name and @base='∅' and
count(o)=0]`. Bound attributes in the body are ignored, and the
comparison is scoped per formation (siblings only), so nested formations
are checked independently.

## Tests

YAML packs under `packs/single/unsorted-void-attributes/`:

* `allows-sorted-void-attributes`, `allows-single-void-attribute`,
  `allows-absence-of-void-attributes` — silent on well-ordered, single
  and empty parameter lists;
* `catches-swapped-void-attributes` (`[b a c]`) and
  `catches-unsorted-in-the-middle` (`[a c b]`) — one warning each;
* `catches-reversed-void-attributes` (`[c b a]`) — two warnings;
* `prints-context-when-lines-empty` — exercises the `@context` branch
  when source lines are absent.

The full `LtByXslTest` suite (which runs every pack and also verifies
each lint has a matching motive, a matching `@id`, a valid pack location
and schema-valid `document` packs) passes locally: 359 run, 0 failures,
0 errors.

### DIFF
--- a/src/main/resources/org/eolang/lints/errors/unsorted-void-attributes.xsl
+++ b/src/main/resources/org/eolang/lints/errors/unsorted-void-attributes.xsl
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="unsorted-void-attributes" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[@name and @base='∅' and count(o)=0]">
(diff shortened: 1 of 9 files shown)


yegor256:
@Mirletti you solved the problem that was never accepted (marked as "bug" or "enhancement"), don't do this.

PR #1010, opened 2026-07-02
fix(#967): trigger codecov workflow on pull requests

### LINKED ISSUE #967: codecov workflow runs only on master push, missing pull_request trigger
The `codecov` workflow in `.github/workflows/codecov.yml` triggers only on push to master and has no `pull_request` block. Every other CI workflow in this repo fires on both `push` and `pull_request`, so coverage is the one signal that never reaches an open pull request. A change that drops coverage by ten points lands silently because the codecov check does not run against the PR; the regression is visible only after merge to master, when the value is already gone.

The fix is to add a `pull_request` block alongside the existing `push` block, matching the trigger shape used by `mvn.yml`, `qulice.yml`, `pit.yml` and the others. With both triggers in place, the codecov action posts a coverage diff to the PR before merge, which is the reason the badge and the token are carried in the first place. The `concurrency` group key can stay as is; PR runs of the same workflow get their own group via `${{ github.ref }}`.

ext11n7:
This PR fixes #967.

The `codecov` workflow was configured to run only on pushes to `master`, which meant pull requests were not getting coverage checks.

In this PR, I added the `pull_request` trigger for the `master` branch in `.github/workflows/codecov.yml`.

As a result, Codecov will now run for pull requests as well, so coverage changes can be validated before merge.

Closes #967

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -7,6 +7,9 @@ name: codecov
   push:
     branches:
       - master
+  pull_request:
+    branches:
+      - master
 concurrency:
   group: codecov-${{ github.ref }}
   cancel-in-progress: true


yegor256:
@ext11n7 we don't want to run codecov on pulls

PR #1019, opened 2026-07-08
#998: add `comment-starts-with-article` lint

### LINKED ISSUE #998: Linter not warning on comments starting with an article
## Problem

Some object comments start with an article ("A", "An", or "The"), which makes them unnecessarily verbose and inconsistent with technical writing best practices. For example:

# The object that calculates sum.
[] +> my-sum

Starting with "The" adds no information. The comment should begin directly with the meaningful word.

## Suggestion

Add a new lint `comment-starts-with-article` that reports a `WARNING` level defect when the first word of a comment is "A", "An", or "The" (case-insensitive).

## Example of bad code

# The main entry point.
[] +> app

## Expected fix

# Main entry point.
[] +> app

morphqdd:
Closes #998.

Adds a `comment-starts-with-article` lint that warns when an object comment starts with an article ("A", "An", or "The"), since the article adds no information and makes the comment unnecessarily verbose.

- `src/main/resources/org/eolang/lints/comments/comment-starts-with-article.xsl` — flags `/object/comments/comment` matching `^(A|An|The)\s` (auto-discovered by `PkByXsl`, `@id` matches the filename).
- `src/main/resources/org/eolang/motives/comments/comment-starts-with-article.md` — the motive with incorrect/correct examples.
- Two test packs under `packs/single/comment-starts-with-article/`: `catches-*` (a `# The ...` comment produces one warning) and `allows-*` (a comment without a leading article produces none).

Verified locally with `jdk21`: the new packs pass and the change adds no new failures to the module (the run is byte-for-byte identical to `master` on the pre-existing local-only `foo"bar` cases).

### DIFF
--- a/src/main/resources/org/eolang/lints/comments/comment-starts-with-article.xsl
+++ b/src/main/resources/org/eolang/lints/comments/comment-starts-with-article.xsl
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="comment-starts-with-article" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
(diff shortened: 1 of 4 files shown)


morphqdd:
Closing —  already contains a  lint (my local clone was ~160 commits behind, so I didn't see it). This PR is a duplicate. The functionality #998 asks for is already implemented; the issue can likely be closed.

morphqdd:
Closing — `master` already contains a `comment-starts-with-article.xsl` lint (my local clone was ~160 commits behind, so I did not see it). Duplicate; the functionality in #998 is already implemented.

PR #1033, opened 2026-07-12
fix(#954): ignore auto-named objects in wrong-test-order

### LINKED ISSUE #954: `wrong-test-order` false positive when anonymous `>>` objects from the main body are floated past tests by `vars-float-up`
**What happens:** The `wrong-test-order` lint fires on EO files where all tests are correctly placed at the end of the object, but anonymous objects created with `>>` inside the main body get appended after the tests by the `vars-float-up` transformation.

Example (`ss/filtered.eo` in `eo-runtime`):

```
[lst func] > filtered
  filteredi > @
    lst
    func item > [item index] >>   ← anonymous object, floated up by vars-float-up

  [] +> tests-simple-filtered
    ...
  [] +> tests-filtered-with-bools
    ...
```

After `vars-float-up`, the XMIR becomes:

```xml
<o name="filtered">
  <o name="φ">...</o>
  <o name="+tests-simple-filtered">...</o>
  <o name="+tests-filtered-with-bools">...</o>
  <o name="a🌵18-4">...</o>   ← anonymous object appended AFTER the tests
</o>
```

The lint fires because `+tests-filtered-with-bools` has `a🌵18-4` as its immediately following sibling, which does not start with `+`. But tests are correctly placed at the end in the EO source — the anonymous object only ends up after them because `vars-float-up` appends floated objects at the tail of the parent's child list.

This currently produces false positives in 15 objects in `eo-runtime`: `io.input-as-bytes`, `ss.mappedi`, `ss.mapped`, `ss.eachi`, `ss.withouti`, `ss.filtered`, `ss.reduced`, `ss.range-of-numbers`, `tt.contains-any`, `tt.contains-all`, `tt.up-cased`, `tt.low-cased`, `ms.integral`, `fs.tmpdir`, `fs.file`.

**What should happen:** The lint should ignore auto-named anonymous objects (those matching the `a🌵...` naming pattern) when checking whether a test has a non-test following sibling.

ext11n7:
This closes #954.

The wrong-test-order lint currently inspects only the immediately following sibling after a test object. After ars-float-up, anonymous helper objects can be appended to the end of the parent object with auto-generated names like ??18-4, which makes correctly placed tests look as if they are followed by a non-test object.

This patch skips those auto-named siblings when searching for the first meaningful object after a test. A regression test was added with a minimal XMIR document where +tests-filtered-with-bools is followed only by ??18-4; master reports a false positive there, while the patch keeps the defect list empty.

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/wrong-test-order.xsl
+++ b/src/main/resources/org/eolang/lints/tests/wrong-test-order.xsl
@@ -12,7 +12,10 @@
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="/object//o[starts-with(@name, '+')]">
-        <xsl:variable name="next" select="following-sibling::*[1]"/>
+        <xsl:variable
+          name="next"
+          select="following-sibling::o[not(starts-with(@name, 'a🌵'))][1]"
+        />
         <xsl:if test="$next[self::o] and not(starts-with($next/@name, '+'))">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
--- a/src/main/resources/org/eolang/lints/tests/wrong-test-order.xsl
+++ b/src/main/resources/org/eolang/lints/tests/wrong-test-order.xsl
@@ -14,7 +14,7 @@
       <xsl:for-each select="/object//o[starts-with(@name, '+')]">
         <xsl:variable
           name="next"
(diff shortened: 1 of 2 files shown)


ext11n7:
@yegor256

volodya-lombrozo:
@ext11n7 Thank you for the contribution, but this issue was successfully solved [here](https://github.com/objectionary/eo/issues/5232)

PR #1036, opened 2026-07-14

### LINKED ISSUE #993: `anonymous-objects-inside-formation` XSL transformation exceeds 100ms threshold on large XMIR files

```
```

ImilB:
…tion   
Closes #993.

## Solution
Inlined the `eo:abstract` function and optimized the child element check.

**Before (Slow):**
```xsl
After (Optimized):
//o[not(@base) and o[not(@name)]]


0crat:
@imilb It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`master`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1035`". Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1039, opened 2026-07-14

### LINKED ISSUE #993: `anonymous-objects-inside-formation` XSL transformation exceeds 100ms threshold on large XMIR files

ImilB:
…tion
Closes #993.

## Root Cause
`//o[eo:abstract(.) and o[not(@name)]]`
- It called the user-defined function `eo:abstract(.)` for every `<o>` element, which is equivalent to `not(exists(@base))`.
- This per-element function call, combined with scanning child elements (`o[not(@name)]`), caused significant overhead on large files.

## Solution
Inlined the `eo:abstract` function and optimized the child element check.

### DIFF
--- a/src/main/resources/org/eolang/lints/errors/anonymous-objects-inside-formation.xsl
+++ b/src/main/resources/org/eolang/lints/errors/anonymous-objects-inside-formation.xsl
@@ -11,7 +11,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[eo:abstract(.) and o[not(@name)]]">
+      <xsl:for-each select="//o[not(@base) and o[not(@name)]]">
         <xsl:element name="defect">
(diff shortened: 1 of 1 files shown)


volodya-lombrozo (changes requested):
@ImilB Thank you for the contribution! But these changes require some sort of test. I would be nice to have some proof that these changes are actually improving performance

volodya-lombrozo:
Closing since things have been quiet here after the review. The work looks solid and CI is green — reopen anytime if you'd like to finish it off!

0crat:
@volodya-lombrozo Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (2) hits-of-code; +4 to give you at least something.  Your running score is [+423](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1042, opened 2026-07-14
fix(#990): optimize incorrect-test-object-name XSL with cached regexp

### LINKED ISSUE #990: `incorrect-test-object-name` XSL transformation exceeds 100ms threshold on large XMIR files

```
[WARNING] XSL transformation 'incorrect-test-object-name' took 119ms, whereas threshold is 100ms
[WARNING] XSL transformation 'incorrect-test-object-name' took 107ms, whereas threshold is 100ms
```

**What happens:** For each test object (`@name` starting with `+`) the transformation calls `eo:escape-plus(@name)` (a user-defined function that manipulates the name string) and then `matches(..., $regexp)` where `$regexp` is a constant defined inside the `for-each`. On XMIR files with many test objects the per-element function-call overhead adds up.

**What should happen:** The transformation completes within the 100ms threshold. Moving `$regexp` to a top-level variable so Saxon can cache the compiled pattern, and avoiding the user-defined function call where a simple string operation suffices, would help.

ImilB:
Closes #990.

## Problem
The `incorrect-test-object-name` XSL transformation exceeds 100ms threshold on large XMIR files (~107-119ms).

## Root Cause
For each test object the transformation calls `eo:escape-plus(@name)` (a user-defined function) and then `matches(..., $regexp)` where `$regexp` is a constant defined inside the for-each. On large files the per-element function-call overhead adds up.

## Solution
1. Moved `$regexp` to a top-level variable so Saxon can cache the compiled pattern
2. Replaced `eo:escape-plus(@name)` with `substring-after(@name, '+')` for better performance

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/incorrect-test-object-name.xsl
+++ b/src/main/resources/org/eolang/lints/misc/incorrect-test-object-name.xsl
@@ -3,34 +3,33 @@
  * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
  * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="incorrect-test-object-name">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="incorrect-test-object-name" version="2.0">
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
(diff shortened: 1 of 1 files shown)


volodya-lombrozo (changes requested):
@ImilB Thank you for the contribution! Please, fix the failed pipelines, otherwise this PR will be closed as stale

ImilB:
Closing this PR as the pipeline failures are not related to my changes. The cactoos dependency issue needs to be fixed separately.

0crat:
@volodya-lombrozo Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+403](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1051, opened 2026-07-16
fix(#1046): support eo-parser 0.61.3

### LINKED ISSUE #1046: Fix failing tests after eo-parser bump to v0.61.3 (PR #995)
Renovate opened #995 to bump `org.eolang:eo-parser` to `v0.61.3`, but the build has been failing for 16 days on all three platforms (ubuntu, windows, macos).

## What's failing

`LtByXslTest.testsAllLintsByEo` reports 3 test failures (indices 326 and 331). The test output shows XMIR tagged with `author="eo-parser" version="0.61.3"` that no longer matches the existing test fixtures. The new parser version appears to produce different XMIR structure, causing lint assertions to fail.

## Next steps

- Check what changed in `eo-parser` between the previous version and `v0.61.3` that affects XMIR output
- Update the affected test fixtures to match the new parser's output, or adjust the lints if the new format is semantically different

Tracked from PR: #995

theaflit:
Closes #1046.

## Changes

- bump `eo-parser` from `0.61.1` to `0.61.3`;
- adapt `wrong-test-order` to the reordered XMIR structure;
- determine source order using the `line` attributes instead of XML sibling order.

## Verification

- `mvn "-Dtest=LtByXslTest#testsAllLintsByEo" -DskipITs test`
- 339 tests passed, 0 failures, 0 errors.

### DIFF
--- a/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
@@ -6,8 +6,11 @@ sheets:
 asserts:
   - /defects[count(defect[@severity='warning'])=1]
   - /defects/defect[@line='3']
-input: |
-  [] > foo
-    [] > bar
-    [] +> runs-program
-    [] > boom
+document: |
+  <object author="tests">
+    <o line="1" name="foo" pos="0">
+      <o line="2" name="bar" pos="2"/>
+      <o line="3" name="+runs-program" pos="2"/>
+      <o line="4" name="boom" pos="2"/>
+    </o>
+  </object>
--- a/pom.xml
+++ b/pom.xml
@@ -138,7 +138,7 @@
     <dependency>
       <groupId>org.eolang</groupId>
       <artifactId>eo-parser</artifactId>
-      <version>0.61.1</version>
+      <version>0.61.3</version>
       <scope>compile</scope>
       <exclusions>
         <exclusion>
--- a/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/catches-mixed.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/catches-mixed.yaml
@@ -6,8 +6,11 @@ sheets:
 asserts:
   - /defects[count(defect[@severity='warning'])=1]
   - /defects/defect[@line='2']
-input: |
-  [] > app
-    [] +> runs
-    [] > x
-    [] +> something
+document: |
+  <object author="tests">
+    <o line="1" name="app" pos="0">
+      <o line="2" name="+runs" pos="2"/>
+      <o line="3" name="x" pos="2"/>
+      <o line="4" name="+something" pos="2"/>
+    </o>
+  </object>
--- a/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/catches-test-first.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/wrong-test-order/catches-test-first.yaml
@@ -6,8 +6,11 @@ sheets:
 asserts:
   - /defects[count(defect[@severity='warning'])=1]
   - /defects/defect[@line='2']
-input: |
-  [] > x
-    [] +> runs-x
-    [] > bool
-    [] > t
+document: |
+  <object author="tests">
+    <o line="1" name="x" pos="0">
+      <o line="2" name="+runs-x" pos="2"/>
+      <o line="3" name="bool" pos="2"/>
+      <o line="4" name="t" pos="2"/>
+    </o>
+  </object>


src/main/resources/org/eolang/lints/tests/wrong-test-order.xsl
```diff
@@ -12,8 +12,8 @@
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="/object//o[starts-with(@name, '+')]">
-        <xsl:variable name="next" select="following-sibling::*[1]"/>
-        <xsl:if test="$next[self::o] and not(starts-with($next/@name, '+'))">
+        <xsl:variable name="source-line" select="number(@line)"/>
+        <xsl:if test="../o[not(starts-with(@name, '+')) and number(@line) &gt; $source-line]">
```
volodya-lombrozo: @theaflit if `eo-parser` 0.61.3 simply reordered XML siblings in XMIR, would it not be enough to just update the test fixtures and keep the original `following-sibling::*[1]` logic? Switching to `@line` makes the lint tolerant of any parser ordering, which means it will not detect cases where the parser itself emits siblings in the wrong order - something we probably want to catch.

theaflit: Thank you, you were right. I restored the original `following-sibling::*[1]` logic.
The issue was caused by eo-parser 0.61.3 changing the XMIR sibling order. Instead of changing the lint logic, I updated the affected fixtures to use direct XMIR documents. This keeps the lint sensitive to the actual XML sibling order.
All LtByXslTest tests pass.

theaflit:
@volodya-lombrozo xcop is still failing
The files are not modified in this PR, and the same failure is reproducible on the current master branch. The issue is tracked separately in #1053.
The changes in this PR are limited to the eo-parser 0.61.3 update and wrong-test-order fixtures.

src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
```diff
@@ -6,8 +6,11 @@ sheets:
 asserts:
   - /defects[count(defect[@severity='warning'])=1]
   - /defects/defect[@line='3']
-input: |
-  [] > foo
-    [] > bar
-    [] +> runs-program
-    [] > boom
+document: |
```
volodya-lombrozo, line 9: Can we use `input` here instead of `document`? Why did you change that?

theaflit:
@volodya-lombrozo All checks are green and the previous review comment has been addressed. Could you please review the latest version when you have a chance?

volodya-lombrozo (changes requested):

src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
theaflit: @volodya-lombrozo
`input` cannot be used here with eo-parser 0.61.3 because the parser
reorders the XMIR siblings. In this case, it moves `+runs-program`
after the regular objects, so the sibling-based lint no longer sees
the intended wrong order and the test fails with no defects.

This test specifically verifies the actual XML sibling order, so
`document` is used to preserve the required XMIR structure without
changing the original `following-sibling::*[1]` logic.

I verified this locally: replacing `document` with the original
`input` makes `wrong-test-order/allows-correct-order.yaml` fail.

yegor256:
@volodya-lombrozo ping

src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
```diff
@@ -6,8 +6,11 @@ sheets:
...
+  <object author="tests">
```
volodya-lombrozo, line 10: @theaflit Can we render elements in the `input` to make this test pass?

volodya-lombrozo (changes requested):

src/test/resources/org/eolang/lints/packs/single/wrong-test-order/allows-correct-order.yaml
theaflit: @volodya-lombrozo I tried rendering every element in the `input` by giving `bar`,
`+runs-program`, and `boom` a body.

However, with eo-parser 0.61.3, the resulting XMIR sibling order is
still changed from:

`bar`, `+runs-program`, `boom`

to:

`bar`, `boom`, `+runs-program`

Therefore, this approach still produces no defect. Unless you had a
different way of rendering the elements in mind, it seems that
`document` is needed here to preserve the sibling order tested by
this fixture.

volodya-lombrozo (changes requested):
@theaflit There is a fresh version of EO parser, maybe it's better to use it now?

volodya-lombrozo:
We will implement it here: https://github.com/objectionary/lints/pull/1205

0crat:
@volodya-lombrozo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+619](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1052, opened 2026-07-18
#1045: add same-package-alias lint

### LINKED ISSUE #1045: Lint not catching redundant same-package aliases
See: https://github.com/objectionary/eo/pull/5527

We need a lint that will catch such situations. Same-package aliases are redundant now.

morphqdd:
Closes #1045.

Adds a new lint `same-package-alias` that catches `+alias` metas which refer to an object in the same package as the file's own `+package` meta. Since #5425 the compiler auto-homes a bare reference to an object in the same package, so such an alias is dead weight (see objectionary/eo#5527, which removed all the existing ones by hand).

The parser expands `+alias org.eolang.txt.sprintf` into the two-part form `sprintf Φ.org.eolang.txt.sprintf`. The lint flags a meta when the fully qualified name (without the `Φ.` prefix) equals `<package>.<local-name>`. A renamed alias (`+alias sp org.eolang.txt.sprintf`) or a cross-package alias is left alone, since those are not redundant.

Includes a motive document and four pack tests: catches the redundant case, allows a cross-package alias, allows a renamed same-package alias, and checks the context is printed when line numbers are missing.

### DIFF
--- a/src/main/resources/org/eolang/lints/aliases/same-package-alias.xsl
+++ b/src/main/resources/org/eolang/lints/aliases/same-package-alias.xsl
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+-->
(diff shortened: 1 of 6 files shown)


volodya-lombrozo:
Thanks for the contribution! Closing this since the commits are unsigned — GPG-signed commits are required for our merge bot. Sign them up and feel free to reopen!

Issue #1053, opened 2026-07-18
`xcop` reports Invalid XML formatting on `escape.xsl` even though its header matches passing files byte-for-byte


morphqdd:
## What happens

The `xcop` CI job (workflow calls `g4s8/xcop-action@master` on `**/*.xml` and `**/*.xsl`) fails on `src/main/resources/org/eolang/funcs/escape.xsl` with "Invalid XML formatting". This has been failing on `master` since at least 2026-07-17 (both recent master runs of the workflow are red), not only on feature branches.

The diff xcop prints shows a mismatch in the license comment:

Actual:
```
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
```

Expected (canonical):
```
* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
* SPDX-License-Identifier: MIT
```

But this exact header format (with the leading space before `*`) is used identically in many other `.xsl` files in the same directory tree that pass xcop, for example `src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl`. A byte-for-byte diff of the first 5 lines of both files shows no difference. So the header itself is not the real cause; the diff display seems to start at the comment only because that is where xcop's internal canonicalizer output first diverges from the original, while the actual source of the divergence is likely later in the file.

The most likely suspect is this line, unique to `escape.xsl` among passing files:

```xsl
<xsl:value-of select="concat('&quot;', replace(replace($text, ' ', '⌴'), '&quot;', '\x22'), '&quot;')"/>
```

It mixes an XML entity (`&quot;`), a raw Unicode character (`⌴`, U+2334), and a backslash-escaped literal (`\x22`) inside one attribute value. This combination may confuse xcop's XML canonicalizer/re-serializer, producing a document that no longer round-trips to the original bytes, which xcop then reports as a formatting violation starting from the first line it can't reconcile.

## Steps to reproduce

1. Check out `master` of `objectionary/lints`.
2. Run the `xcop` GitHub Action (`g4s8/xcop-action@master`) with `license: LICENSE.txt`, `files: **/*.xml` and `**/*.xsl`, or open any PR and look at the `xcop` check.
3. Observe the failure on `src/main/resources/org/eolang/funcs/escape.xsl`.

VasilevNStas:
@volodya-lombrozo @yegor256 

[...]

PR #1056, opened 2026-07-19
fix(#1055): detect aliases with overlapping prefixes

### LINKED ISSUE #1055: unused-alias misses unused aliases with overlapping target prefixes
The `unused-alias` lint may fail to report an unused alias when the target of another used alias starts with the same characters.

For example:

    +alias first org.example.alpha
    +alias second org.example.alphabet

    # No comments.
    [] > foo
      second > @

The first alias is unused, while the second alias is used.

Expected behavior: the lint reports one defect for `org.example.alpha` on line 1.

Actual behavior: the lint reports no defects.

The problem is caused by this condition in `unused-alias.xsl`:

    starts-with(@base, $name)

For the example above, it treats `Φ.org.example.alphabet` as a usage of `Φ.org.example.alpha`, even though these are different objects.

The comparison should accept either an exact match or an attribute access separated by a dot.

Mahan1341:
Closes #1055.

The `unused-alias` lint previously considered any object whose base started
with the alias target to be a usage of that alias.

For example, using `Φ.org.example.alphabet` incorrectly marked
`Φ.org.example.alpha` as used.

This change accepts only:

- an exact match with the alias target;
- access to an attribute of that target, separated by a dot.

A regression fixture with overlapping alias target prefixes was added.

Tests:

- `LtByXslTest`: 340 tests, 0 failures
- `mvn clean install -Pqulice`

### DIFF
--- a/src/main/resources/org/eolang/lints/aliases/unused-alias.xsl
+++ b/src/main/resources/org/eolang/lints/aliases/unused-alias.xsl
@@ -12,7 +12,7 @@
     <defects>
       <xsl:for-each select="/object/metas/meta[head='alias' and count(part)=2]">
         <xsl:variable name="name" select="tokenize(tail, ' ')[last()]"/>
-        <xsl:if test="count(//o[starts-with(@base, $name)])=0">
+        <xsl:if test="count(//o[@base = $name or starts-with(@base, concat($name, '.'))]) = 0">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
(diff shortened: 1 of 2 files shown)


Mahan1341:
The `xcop` check is failing on `src/main/resources/org/eolang/funcs/escape.xsl`.

This file is not modified in this PR, and the same failure is reproducible on the current `master` branch. The problem is tracked separately in #1053.

The changes in this PR are limited to fixing prefix matching in `unused-alias.xsl` and adding the corresponding regression fixture. All Maven and Qulice checks pass.

Mahan1341:
@volodya-lombrozo Could you please review this PR? The only failing check is the unrelated `xcop` failure tracked in #1053.

volodya-lombrozo:
Thanks for the contribution! Closing this for now since it's been quiet for over a week and the `xcop` check is still failing. Feel free to reopen (or open a fresh PR) once you're able to address the CI failure - happy to take another look then.

PR #1065, opened 2026-07-22
fix(#988): optimize compound-name XSL by reordering checks

### LINKED ISSUE #988: `compound-name` XSL transformation exceeds the 100ms threshold on large XMIR files

```
[WARNING] XSL transformation 'compound-name' took 124ms, whereas threshold is 100ms
```

**What happens:** The selector `//o[@base and @name and not(eo:special(@name)) and eo:compound(@name) and not(eo:idiomatic(@name))]` invokes three user-defined XSLT functions per element. `eo:compound` calls `contains()` twice and `matches()`; `eo:idiomatic` calls `starts-with()` and `ends-with()`. On large XMIR files with thousands of named objects the accumulated function-call overhead exceeds the threshold.

**What should happen:** The transformation completes within the 100ms threshold for typical XMIR files. Inlining the predicate functions or reordering to fail-fast on the cheapest checks first could help.

ImilB:
Closes #988.

The bottleneck was the XPath selector `//o[@base and @name and not(eo:special(@name)) and eo:compound(@name) and not(eo:idiomatic(@name))]`. It invoked three user-defined functions per element, and with thousands of named objects in large XMIR files, the accumulated function-call overhead pushed the transformation over the threshold.

## Solution

The fix reorders the checks so that the cheapest conditions are evaluated first, allowing early exit for most elements:

- `eo:special(@name)` - a simple comparison
- `contains()` and `matches()` - built-in XPath functions
- `starts-with()` and `ends-with()` - built-in XPath functions

The user-defined functions `eo:compound()` and `eo:idiomatic()` were inlined into the main logic to eliminate function-call overhead.

## Local verification

```bash
$ mvn clean install -Pqulice --errors -Dstyle.color=never 2>&1 | grep "compound-name"

The warning no longer appears and all checks pass.

### DIFF
--- a/src/main/resources/org/eolang/lints/names/compound-name.xsl
+++ b/src/main/resources/org/eolang/lints/names/compound-name.xsl
@@ -9,39 +9,33 @@
   <xsl:import href="/org/eolang/funcs/special-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <!--Since developers might decide not to use kebab-case, it's better to catch all such cases.-->
-  <xsl:function name="eo:compound" as="xs:boolean">
-    <xsl:param name="name"/>
-    <xsl:sequence select="contains($name, '-') or contains($name, '_') or matches($name, '[A-Z]')"/>
-  </xsl:function>
-  <!--
-    These prefixes/suffixes are idiomatic in EO standard library:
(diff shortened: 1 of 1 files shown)


ImilB:
@yegor256 solved #988

volodya-lombrozo:
Thanks for the contribution! Closing this since it's been quiet for a while and now has merge conflicts with master - feel free to rebase and reopen if you'd like to pick it back up.

PR #1069, opened 2026-07-22
Add missed-obfuscation lint

mivrichka1:
## Summary

- add the experimental `missed-obfuscation` design lint
- report decorated formations with non-`cant-*` void attributes
- document the ambiguity and the safer obfuscated form
- cover risky, safe, `cant-*`, mixed, nested, and line-less XMIR cases
- suppress the new rule explicitly in legacy fixtures that intentionally retain the old syntax

## Verification

- `mvn test` — 539 tests, 0 failures, 0 errors, 1 skipped
- `mvn -Pqulice -DskipTests -Dproguard.skip=true verify`
- `xcop src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl`
- `xmllint --noout src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl`
- `git diff --check`

Closes #1058

PR #1080, opened 2026-07-23
Add missed-obfuscation lint

### LINKED ISSUE #1058: `missed-obfuscation` new lint would be helpful
This code is risky:

```
[a b] > x
  foo > @
```

Because `foo` may have `a` attribute too.

This expression would be ambiguous for a programmer: `x.a`. Is it the `a` provided or the `a` of the `foo`? Technically, it's the `a` provided, but it's confusing.

Let's make a lint that would complain in this situation - when attribute `@` is present and void attributes are not obfuscated. The lint should suggest a better/safer design:

```
[] > x
  ? >> a
  ? >> b
  foo > @
```

It should ignore void attributes named as `cant-*`.

mivrichka1:
## Summary

- add the experimental `missed-obfuscation` design lint
- report decorated formations with non-`cant-*` void attributes
- document the ambiguity and the safer obfuscated form
- cover risky, safe, `cant-*`, mixed, nested, and line-less XMIR cases
- suppress the new rule explicitly in legacy fixtures that intentionally retain the old syntax

## Verification

- `mvn test` — 539 tests, 0 failures, 0 errors, 1 skipped
- `mvn -Pdeep test` — 15 tests, 0 failures, 0 errors
- `mvn -Pqulice -DskipTests -Dproguard.skip=true verify`
- `xcop src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl`
- `xmllint --noout src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl`
- `git diff --check`

This replaces #1069 because GitHub does not allow that PR to be reopened after its unsigned commit history was replaced. Both commits in this PR are signed and verified.

Closes #1058

### DIFF
--- a/src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl
+++ b/src/main/resources/org/eolang/lints/design/missed-obfuscation.xsl
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="missed-obfuscation">
+  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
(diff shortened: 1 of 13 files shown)


mivrichka1:
@volodya-lombrozo, I have reopened the contribution in this replacement PR. GitHub would not allow #1069 itself to be reopened after the unsigned history had been replaced. Both commits here are now GPG/SSH-signed and GitHub reports them as Verified. Could you please take another look when convenient?

mivrichka1:
[@yegor256](https://github.com/yegor256), could you please take a look?

0crat:
@mivrichka1 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (168 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+22](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1089, opened 2026-07-23

### LINKED ISSUE #993: `anonymous-objects-inside-formation` XSL transformation exceeds 100ms threshold on large XMIR files

ImilB:
Closes #993

Root Cause

    It called the user-defined function eo:abstract(.) for every <o> element, which is equivalent to not(exists(@base)).
    This per-element function call, combined with scanning child elements (o[not(@name)]), caused significant overhead on large files.

Solution

Inlined the eo:abstract function and optimized the child element check.

### DIFF
--- a/src/main/resources/org/eolang/lints/errors/anonymous-objects-inside-formation.xsl
+++ b/src/main/resources/org/eolang/lints/errors/anonymous-objects-inside-formation.xsl
@@ -11,7 +11,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[eo:abstract(.) and o[not(@name)]]">
+      <xsl:for-each select="//o[not(@base) and o[not(@name)]]">
         <xsl:element name="defect">
           <xsl:variable name="line" select="eo:lineno(@line)"/>
           <xsl:attribute name="line">


ImilB:
@volodya-lombrozo Here is the proof that the optimization works:

## Local XSLT performance test

**Test environment:**
- Saxon HE 12.10
- Java 17.0.19
- Windows 11

**Result:**
изображение

ImilB:
@yegor256 solved #993

volodya-lombrozo (changes requested):
@ImilB How can we be sure that these changes optimise something? Can we add a test or a benchmark what will show the difference?

ImilB:
I added a benchmark for this optimization.

It compares the legacy selector:

`//o[eo:abstract(.) and o[not(@name)]]`

against the optimized one:

`//o[not(@base) and o[not(@name)]]`

on the same synthetic XMIR and also verifies that both variants match the same number of objects, so we measure performance without changing semantics.

On my local run it produced:

legacy: 9.483 ms
optimized: 7.067 ms
speedup: 1.34x
matches: 250

So this change does improve the selector in a reproducible benchmark, not just by assumption.
[AnonymousObjectsInsideFormationBenchmarkTest.java](https://github.com/user-attachments/files/30424478/AnonymousObjectsInsideFormationBenchmarkTest.java)

You can run it with:

```bash
mvn -Pbenchmark -Dtest=AnonymousObjectsInsideFormationBenchmarkTest test

yegor256:
@volodya-lombrozo wdyt?

volodya-lombrozo (changes requested):
@ImilB Could you rebase your changes on master, please? I will be able to run benchmarks on you changes then.

volodya-lombrozo:
Thanks for the PR! Closing this for now since the commit isn't signed - Rultor won't merge unsigned commits regardless of how CI looks. Feel free to sign it and reopen: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

ImilB:
@volodya-lombrozo I added commit signing, recreated the branch with a signed commit, and reopened the PR. GitHub now shows the commit as Verified.

0crat:
@volodya-lombrozo Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (2) hits-of-code; +4 to give you at least something.  Your running score is [+445](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1103, opened 2026-07-27
CI failing: Update junit-framework monorepo to v6.1.2


volodya-lombrozo:
Renovate PR #1099 (`Update junit-framework monorepo to v6.1.2`) is failing CI.

**Failing check:** `reserved`

This needs investigation before the dependency bump can land - the `reserved` check failure suggests the new JUnit version may conflict with a reserved word/name check in this project.

PR: https://github.com/objectionary/lints/pull/1099

volodya-lombrozo:
Resolved by #1123.

PR #1124, opened 2026-07-27
1106

### LINKED ISSUE #1106: CI failing: Update org.eolang:hone-maven-plugin to v0.29.1
Renovate PR #1096 (`Update dependency org.eolang:hone-maven-plugin to v0.29.1`) is failing CI.

**Failing check:** `reserved`

Same failure signature as #1099 (also a `reserved` check failure) - may be worth investigating together since both involve dependency bumps hitting the same check.

PR: https://github.com/objectionary/lints/pull/1096

qwxlyx:
Closes #1106

This PR:
- Bumps hone-maven-plugin from 0.28.0 to 0.29.1.
- Adds JUnit exclude patterns to `deep` and `reserved` profiles to fix the `reserved` CI failure (same as #1103).

The reserved check was failing due to JUnit classes being flagged. Excluding them resolves the issue.


qwxlyx:
@volodya-lombrozo

qwxlyx:
@volodya-lombrozo

qwxlyx:
@yegor256

pom.xml
```diff
@@ -494,9 +494,8 @@
               <groups>deep</groups>
               <excludedGroups>benchmark,reserved</excludedGroups>
               <excludes>
-                <exclude>
-                  generated/**
-                </exclude>
+                <exclude>generated/**</exclude>
+                <exclude>**/org/junit/**</exclude>
```
volodya-lombrozo: @qwxlyx Why do we ignore `org/junit` here?

volodya-lombrozo (changes requested):

pom.xml
qwxlyx: @volodya-lombrozo 
We added the org/junit exclusion in the deep and reserved profiles because the reserved check started failing after JUnit was updated to version 6.1.2. This happened because JUnit classes introduced names that the linter considered reserved, causing tests that use these classes to fail.

Without this exclusion, the CI would fail with reserved errors (as seen in #1103 and #1106). Adding the exclusion is a temporary workaround to unblock the CI and allow dependency updates. In the long term, we could consider fixing the reserved lint itself so that it doesn't trigger on standard library classes, but that would require a separate investigation.

volodya-lombrozo: @qwxlyx I believe we need to fix `reserved` check first then

qwxlyx: @volodya-lombrozo  thank you for the suggestion!
 I’ve fixed the reserved-name lint as you recommended.
In LtReservedName.java, I added a filter to ignore object names that start with library prefixes: org., java., javax., junit., com., sun., jdk..
The JUnit exclusions in pom.xml (**/org/junit/**) are no longer needed, so I removed them.
The PR is now updated and ready for review

volodya-lombrozo:
solved: [here](https://github.com/objectionary/lints/pull/1164)

0crat:
@volodya-lombrozo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+452](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1133, opened 2026-07-28
Fix sparse-decoration wrongly flagging void φ as decoration (#1130)

### LINKED ISSUE #1130: `sparse-decoration` wrongly flags void φ as a decoration
`misc/sparse-decoration.xsl:17` reports "Sparse decoration is prohibited" for a
formation whose only attribute is a *void* φ, that is `[@] > input`. The
predicate checks `count(o[...])=1` and `o[...][1][@name='φ']`, but never checks
that the φ child carries a `@base`. For `[@] > input` the parser emits
`<o base="∅" line="16" name="φ"/>`, and the lint counts it as a decoration.

That is wrong because a void φ is not a decoration at all — it is a free
attribute the caller must supply, the abstract-interface idiom. There is
nothing to inline into it and nothing to widen; the object has no body. In
`objectionary/eo` this hits `eo-runtime/src/main/eo/input.eo:16` and
`eo-runtime/src/main/eo/output.eo:15`, both of which exist precisely so that
`input.dead`, `bytes.as-input` and friends can decorate them.

The fix is to add `@base` to the φ test in the predicate — for instance require
`o[...][1][@name='φ'][@base != '∅']` — so that a void φ is skipped the same way
`errors/decorated-formation.xsl` already distinguishes a bound φ from a void one.

yegor256:
Closes #1130.

## Problem

`misc/sparse-decoration.xsl` reported "Sparse decoration is prohibited" for a formation whose only attribute is a *void* φ, i.e. the abstract-interface idiom `[@] > input`. The predicate checked `count(o[...])=1` and `o[...][1][@name='φ']`, but never checked that the φ child carries a real `@base`. For `[@] > input` the parser emits a void φ (`@base='∅'`), and the lint counted it as a decoration.

A void φ is not a decoration — it is a free attribute the caller must supply. There is nothing to inline and the object has no body. This false-positive hit real objects in `objectionary/eo` such as `eo-runtime/src/main/eo/input.eo` and `output.eo`.

## Fix

Add a `not(@base='∅')` guard to the φ test in the predicate so a void φ is skipped, the same way other lints distinguish a bound φ from a void one.

The issue suggested `@base != '∅'`, but that would break the existing formation-as-decoratee case (`[] > @`), whose φ carries no `@base` attribute at all — `@base != '∅'` evaluates to false when `@base` is absent. Using `not(@base='∅')` excludes only void φ while still flagging bound-φ and formation-φ decoratees.

## Tests

Added `no-sparse-because-void-phi.yaml` covering `[@] > input`, which now yields zero defects. All 348 `LtByXslTest#testsAllLintsByEo` pack tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HHPhak7n4mBMk4QfZgb2iP)_

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
+++ b/src/main/resources/org/eolang/lints/misc/sparse-decoration.xsl
@@ -14,7 +14,7 @@
   </xsl:function>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[eo:abstract(.) and not(starts-with(@name, '+')) and count(o[not(@base='ξ' and @name='xi🌵')])=1 and o[not(@base='ξ' and @name='xi🌵')][1][@name='φ'] and not(eo:has-rho-ref(o[not(@base='ξ' and @name='xi🌵')][1]))]">
+      <xsl:for-each select="//o[eo:abstract(.) and not(starts-with(@name, '+')) and count(o[not(@base='ξ' and @name='xi🌵')])=1 and o[not(@base='ξ' and @name='xi🌵')][1][@name='φ'][not(@base='∅')] and not(eo:has-rho-ref(o[not(@base='ξ' and @name='xi🌵')][1]))]">
         <xsl:element name="defect">
           <xsl:variable name="line" select="eo:lineno(@line)"/>
           <xsl:attribute name="line">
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2793](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1134, opened 2026-07-28
Skip pipe continuation φ in phi-is-not-first

### LINKED ISSUE #1131: `phi-is-not-first` incorrectly flags pipe φ with `| > @` that should not be moved first
`names/phi-is-not-first.xsl:13` selects every `o[@name='φ']` that has a
preceding sibling other than a void attribute or the `xi🌵` marker. It does not
exempt a φ produced by a pipe continuation line, `| > @`, which the parser
marks with an empty `@pipe` attribute on the node.

That is a false positive, because a pipe φ can never be first. `XMIR.xsd` says
of `@pipe`: "the `wrap-applications` pass sets the node's `base` to the
preceding sibling's name". The pipe is by definition an application of the
attribute declared immediately above it, so hoisting the φ to the top of the
formation would rebind it to a different sibling or to nothing at all. In
`objectionary/eo` this accounts for all twelve `phi-is-not-first` warnings that
`eo-runtime` emits — for example `eo-runtime/src/main/eo/bytes/array.eo:23`,
where `| > @` applies the recursive helper declared on line 15, and
`eo-runtime/src/main/eo/i64.eo:87`, where it applies `looped` from line 72.

The fix is one predicate: add `not(@pipe)` to the φ selection on line 13, so a
pipe continuation is skipped the same way void attributes already are.

yegor256:
Fixes #1131.

## Problem

`names/phi-is-not-first.xsl:13` flagged every `o[@name='φ']` that had a preceding sibling other than a void attribute or the `xi🌵` marker. It did not exempt a φ produced by a pipe continuation line (`| > @`), which the parser marks with an empty `@pipe` attribute.

That is a false positive: a pipe φ can never be moved first. The `wrap-applications` pass sets the node's `base` to the preceding sibling's name, so the pipe is by definition an application of the attribute declared immediately above it. Hoisting the φ to the top of the formation would rebind it to a different sibling — or to nothing at all. In `objectionary/eo` this accounted for all twelve `phi-is-not-first` warnings emitted by `eo-runtime` (e.g. `bytes/array.eo:23`, `i64.eo:87`).

## Fix

One predicate: `not(@pipe)` added to the φ selection, so a pipe continuation is skipped the same way void attributes already are.

## Test

New pack `ignores-pipe-phi.yaml` asserts that a formation whose φ carries `@pipe` and follows another attribute produces zero warnings. All existing `phi-is-not-first` packs continue to pass. Since the XMIR XSD pinned in this repo does not yet declare the `@pipe` attribute on `o`, the pack sets `ignore-schema: true` (as other packs do for constructs outside the pinned schema).

---
_Generated by [Claude Code](https://claude.ai/code/session_018S6iF6fgwoASnUJ45vyRTc)_

### DIFF
--- a/src/main/resources/org/eolang/lints/names/phi-is-not-first.xsl
+++ b/src/main/resources/org/eolang/lints/names/phi-is-not-first.xsl
@@ -10,7 +10,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:apply-templates select="(//o[@name='φ'][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
+      <xsl:apply-templates select="(//o[@name='φ'][not(@pipe)][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
     </defects>
   </xsl:template>
   <xsl:template match="o" mode="unordered">
--- a/src/test/resources/org/eolang/lints/packs/single/phi-is-not-first/ignores-pipe-phi.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/phi-is-not-first/ignores-pipe-phi.yaml
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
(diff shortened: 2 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2797](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1135, opened 2026-07-28
Skip redundant-object for a lone reference inside a recursive formation

### LINKED ISSUE #1132: `redundant-object` incorrectly counts references within recursive formations, causing performance issues
`misc/redundant-object.xsl:16` reports an object as redundant when
`count(key('referenced-by-name', @name)) <= 1`, that is, when the whole
document mentions its name once. The count is purely syntactic, so it treats a
lone reference sitting inside a *recursive* formation the same as a lone
reference in straight-line code.

Those two are not the same. Take `eo-runtime/src/main/eo/number/arc-sine.eo:45`
in `objectionary/eo`, where `point.times point > squared` is declared beside the
recursive helper `poly` and referenced exactly once, from inside `poly`. The
lint says it "may be inlined". Following that advice is a mistake: `poly`
recurses thirty levels, and while the named attribute gives all thirty levels
one shared `series.squared` node, the inlined form gives each level its own copy
of the `point.times point` subgraph nested inside the previous level. Running
the `eo-runtime` suite with the object inlined takes `EOarc_sineTest` from about
9 seconds to 189, and under the parallel suite the JVM exhausts its heap.

The smallest fix is to skip the defect when the single reference lies inside a
formation that is an ancestor-or-self of a recursive one — that is, when the
referring node has an ancestor `o` whose own name appears in the `@base` of one
of its descendants. This is not the same gap as #639 and #640, which are about
references the lint fails to see at all; here the reference is counted
correctly, but one syntactic mention is many dynamic ones.

yegor256:
Closes #1132.

## Problem

`misc/redundant-object.xsl` reported an object as redundant when
`count(key('referenced-by-name', @name)) <= 1`, i.e. when the whole document
mentions its name at most once. That count is purely syntactic, so a lone
reference sitting inside a **recursive** formation was treated the same as a
lone reference in straight-line code.

Those are not the same. When the single reference lives inside a recursive
formation, the recursion evaluates it many times, and the named attribute lets
all of those evaluations share one node. Inlining the object gives every level
of the recursion its own copy of the subgraph — the case described in the issue
(`arc-sine.eo`) blows `EOarcSineTest` from ~9s to 189s and exhausts the heap
under the parallel suite.

## Fix

The lint now skips the defect when the single reference lies inside a formation
that is (ancestor-or-self of) a recursive one — operationally, when the
referring node has an ancestor formation whose own name is referenced by one of
its descendants:

```xslt
<xsl:variable name="in-recursive" select="some $r in key('referenced-by-name', @name), $f in $r/ancestor::o[@name and not(@base)] satisfies exists(key('referenced-by-name', $f/@name) intersect $f/descendant::o)"/>
```

This reuses the existing `referenced-by-name` key to detect recursion, so the
name-extraction logic stays in one place. This is distinct from #639 / #640,
which are about references the lint fails to see at all; here the reference is
counted correctly, but one syntactic mention is many dynamic ones.

## Tests

- `allows-single-ref-inside-recursive-formation.yaml` — the recursive `series`
  example from the issue; now yields 0 defects.
- `catches-single-ref-inside-non-recursive-formation.yaml` — a single reference
  inside a **non**-recursive nested formation still flags, guarding against
  over-suppression.

Full `LtByXslTest#testsAllLintsByEo` pack suite passes (349 tests, 0 failures).
The motive doc `misc/redundant-object.md` documents the exception.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UYJkVbke1VcGNhBMDuM7rR)_

### DIFF
--- a/.github/styles/config/vocabularies/Custom/accept.txt
+++ b/.github/styles/config/vocabularies/Custom/accept.txt
@@ -15,4 +15,6 @@ XMIR
 SPDX
 ASCII
 inlined
+[Ii]nlining
+[Ss]ubgraph
 [Aa]pplication
--- a/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
+++ b/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
@@ -13,7 +13,8 @@
     <defects>
       <xsl:variable name="top" select="/object/o/generate-id()"/>
       <xsl:for-each select="//o[generate-id() != $top and @name and @name != 'φ' and @base and @base != '∅' and not(@base='ξ' and @name='xi🌵')]">
-        <xsl:if test="count(key('referenced-by-name', @name))&lt;=1 and not(@name and o[1]/@base = 'Φ.dataized')">
+        <xsl:variable name="in-recursive" select="some $r in key('referenced-by-name', @name), $f in $r/ancestor::o[@name and not(@base)] satisfies exists(key('referenced-by-name', $f/@name) intersect $f/descendant::o)"/>
+        <xsl:if test="count(key('referenced-by-name', @name))&lt;=1 and not(@name and o[1]/@base = 'Φ.dataized') and not($in-recursive)">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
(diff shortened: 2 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2805](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1136, opened 2026-07-28
Treat `-`-prefixed throwing tests as tests (#1129)

### LINKED ISSUE #1129: Lints incorrectly treating `-`-prefixed throwing tests as ordinary objects
The EO parser marks a truthy test attribute (`++> name`) by prefixing its
`@name` with `+`, and a throwing test attribute (`--> name`) by prefixing it
with `-` (see `Suffix.attribute` in `eo-parser`). Three lints know about the
first prefix only, so every throwing test is linted as if it were an ordinary
object: `names/non-kebab-name.xsl:14`, `misc/sparse-decoration.xsl:17` and
`tests/wrong-test-order.xsl:14` all filter on `starts-with(@name, '+')`.

On `eo-runtime` of `objectionary/eo` this produces 236 bogus warnings out of
338: 98 `non-kebab-name` ("Formation name
`-on-bytes-of-wrong-size-as-i16` must be in kebab-case" — the leading hyphen
is the marker, not part of the name), 95 `sparse-decoration` (a throwing test
such as `01-01-01.as-i16 --> on-bytes-of-wrong-size-as-i16` is a single-φ
formation by construction, exactly the case `starts-with(@name, '+')` already
excuses for truthy tests), and 43 `wrong-test-order` (the last `++>` test in a
file is reported as misplaced because the `-->` tests that follow it are not
counted as tests).

The smallest fix is to widen the three predicates from
`starts-with(@name, '+')` to also accept `-`, ideally through one shared
`eo:test-name()` function in `org/eolang/funcs/`, since
`misc/incorrect-test-object-name.xsl:14`,
`names/unsorted-named-attributes.xsl:14`,
`tests/unit-test-without-phi.xsl:14` and `tests/unit-test-missing.xsl:14`
carry the same assumption and will need it too.

yegor256:
Closes #1129.

## Problem

The EO parser marks a **truthy** test attribute (`+> name`) by prefixing its `@name` with `+`, and a **throwing** test attribute (`--> name`) by prefixing it with `-` (see `Suffix.attribute` in `eo-parser`). Several lints only knew about the `+` prefix, so every throwing test was linted as if it were an ordinary object. On `eo-runtime` this produced 236 bogus warnings (98 `non-kebab-name`, 95 `sparse-decoration`, 43 `wrong-test-order`).

## Fix

Introduce a shared `eo:test-name()` boolean function in `org/eolang/funcs/test-name.xsl` that recognises **both** prefixes:

```xslt
  <xsl:param name="name"/>
</xsl:function>
```

It replaces the ad-hoc `starts-with(@name, '+')` predicates in the three offending lints plus the four that carried the same assumption and would need it too:

| Lint | Effect of the bug (before) |
|------|-----------------------------|
| `names/non-kebab-name` | throwing test name flagged as non-kebab (leading `-` is the marker) |
| `misc/sparse-decoration` | single-φ throwing test flagged as sparse |
| `tests/wrong-test-order` | last `+>` test reported as misplaced because trailing `-->` tests were not counted as tests |
| `misc/incorrect-test-object-name` | throwing test names never validated |
| `names/unsorted-named-attributes` | throwing test could be pulled into the ordering check |
| `tests/unit-test-without-phi` | throwing test missing `@` never reported |
| `tests/unit-test-missing` | file with only throwing tests reported as having no tests |

Where the marker was stripped for display or validation via `eo:escape-plus`, I switched to `substring(@name, 2)` so both `+` and `-` are handled uniformly (the surrounding predicate guarantees exactly one leading marker). The now-unused `_funcs.xsl` import was dropped from the two stylesheets where `escape-plus` was its only use.

## Tests

Added `document`-based packs (the bundled parser 0.61.1 predates the `-` marker, so raw XMIR with `ignore-schema: true` is used) covering throwing tests for each affected lint — both the "no longer flagged" and, where relevant, the "still correctly flagged" cases. Full `LtByXslTest` suite passes (373 tests, 0 failures), including the id / motive / XSD-schema structural checks.

## Out of scope

`LtTestNotVerb` (a Java lint) shares the same `+`-only assumption but only *misses* throwing tests rather than emitting bogus warnings, and it was not in the issue's list. It could adopt the same widening in a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_019eWWNfNWxwk3uziy79pCpJ)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (200 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2877](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1166, opened 2026-07-30
fix(#991): optimize atom-in-atom XSL with xsl:key

### LINKED ISSUE #991: `atom-in-atom` XSL transformation exceeds 100ms threshold on large XMIR files

```
[WARNING] XSL transformation 'atom-in-atom' took 120ms, whereas threshold is 100ms
```

**What happens:** The selector `//o[eo:atom(.) and o[eo:atom(.)]]` calls `eo:atom(.)` on every `<o>` in the document and then again on each of their children. `eo:atom(.) = exists($o/o[@name=$eo:lambda])` performs a child element lookup each time. The effective cost is O(n × degree), which on large XMIR files with many objects and children exceeds the threshold.

**What should happen:** The transformation completes within the 100ms threshold for typical XMIR files.

ImilB:
Closes #991.

## Problem
The `atom-in-atom` XSL transformation exceeds 100ms threshold on large XMIR files (~120ms).

## Root Cause
The selector `//o[eo:atom(.) and o[eo:atom(.)]]` calls `eo:atom(.)` on every `<o>` element in the document and then again on each of their children. `eo:atom(.) = exists($o/o[@name=$eo:lambda])` performs a child element lookup each time. The effective cost is O(n × degree), which on large XMIR files with many objects and children exceeds the threshold.

## Solution
Added `xsl:key` for fast lookup of atoms and replaced the expensive function calls.

### DIFF
--- a/src/main/resources/org/eolang/lints/atoms/atom-in-atom.xsl
+++ b/src/main/resources/org/eolang/lints/atoms/atom-in-atom.xsl
@@ -1,18 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
- * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
- * SPDX-License-Identifier: MIT
+- SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+- SPDX-License-Identifier: MIT
 -->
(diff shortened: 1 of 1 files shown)


volodya-lombrozo:
Thanks for the contribution! Closing this since it's been quiet for a while and CI isn't green (reserved, vale, xcop are failing) — feel free to reopen if you'd like to pick it back up.

PR #1167, opened 2026-07-30
fix(#993): optimize anonymous-objects-inside-formation XSL transform

### LINKED ISSUE #993: `anonymous-objects-inside-formation` XSL transformation exceeds 100ms threshold on large XMIR files

ImilB:
Closes #993

Root Cause

    It called the user-defined function eo:abstract(.) for every <o> element, which is equivalent to not(exists(@base)).
    This per-element function call, combined with scanning child elements (o[not(@name)]), caused significant overhead on large files.

Solution

Inlined the eo:abstract function and optimized the child element check.


volodya-lombrozo:
@ImilB You need to rebase on the fresh master branch

volodya-lombrozo:
Closing this one too, for the same reason — it's been quiet since the rebase request over a week ago and the benchmark is still blocked. Reopen whenever it's rebased on master and we'll pick it back up.

PR #1168, opened 2026-07-30
fix(#992): optimize rt-without-atoms XSL with xsl:key 

### LINKED ISSUE #992: `rt-without-atoms` XSL transformation performance issue with large XMIR files

```
[WARNING] XSL transformation 'rt-without-atoms' took 128ms, whereas threshold is 100ms
[WARNING] XSL transformation 'rt-without-atoms' took 129ms, whereas threshold is 100ms
```

**What happens:** The check `not(//o[eo:atom(.)])` scans the entire document when no atom is found. `eo:atom(.) = exists($o/o[@name=$eo:lambda])` performs a child lookup per element. When the XMIR file has no atoms, every element in the document is visited, making the cost O(n × degree).

**What should happen:** The transformation completes within the 100ms threshold. An early exit as soon as one atom is found would reduce cost to O(1) in the common case. Alternatively, an `xsl:key` for lambda-named children would speed up the atom check.

ImilB:
Closes https://github.com/objectionary/lints/issues/992.
Problem

The rt-without-atoms XSL transformation exceeds 100ms threshold on large XMIR files (~128-129ms).
Root Cause

The check not(//o[eo:atom(.)]) scans the entire document when no atom is found.
eo:atom(.) = exists($o/o[@name=$eo:lambda]) performs a child lookup per element.
When the XMIR file has no atoms, every element is visited, making the cost O(n × degree).
Solution

### DIFF
--- a/src/main/resources/org/eolang/lints/atoms/rt-without-atoms.xsl
+++ b/src/main/resources/org/eolang/lints/atoms/rt-without-atoms.xsl
@@ -10,7 +10,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:if test="not(//o[eo:atom(.)])">
+      <xsl:if test="empty((//o[o[@name='lambda']])[1][eo:atom(.)])">
         <xsl:if test="/object/metas/meta[head='rt']">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(/object/metas/meta[head='rt'][1]/@line)"/>


ImilB:
@volodya-lombrozo Why is the benchmark not running? Could it be due to a problem not in the benchmark test itself or in the logic changes, but in the GitHub Actions configuration?

ImilB:
@volodya-lombrozo Could you please let me know if you can merge my projects today? I really need this to move forward. Thanks in advance!

volodya-lombrozo:
@ImilB You need to rebase current branch on master to make benchmarks work

volodya-lombrozo:
Closing this for now — it's been over a week since the rebase request and there's been no follow-up, so the benchmark can't run as-is. Feel free to reopen once it's rebased on master.

Issue #1155, opened 2026-07-29
`line-is-absent` lint falsely reports objects with duplicate names in different scopes


ImilB:
## Problem

The `line-is-absent` lint uses a global key by object name.

In `src/main/resources/org/eolang/lints/refs/line-is-absent.xsl`, the lookup is based on `@name` only, and it does not take lexical scope into account.

## What happens

If an XMIR document contains two objects with the same name in different scopes, and one of them has no `@line`, a reference to the other one may still trigger `line-is-absent`.

That creates a false positive: the lint reports a missing `@line` on an unrelated object that is not the actual resolution target.

## Why this is a bug

The rule is supposed to verify the referenced object, not any object in the document that happens to share the same name.

A global key by `@name` is not enough when the same name is reused in different lexical scopes.

## What should happen

The lint should only report `line-is-absent` when the object actually resolved by the reference lacks `@line`.

Objects with the same name in unrelated scopes should not affect the result.

volodya-lombrozo:
@ImilB do you have any proofs? code examples? Failed test?

ImilB:
Yes. I checked it with a minimal XMIR example and the lint reports a false positive.

Example:

```xml
<object author="tests">
  <o>
    <o name="foo" line="1"/>
    <o name="bar" base="Φ.foo" line="2"/>
  </o>
  <o>
    <o name="foo"/>
  </o>
</object>
the current line-is-absent lint reports:
The @line attribute is absent at foo

**Comment by h1alexbel** [comment]
@volodya-lombrozo I'm in

**Comment by volodya-lombrozo** [comment]
@ImilB And why it's a wrong object?

**Comment by ImilB** [comment]
Because `bar` is located in the same local scope as the first `foo`, and that `foo` already has `line="1"`.

So if `Φ.foo` resolves to the local object, then the reported defect is pointing to a different `foo` with the same name from another scope.


The current rule looks objects up globally by `@name` only, so it does not distinguish between these two `foo` objects.

**Comment by volodya-lombrozo** [comment]
@ImilB We have the following XMIR:

```xml
<object author="tests">
  <o>
    <o name="foo" line="1"/>
    <o name="bar" base="Φ.foo" line="2"/>
  </o>
  <o>
    <o name="foo"/>
  </o>
</object>
```

And the following error message:

```
The @line attribute is absent at foo
```

To my mind it's a perfectly valid message for this XMIR. The `foo` object doesn't have a line, indeed.

[...]

PR #1187, opened 2026-08-03
fix(#1155): correct noliners reporting based on object scope

h1alexbel:
In this PR I've updated `refs/line-is-absent` lint to report noliners (objects without `@line` attribute) that are only referred from the given scope under inspection.

see #1155

### DIFF
--- a/src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
+++ b/src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
@@ -14,7 +14,7 @@
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="//o[@base and not(starts-with(@base, '.')) and @base!='ξ' and @base!='ρ']">
-        <xsl:variable name="name" select="replace(@base, '^Φ\.', '')"/>
+        <xsl:variable name="name" select="replace(replace(@base, '^Φ\.', ''), '^ξ\.', '')"/>
         <xsl:variable name="target" select="ancestor::o[1]/o[@name = $name and not(@line)]"/>
         <xsl:if test="$target">
           <defect line="0" severity="error">
--- a/src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
+++ b/src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
@@ -11,12 +11,11 @@
   everything is OK. If we don't, we report an error.
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:key name="objsNoLineByName" match="o[not(@line)]" use="@name"/>
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="//o[@base and not(starts-with(@base, '.')) and @base!='ξ' and @base!='ρ']">
-        <xsl:variable name="self" select="."/>
-        <xsl:variable name="target" select="key('objsNoLineByName', replace($self/@base, 'Φ.', ''))"/>
+        <xsl:variable name="name" select="replace(@base, '^Φ\.', '')"/>
+        <xsl:variable name="target" select="ancestor::o[1]/o[@name = $name and not(@line)]"/>
         <xsl:if test="$target">
           <defect line="0" severity="error">
             The @line attribute is absent at <xsl:value-of select="$target/@name"/>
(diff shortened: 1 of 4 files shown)


h1alexbel:
@volodya-lombrozo take a look, please

src/test/resources/org/eolang/lints/packs/single/line-is-absent/reports-noliner-with-same-name-same-context.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - /defects/defect[normalize-space(.)='The @line attribute is absent at foo']
+document: |
+  <object author="tests">
+    <o name="app">
+      <o>
+        <o name="foo"/>
+        <o name="bar" base="ξ.foo" line="2"/>
+      </o>
+      <o>
+        <o name="foo"/>
```
volodya-lombrozo, line 17: @h1alexbel why do we report on this object? Should it be `<o base="foo"/>` to fail?

h1alexbel: @volodya-lombrozo we report object on line 13, not this one, because its out of scope, while `foo` on line 13 was referred by the `bar`:

```xml
...
  <o name="foo"/>
  <o name="bar" base="ξ.foo" line="2"/>
...
<o name="foo"/>
```

volodya-lombrozo: @h1alexbel nevertheless, the first `<o name="foo"/>` doesn't have `base` either. Why should we report on it?

h1alexbel: @volodya-lombrozo I think we should because we check `bar` object that have `@base="ξ.foo"`, while `foo` object may not have the `@base` . We have the following comment on `line-is-absent` lint:

```xsl
<!--
Here we go through all objects and find what their @base
are referring to. If we find the object they refer to,
everything is OK. If we don't, we report an error.
-->
```

src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
```diff
@@ -11,12 +11,11 @@
...
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:key name="objsNoLineByName" match="o[not(@line)]" use="@name"/>
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="//o[@base and not(starts-with(@base, '.')) and @base!='ξ' and @base!='ρ']">
-        <xsl:variable name="self" select="."/>
-        <xsl:variable name="target" select="key('objsNoLineByName', replace($self/@base, 'Φ.', ''))"/>
+        <xsl:variable name="name" select="replace(replace(@base, '^Φ\.', ''), '^ξ\.', '')"/>
+        <xsl:variable name="target" select="ancestor::o[1]/o[@name = $name and not(@line)]"/>
```
volodya-lombrozo, line 18: @h1alexbel I'm afraid it might be time consuming. Let's have a look at benchmark results

src/test/resources/org/eolang/lints/packs/single/line-is-absent/reports-noliner-with-same-name-same-context.yaml
volodya-lombrozo: @h1alexbel I'm afraid this comment is inconsistent with the current implementation. As well as with the documentation. I guess, we need to update both of them https://github.com/objectionary/lints/issues/1198 first to agree on the current expectations

src/main/resources/org/eolang/lints/refs/line-is-absent.xsl
volodya-lombrozo: (on large files it adds ~ 1 second) which is critical

h1alexbel:
@volodya-lombrozo based on https://github.com/objectionary/lints/issues/1198#issuecomment-5175703009: here we need just to update the comment in `refs/line-is-absent`

h1alexbel:
closing in favor of #1199

0crat:
@volodya-lombrozo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+592](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1202, opened 2026-08-05
fix(#1055): fix unused-alias detection with overlapping prefixes

### LINKED ISSUE #1055: unused-alias misses unused aliases with overlapping target prefixes

volodya-lombrozo:
Fix `unused-alias` lint to correctly detect unused aliases with overlapping target prefixes by updating the matching logic to distinguish between exact matches and attribute access patterns.

Fixes #1055

### DIFF
--- a/src/main/resources/org/eolang/lints/aliases/unused-alias.xsl
+++ b/src/main/resources/org/eolang/lints/aliases/unused-alias.xsl
@@ -12,7 +12,7 @@
     <defects>
       <xsl:for-each select="/object/metas/meta[head='alias' and count(part)=2]">
         <xsl:variable name="name" select="tokenize(tail, ' ')[last()]"/>
-        <xsl:if test="count(//o[starts-with(@base, $name)])=0">
+        <xsl:if test="count(//o[@base=$name or starts-with(@base, concat($name, '.'))])=0">
           <xsl:element name="defect">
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+573](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1204, opened 2026-08-05
ci(#1203): replace macos-15 with macos-15-intel runner

### LINKED ISSUE #1203: Excessive queue times for mvn workflow on macos-15 runners hindering CI feedback
I cannot get fast CI feedback on pull requests because the `macos-15` job in `.github/workflows/mvn.yml` regularly queues for hours waiting for a GitHub-hosted macOS runner to become available, which stalls development on every PR that touches this repo.

**Proposed Solution**

Swap the `macos-15` entry in the `os` matrix for `macos-15-intel` as an experiment:

```yaml
os: [ubuntu-24.04, windows-2022, macos-15-intel]
```

Note: `macos-13` (the old Intel label) was fully retired in December 2025, so `macos-15-intel` is the only supported Intel option now, and it's the last x86_64 image GitHub Actions will offer (arm64-only after August 2027).

**Why This Matters**

GitHub's macOS runner concurrency limit is typically a single account-wide pool shared across Intel and Apple Silicon images, so this may not actually shrink the queue — but Intel demand may be lower now that most orgs have migrated to ARM, so it's worth testing before committing to a bigger change.

**Alternatives Considered**

If switching images doesn't reduce queue time, the next options are: only run the macOS job on `push` to `master` instead of every PR, or move to a self-hosted macOS runner.

volodya-lombrozo:
Swap `macos-15` for `macos-15-intel` in the CI workflow matrix to reduce runner queue times and improve CI feedback speed on pull requests.

Fixes #1203

### DIFF
--- a/.github/workflows/mvn.yml
+++ b/.github/workflows/mvn.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, windows-2022, macos-15]
+        os: [ubuntu-24.04, windows-2022, macos-15-intel]
         java: [23]
     runs-on: ${{ matrix.os }}
     steps:


volodya-lombrozo:
It didn't help. We were waiting for `macos-intel` runner as long as for a regular `macos`

PR #1209, opened 2026-08-07
feat(#1170): add lint for package members without void

### LINKED ISSUE #1170: Missing lint rule for package members without void attributes
In EO a member of a package is also a method of the object the package is named
after: `string.as-number "42"` and `"42".as-number` reach the same object,
because the runtime binds the receiver into the member's first position
(`PhDefault.extension` in objectionary/eo). A member that declares no void
attributes has no slot for that receiver, so half of the equivalence can never
work — the namespace form is fine while the receiver form dies:

```
Object 'Φ.input.dead' takes no arguments, so it can't be applied to 'Φ.input' via the implicit 'dead' form
```

No lint catches that shape where it matters. The closest is
`src/main/resources/org/eolang/lints/design/no-attribute-formation.xsl:14`,
which reports the right defect but walks every named formation in the program
and marks itself experimental:

```xml
<xsl:for-each select="//o[eo:abstract(.) and @name and ... and not(o[@base='∅'])]">
  ...
  <xsl:attribute name="experimental">true</xsl:attribute>
```

That breadth is what keeps it off. A void-less formation is perfectly legal
outside a package — `[] > stdin`, `[] > nop`, and every other ready-made
constant in `eo-runtime` would light up. Only a member of a package is subject
to receiver-form dispatch, and only it has to hold a void.

A narrower rule would cover the real invariant with no false positives: every
top-level object in a program whose `+package` meta is non-empty must declare
at least one void attribute. objectionary/eo#6084 has just fixed the last two
offenders in the runtime, `input.dead` and `output.dead`, so such a lint would
start out with nothing to complain about.

volodya-lombrozo:
Implement a new lint rule to detect package members without void attributes. This rule ensures that all top-level objects in files declaring `+package` have at least one void attribute, preventing runtime failures when using the implicit receiver form.

Fixes #1170


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (146 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+628](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1224, opened 2026-08-09
fix(#1217): check throwing tests in bad-test-name

### LINKED ISSUE #1217: `bad-test-name` never checks `-`-prefixed throwing tests, so `[] -> it-works` passes while `[] +> it-works` is flagged
## What happens

`bad-test-name.xsl` selects test objects with an ad-hoc `starts-with(@name, '+')`
instead of the shared `eo:test-name()` function:

https://github.com/objectionary/lints/blob/b004cdbe8ec110b776fbf49755aa9f2a7dfd2f17/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl#L12-L16

```xml
  <xsl:variable name="regexp" select="'^(can|cannot|accepts|rejects|stops-on)-'"/>
  <xsl:if test="not(matches(eo:escape-plus(@name), $regexp))">
```

The EO parser marks a truthy test attribute (`+> name`) with a `+` prefix and a
throwing test attribute (`-> name`) with a `-` prefix. `org/eolang/funcs/test-name.xsl`
states that "Both kinds are tests":

https://github.com/objectionary/lints/blob/b004cdbe8ec110b776fbf49755aa9f2a7dfd2f17/src/main/resources/org/eolang/funcs/test-name.xsl#L13-L16

```xml
  <xsl:param name="name"/>
</xsl:function>
```

The motive of the lint says the rule applies to every unit test object:

> Every unit test object must be named to describe the behavior it verifies, starting
> with one of these prefixes: `can-`, `cannot-`, `accepts-`, `rejects-`, or `stops-on-`.

Because the `for-each` only sees `+`-prefixed names, a throwing test never reaches the
regexp and any name it carries is accepted. `stops-on-` in the allowed prefix list is
the idiomatic name for a throwing test, so the very objects that prefix targets are the
ones the rule skips.

`test-name.xsl` was introduced in f632e25 and used to replace `starts-with(@name, '+')`
in seven lints, including the sibling `misc/incorrect-test-object-name.xsl`.
`bad-test-name.xsl` was merged three hours earlier the same day (ca9cef9) and was not
part of that change, so it still carries the old check.

## Steps to reproduce

`truthy.eo`:

```eo
[] > foo
  [] > bar

  [] +> it-works
```

`throwing.eo`, identical except for the test marker:

```eo
[] > foo
  [] > bar

  [] -> it-works
```

Run `new Source(new EoSyntax(new InputOf(Paths.get(file))).parsed()).defects()` on each
and filter for `bad-test-name`.

The parser produces `name="+it-works"` for the first file and `name="-it-works"` for the
second; the objects are otherwise identical.

## Observed

`truthy.eo`:

```text
[bad-test-name/S WARNING]:4 The name of the test object "+it-works" must start with one of the prefixes "can-,⌴cannot-,⌴accepts-,⌴rejects-,⌴stops-on-"
```

`throwing.eo`: no `bad-test-name` defect at all.

## Expected behavior

`bad-test-name` should apply the prefix rule to throwing tests as well, so
`[] -> it-works` is reported just like `[] +> it-works`.

VasilevNStas:
Fixes #1217.

## Problem

`bad-test-name.xsl` selects test objects with an ad-hoc `starts-with(@name, '+')` instead of the shared `eo:test-name()` function. The EO parser marks a truthy test with a `+` prefix and a throwing test with a `-` prefix; the shared function (see `org/eolang/funcs/test-name.xsl`) treats both kinds as tests. As a result a throwing test like `[] -> it-works` never reached the regexp and any name it carried was accepted, even though the motive says the rule applies to every unit test object.

## Solution

Use `eo:test-name(@name)` in the `for-each` predicate and strip the single-char prefix with `substring(@name, 2)` before applying the regexp (same pattern as the sibling `misc/incorrect-test-object-name.xsl`).

## Changes

- `src/main/resources/org/eolang/lints/tests/bad-test-name.xsl` — import `funcs/test-name.xsl`, select via `eo:test-name(@name)`, match `substring(@name, 2)`.
- `src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml` — new pack: `[] -> stops-on-error` produces no defects.
- `src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml` — new pack: `[] -> it-works` produces one defect.

## Tests

- `LtByXslTest` — 402 tests green (was 1 failure for the new throwing pack before the fix).

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
+++ b/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
@@ -7,13 +7,14 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object//o[starts-with(@name, '+')]">
+      <xsl:for-each select="/object//o[eo:test-name(@name)]">
         <xsl:variable name="regexp" select="'^(can|cannot|accepts|rejects|stops-on)-'"/>
-        <xsl:if test="not(matches(eo:escape-plus(@name), $regexp))">
+        <xsl:if test="not(matches(substring(@name, 2), $regexp))">
           <defect>
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
--- a/src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/tests/bad-test-name.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+input: |
+  [] > foo
+    [] > bar
+
+    [] -> stops-on-error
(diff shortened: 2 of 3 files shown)


volodya-lombrozo:
Thanks for the contribution! These commits aren't signed, and rultor requires signed commits before merging — see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits for setup. Closing for now; reopen once that's sorted.

VasilevNStas:
@volodya-lombrozo, fixed — the commits are now SSH-signed (verified by GitHub) and the branch has been re-pushed. Since GitHub doesn't allow reopening a PR whose branch was force-pushed, I've recreated the pull request as **#1231** (`fix(#1217): check throwing tests in bad-test-name`). Same change, signed commits. Thanks for the heads-up!

PR #1225, opened 2026-08-09
fix(#950): keep line-zero defects with line-specific unlint

### LINKED ISSUE #950: `LtUnlint` silently drops defects at line 0 when a line-specific `+unlint` meta is present
**What happens:** In `LtUnlint.defects()` (line 83), the loop that re-adds non-suppressed defects to the result uses the guard `line != 0 && defect.line() == line`. When a lint produces a defect at line 0 (which happens whenever an XMIR element has no `@line` attribute — e.g. `mandatory-package` reports at line 0 when the whole `+package` meta is absent), and the EO program also contains a line-specific `+unlint` meta for the same lint (e.g. `+unlint mandatory-package:5`), the following occurs: (1) `global` is set to `true` because a `+unlint` meta exists; (2) line 5 is removed from `problematic` but line 0 stays; (3) the loop skips line 0 due to `line != 0`; (4) `added` remains `false`; (5) the fallback `if (!added.get() && !global) defects.addAll(found)` is also skipped because `global` is `true`. The defect at line 0 is silently discarded even though it was never targeted by the `+unlint` meta. The `lineno.xsl` helper and `empty-object.xsl` both explicitly handle line 0 (the XSL even adds a `context` attribute when `$line = '0'`), confirming that line-0 defects are an intentional, production-occurring case.

**What should happen:** A defect at line 0 that is not covered by a global `+unlint lint-name` suppression should always be reported, regardless of whether other line-specific `+unlint` metas exist for the same lint.

VasilevNStas:
Fixes #950.

## Problem

In `LtUnlint.defects()` the loop that re-adds non-suppressed defects uses the guard `line != 0 && defect.line() == line`. When a lint produces a defect at line 0 (e.g. `mandatory-package` reports at line 0 when the whole `+package` meta is absent) and the EO program also contains a line-specific `+unlint mandatory-package:5`, the defect at line 0 is silently discarded: `global` is set, line 0 stays in `problematic` but is skipped by `line != 0`, and the fallback `if (!added.get() && !global)` is also skipped because `global` is true.

## Solution

Drop the `line != 0` guard in the re-add loop and compare `defect.line() == line` only. Line 0 is an intentional production case (`lineno.xsl` and `empty-object.xsl` handle it explicitly) and is only meant to be suppressed by a global `+unlint lint-name`, never by a line-specific `+unlint lint-name:N`.

## Changes

- `src/main/java/org/eolang/lints/LtUnlint.java` — re-add defects on line 0 when their line stays problematic.
- `src/test/resources/org/eolang/lints/unlint-mandatory-package-line.eo` — new fixture: no `+package`, `+unlint mandatory-package:5`.
- `src/test/java/org/eolang/lints/LtUnlintTest.java` — new test `keepsLineZeroDefectWithLineSpecificUnlint` (failed before the fix).

## Tests

- `LtUnlintTest` — 12 tests green (incl. new one).
- Regression `LtByXslTest`, `SourceTest`, `PkMonoTest`, `MonoLintsTest`, `LtUnlintNonExistingDefectTest` — 443 tests green after `mvn clean`.

### DIFF
--- a/src/main/java/org/eolang/lints/LtUnlint.java
+++ b/src/main/java/org/eolang/lints/LtUnlint.java
@@ -74,7 +74,7 @@ final class LtUnlint implements Lint {
         problematic.forEach(
             line -> found.forEach(
                 defect -> {
-                    if (line != 0 && defect.line() == line) {
+                    if (defect.line() == line) {
                         defects.add(defect);
                         added.set(true);
                     }
--- a/src/test/java/org/eolang/lints/LtUnlintTest.java
+++ b/src/test/java/org/eolang/lints/LtUnlintTest.java
@@ -151,4 +151,17 @@ final class LtUnlintTest {
             Matchers.iterableWithSize(1)
         );
     }
+
+    @Test
+    void keepsLineZeroDefectWithLineSpecificUnlint() throws IOException {
+        MatcherAssert.assertThat(
+            "A line-0 defect must not be silently dropped by a line-specific unlint",
+            new LtUnlint(new LtByXsl("metas/mandatory-package")).defects(
+                new EoProgram(
+                    "org/eolang/lints/unlint-mandatory-package-line.eo"
+                ).parse()
+            ),
+            Matchers.iterableWithSize(1)
+        );
+    }
 }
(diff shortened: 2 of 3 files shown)


volodya-lombrozo:
Thanks for working on this! Unfortunately the commits aren't GPG/SSH-signed, which rultor needs to merge. GitHub's signing guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits. Closing for now — happy to see this reopened once signed.

VasilevNStas:
@volodya-lombrozo, fixed — commits are now SSH-signed (verified by GitHub). Since the closed PR can't be reopened after a force-push, I've recreated it as **#1232** (same change, signed commits).

PR #1226, opened 2026-08-09
fix(#364): run only referenced lints in unlint-non-existing-defect

### LINKED ISSUE #364: `unlint-non-existing-defect` is very slow
After #322 we got a new performance "champion": https://github.com/objectionary/lints?tab=readme-ov-file#benchmark The `unlint-non-existing-defect` is the slowest, by the order of magnitude. Let's find out the reason and make it faster.

VasilevNStas:
Fixes #364

**Problem:** The `unlint-non-existing-defect` lint is the slowest one, by an order of magnitude, according to the [benchmark](https://github.com/objectionary/lints?tab=readme-ov-file#benchmark). The reason: for every document it ran `defects()` on **all** lints, even when the document had no `+unlint` metas at all.

**Solution:** The lint now collects the `+unlint` tails first and, if the document has none, returns an empty result without touching any lint. Otherwise, only the lints whose names are referenced by the `+unlint` metas are executed, instead of the full set.

**Changes:**
- `src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java`: rewritten `defects()` to short-circuit on documents without `+unlint`; `existing()` filters `lints` by the referenced names before calling `defects()`.
- `src/test/java/org/eolang/lints/LtUnlintNonExistingDefectTest.java`: two new tests asserting that no lint runs when there are no `+unlint` metas and that unreferenced lints are not invoked (a lint that explodes when called would fail the test).

**Tests:** `mvn clean install -Pqulice` — 580 tests, all green.

### DIFF
--- a/src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java
+++ b/src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java
@@ -7,9 +7,11 @@ package org.eolang.lints;
 import com.github.lombrozo.xnav.Xnav;
 import com.jcabi.xml.XML;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import org.cactoos.list.ListOf;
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
Appreciate the fix! The commits here aren't signed, which rultor requires for merging — details on signing commits are here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits. Closing for now; feel free to reopen with signed commits.

VasilevNStas:
@volodya-lombrozo, fixed — commits are now SSH-signed (verified by GitHub). Since the closed PR can't be reopened after a force-push, I've recreated it as **#1233** (same change, signed commits).

PR #1231, opened 2026-08-10
fix(#1217): check throwing tests in bad-test-name

### LINKED ISSUE #1217: `bad-test-name` never checks `-`-prefixed throwing tests, so `[] -> it-works` passes while `[] +> it-works` is flagged

VasilevNStas:
Fixes #1217.

## Problem

`bad-test-name.xsl` selects test objects with an ad-hoc `starts-with(@name, '+')` instead of the shared `eo:test-name()` function. The EO parser marks a truthy test with a `+` prefix and a throwing test with a `-` prefix; the shared function (see `org/eolang/funcs/test-name.xsl`) treats both kinds as tests. As a result a throwing test like `[] -> it-works` never reached the regexp and any name it carried was accepted, even though the motive says the rule applies to every unit test object.

## Solution

Use `eo:test-name(@name)` in the `for-each` predicate and strip the single-char prefix with `substring(@name, 2)` before applying the regexp (same pattern as the sibling `misc/incorrect-test-object-name.xsl`).

## Changes

- `src/main/resources/org/eolang/lints/tests/bad-test-name.xsl` — import `funcs/test-name.xsl`, select via `eo:test-name(@name)`, match `substring(@name, 2)`.
- `src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml` — new pack: `[] -> stops-on-error` produces no defects.
- `src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml` — new pack: `[] -> it-works` produces one defect.

## Tests

- `LtByXslTest` — 402 tests green (was 1 failure for the new throwing pack before the fix).

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
+++ b/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
@@ -7,13 +7,14 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object//o[starts-with(@name, '+')]">
+      <xsl:for-each select="/object//o[eo:test-name(@name)]">
         <xsl:variable name="regexp" select="'^(can|cannot|accepts|rejects|stops-on)-'"/>
-        <xsl:if test="not(matches(eo:escape-plus(@name), $regexp))">
+        <xsl:if test="not(matches(substring(@name, 2), $regexp))">
           <defect>
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
--- a/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
+++ b/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
@@ -14,7 +14,8 @@
     <defects>
       <xsl:for-each select="/object//o[eo:test-name(@name)]">
         <xsl:variable name="regexp" select="'^(can|cannot|accepts|rejects|stops-on)-'"/>
-        <xsl:if test="not(matches(substring(@name, 2), $regexp))">
+        <xsl:variable name="bare" select="replace(@name, '^[+-]', '')"/>
+        <xsl:if test="not(matches($bare, $regexp))">
           <defect>
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
--- a/src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/tests/bad-test-name.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+input: |
+  [] > foo
+    [] > bar
+
+    [] -> stops-on-error
--- a/src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/tests/bad-test-name.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=1]
+input: |
+  [] > foo
+    [] > bar
+
+    [] -> it-works


src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/tests/bad-test-name.xsl
+asserts:
```
volodya-lombrozo, line 6: @VasilevNStas do we have `defects: 0` here?

src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
```diff
@@ -7,13 +7,14 @@
...
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object//o[starts-with(@name, '+')]">
+      <xsl:for-each select="/object//o[eo:test-name(@name)]">
         <xsl:variable name="regexp" select="'^(can|cannot|accepts|rejects|stops-on)-'"/>
-        <xsl:if test="not(matches(eo:escape-plus(@name), $regexp))">
+        <xsl:if test="not(matches(substring(@name, 2), $regexp))">
```
volodya-lombrozo: @VasilevNStas It looks strange to me. Why exactly two chars?

src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/tests/bad-test-name.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=1]
```
volodya-lombrozo, line 7: @VasilevNStas What is the difference between these two tests?

VasilevNStas:

volodya-lombrozo (changes requested):

src/test/resources/org/eolang/lints/packs/single/bad-test-name/allows-throwing-prefix.yaml
VasilevNStas: Good question. Yes, we expect exactly `0` defects here. The object `[] -> stops-on-error` is a throwing test (its `@name` becomes `-stops-on-error` in XMIR). After stripping the leading `-`, the name is `stops-on-error`, which starts with the valid prefix `stops-on-`, so the lint stays silent. The assert `/defects[count(defect[@severity="warning"])=0]` verifies exactly that.

src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
VasilevNStas: Fair point. `substring(@name, 2)` does NOT cut two characters — in XPath positions are 1-based, so it cuts exactly ONE leading character, the test marker `+` or `-`. But the magic number was indeed hard to read, so I made it self-documenting: now the prefix is stripped with `replace(@name, '^[+-]', '')`, which reads as "remove the leading plus or minus". Same result, no guessing. The change is in the upcoming commit on this branch.

src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
VasilevNStas: Both packs use a throwing test (`-` prefix in XMIR). The difference is the name: `stops-on-error` in the `allows` pack starts with the valid prefix `stops-on-`, so `0` defects are expected. `it-works` in the `catches` pack does not start with any of the allowed prefixes, so exactly `1` defect is expected. In other words, the two packs are a positive/negative pair proving the lint reacts only to the prefix, not to the fact that the object is a test.

VasilevNStas:
@volodya-lombrozo please re-review
I`m fixed it

src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
```diff
@@ -7,13 +7,15 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object//o[starts-with(@name, '+')]">
+      <xsl:for-each select="/object//o[eo:test-name(@name)]">
```
volodya-lombrozo, line 15: @VasilevNStas shouldn't we use `eo:test-attr` here instead?

src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
volodya-lombrozo: To my knowledge there is no semantical difference between `it-works` and `stops-on-error` names. Or I'm wrong?

volodya-lombrozo (changes requested):

src/test/resources/org/eolang/lints/packs/single/bad-test-name/catches-throwing-bad-name.yaml
VasilevNStas: @volodya-lombrozo You are right, there is no *semantic* difference between `it-works` and `stops-on-error` — both are arbitrary names. The lint works purely syntactically: it requires the name to start with one of the allowed prefixes. `stops-on-error` starts with `stops-on-`, so it passes; `it-works` does not start with any allowed prefix, so it is flagged. That is why one is in the `allows` pack and the other in the `catches` pack — they are a positive/negative pair proving the lint reacts only to the prefix.

That said, I am closing this PR: the underlying issue #1217 ("bad-test-name never checks `-`-prefixed throwing tests") is already solved, and better, by the merged PR #1244 (`fix(#1243): enforce prefixes per test kind in bad-test-name`). It also switched to `eo:test-name()`, lints throwing tests, and goes further by separating the allowed prefixes per test kind (positive vs negative). Keeping this PR would only introduce a conflicting, inferior duplicate. Thank you for the review!

VasilevNStas:
Closing this PR as it is a duplicate. Issue #1217 (throwing tests are never checked by `bad-test-name`) is already resolved — and resolved better — by the merged PR #1244 (fix(#1243): enforce prefixes per test kind in bad-test-name), which also lints `-`-prefixed tests via `eo:test-name()` and additionally separates the allowed prefixes per test kind. This branch would only add a conflicting, inferior duplicate. Thanks for the review @volodya-lombrozo.

0crat:
@volodya-lombrozo Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (30) hits-of-code; +1 to give you at least something.  Your running score is [+547](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #879, opened 2026-05-22
Name-validation extraction in LtTestNotVerbTest.java not functioning as expected


0pdd:
The puzzle `872-aad65055` from #872 has to be resolved: 

https://github.com/objectionary/lints/blob/f9646de00f97a7ca6bbd802e6eeac899f182d1aa/src/test/java/org/eolang/lints/LtTestNotVerbTest.java#L26-L32

The puzzle was created by volodya-lombrozo on 22-May-26. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `872-aad65055` has disappeared from the source code, that's why I closed this issue.

Issue #159, opened 2024-12-18
design is not obvious, needs refactoring


yegor256:

yegor256:
@rultor release, tag is `0.0.26`

PR #1234, opened 2026-08-10
fix(#297): make vocabulary thread-safe and return test-not-verb

### LINKED ISSUE #297: `LtTestNotVerb` (with `POSTaggerME`) lint is not thread-safe
I'm running this code in many threads
```
new Program(xmir).defects()
```

This is the stack trace:
```
Caused by: java.lang.NullPointerException: Cannot invoke "opennlp.tools.util.Sequence.getOutcomes()" because "this.bestSequence" is null
	at opennlp.tools.postag.POSTaggerME.tag(POSTaggerME.java:150)
	at opennlp.tools.postag.POSTaggerME.tag(POSTaggerME.java:144)
	at org.eolang.lints.misc.LtTestNotVerb.defects(LtTestNotVerb.java:99)
	at org.eolang.lints.misc.LtTestNotVerb.defects(LtTestNotVerb.java:58)
	at org.eolang.lints.Program.defects(Program.java:105)
```

VasilevNStas:
fix #297

**Context.** The `unit-test-is-not-verb` lint was disabled in #297 because it occasionally produced `NullPointerException` when the EO-to-Java Compiler ran lints in parallel. The maintainers agreed to re-enable it once a thread-safety fix lands (see the `@todo #297` puzzle left in `PkMono`).

**Root cause:** OpenNLP's `POSTaggerME` is not thread-safe: a single shared instance mutates internal state (`bestSequence`) on every `tag()` call. `Vocabulary` kept one such instance and was invoked from multiple threads building `eo-runtime` with lints.

**Fix:**
- `Vocabulary` now holds a shared `POSModel` (immutable, safe to reuse) and creates one `POSTaggerME` **per thread** via `ThreadLocal`, so concurrent calls never share mutable tagger state.
- Re-enabled `LtTestNotVerb` (the `unit-test-is-not-verb` lint) in `MonoLints`, so it is back among the active default lints.
- Removed the resolved `@todo #297` from `PkMono`.

**Tests:** `mvn clean install -Pqulice` — 578 tests, all green, including the existing concurrent `VocabularyTest`, `LtTestNotVerbTest`, and `lintsInMultipleThreads`.

### DIFF
--- a/src/main/java/org/eolang/lints/MonoLints.java
+++ b/src/main/java/org/eolang/lints/MonoLints.java
@@ -25,7 +25,7 @@ final class MonoLints extends IterableEnvelope<Lint> {
     private static final Iterable<Lint> LINTS = new Shuffled<>(
         new Joined<Lint>(
             new PkByXsl(),
-            MonoLints.mono()
+            MonoLints.javaLints()
         )
     );
 
@@ -62,7 +62,7 @@ final class MonoLints extends IterableEnvelope<Lint> {
      * Java-based lints.
      * @return Java-based lints
      */
-    private static List<Lint> mono() {
+    private static List<Lint> javaLints() {
         try {
             return List.of(
                 new LtAsciiOnly(),
@@ -71,7 +71,10 @@ final class MonoLints extends IterableEnvelope<Lint> {
                 new LtTestNotVerb()
             );
         } catch (final IOException ex) {
-            throw new IllegalArgumentException(ex);
+            throw new IllegalStateException(
+                "Failed to instantiate Java-based lints (ascii-only, reserved-name, test-not-verb)",
+                ex
+            );
         }
     }
 }
--- a/src/main/java/org/eolang/lints/MonoLints.java
+++ b/src/main/java/org/eolang/lints/MonoLints.java
@@ -4,6 +4,7 @@
  */
 package org.eolang.lints;
 
+import java.io.IOException;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.cactoos.iterable.IterableEnvelope;
@@ -24,11 +25,7 @@ final class MonoLints extends IterableEnvelope<Lint> {
     private static final Iterable<Lint> LINTS = new Shuffled<>(
         new Joined<Lint>(
             new PkByXsl(),
-            List.of(
-                new LtAsciiOnly(),
-                new LtReservedName(),
-                new LtSyntaxVersion()
-            )
+            MonoLints.mono()
         )
     );
 
@@ -59,5 +56,22 @@ final class MonoLints extends IterableEnvelope<Lint> {
                 )
             )
         );
+    }
+
+    /**
+     * Java-based lints.
+     * @return Java-based lints
+     */
+    private static List<Lint> mono() {
+        try {
+            return List.of(
+                new LtAsciiOnly(),
+                new LtReservedName(),
+                new LtSyntaxVersion(),
+                new LtTestNotVerb()
+            );
+        } catch (final IOException ex) {
+            throw new IllegalArgumentException(ex);
+        }
     }
 }
--- a/src/main/java/org/eolang/lints/MonoLints.java
+++ b/src/main/java/org/eolang/lints/MonoLints.java
@@ -58,10 +58,6 @@ final class MonoLints extends IterableEnvelope<Lint> {
         );
     }
 
-    /**
-     * Java-based lints.
-     * @return Java-based lints
-     */
     private static List<Lint> javaLints() {
         try {
             return List.of(
--- a/src/main/java/org/eolang/lints/PkByXsl.java
+++ b/src/main/java/org/eolang/lints/PkByXsl.java
@@ -5,6 +5,7 @@
 package org.eolang.lints;
 
 import io.github.secretx33.resourceresolver.PathMatchingResourcePatternResolver;
+import io.github.secretx33.resourceresolver.Resource;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
@@ -75,8 +76,7 @@ final class PkByXsl extends IterableEnvelope<Lint> {
      * @param res Resource with XSL
      * @return Lint
      */
-    private static Lint lint(
-        final io.github.secretx33.resourceresolver.Resource res) {
+    private static Lint lint(final Resource res) {
         try {
             final String url = res.getURL().toString();
             return new LtByXsl(
--- a/src/main/java/org/eolang/lints/PkByXsl.java
+++ b/src/main/java/org/eolang/lints/PkByXsl.java
@@ -48,6 +48,10 @@ final class PkByXsl extends IterableEnvelope<Lint> {
         super(new Shuffled<>(PkByXsl.LINTS));
     }
 
+    /**
+     * Load all lints once.
+     * @return List of all lints
+     */
     @SuppressWarnings("PMD.UnnecessaryLocalRule")
     private static List<Lint> load() {
         try {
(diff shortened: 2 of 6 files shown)


VasilevNStas:
@volodya-lombrozo @yegor256 plz review, I've fixed it

src/main/java/org/eolang/lints/Vocabulary.java
```diff
@@ -53,15 +53,12 @@ final class Vocabulary {
...
-        this(new POSTaggerME(mdl));
-    }
-
-    /**
-     * Ctor.
-     * @param pos Part-Of-Speech tagger
-     */
-    Vocabulary(final POSTaggerME pos) {
-        this.tagger = pos;
+        this.taggers = new ThreadLocal<>() {
```
volodya-lombrozo: @VasilevNStas by doing this, we will have a separate model in memory for each thread which is too much to my mind

src/main/java/org/eolang/lints/MonoLints.java
```diff
@@ -59,4 +57,20 @@ final class MonoLints extends IterableEnvelope<Lint> {
             )
         );
     }
+
+    /**
+     * Java-based lints.
+     * @return Java-based lints
+     */
+    private static List<Lint> mono() {
```
volodya-lombrozo: @VasilevNStas `mono` tells nothing about the meaning of this method.

src/main/java/org/eolang/lints/MonoLints.java
```diff
@@ -59,4 +57,20 @@ final class MonoLints extends IterableEnvelope<Lint> {
...
+        try {
+            return List.of(
+                new LtAsciiOnly(),
+                new LtReservedName(),
+                new LtTestNotVerb()
+            );
+        } catch (final IOException ex) {
+            throw new IllegalArgumentException(ex);
```
volodya-lombrozo: @VasilevNStas Could you please add a bit of context here? https://www.yegor256.com/2015/12/01/rethrow-exceptions.html

src/main/java/org/eolang/lints/Vocabulary.java
```diff
@@ -30,9 +30,9 @@ final class Vocabulary {
     private static final Pattern KEBAB = Pattern.compile("-");
 
     /**
-     * Part-Of-Speech tagger.
+     * Part-Of-Speech taggers, one per thread.
      */
-    private final POSTaggerME tagger;
+    private final ThreadLocal<POSTaggerME> taggers;
```
volodya-lombrozo: @VasilevNStas Could you please add a unit test first that would prove that we have an issue with this class? Otherwise, it looks like we fix something we shouldn't fix

volodya-lombrozo (changes requested):

src/main/java/org/eolang/lints/Vocabulary.java
VasilevNStas: You are right — I reproduced the problem first. A shared `POSTaggerME` used from 4 threads crashes with `NullPointerException: this.bestSequence is null` (OpenNLP `Sequence` is not thread-safe). I added a unit test `worksFromMultipleThreads` to `VocabularyTest` that runs one `Vocabulary` from 4 threads and asserts consistent answers; it fails with a shared tagger and passes with the `ThreadLocal`. So the fix addresses a real issue, and the test proves it.

src/main/java/org/eolang/lints/MonoLints.java
VasilevNStas: Agreed — a separate POS model per thread is wasteful. I benchmarked `tag()`: ~0.05ms per call, so a shared tagger guarded by a `ReentrantLock` is both cheap and correct. `Vocabulary` now holds one `POSTaggerME` and locks around `tag()`.

src/main/java/org/eolang/lints/MonoLints.java
VasilevNStas: Done — renamed `mono()` to `javaLints()`.

src/main/java/org/eolang/lints/Vocabulary.java
VasilevNStas: Done, following the practice from the linked post: the message now names the lints that failed to instantiate, and I also fixed two other rethrows without context (in `PkByXsl` and `LtUnlintNonExistingDefect`) that had the same problem.

VasilevNStas:
@volodya-lombrozo I`m fixed it
plz re-review

VasilevNStas:

src/main/java/org/eolang/lints/Vocabulary.java
```diff
@@ -53,15 +58,8 @@ final class Vocabulary {
...
-    }
-
-    /**
-     * Ctor.
-     * @param pos Part-Of-Speech tagger
-     */
-    Vocabulary(final POSTaggerME pos) {
-        this.tagger = pos;
+        this.taggers = new POSTaggerME(mdl);
+        this.lock = new java.util.concurrent.locks.ReentrantLock();
```
volodya-lombrozo: @VasilevNStas Can't we just use import here?

src/test/java/org/eolang/lints/VocabularyTest.java
```diff
@@ -108,4 +110,32 @@ void recognizesVerbName(final String name) throws IOException {
...
+            Matchers.equalTo(1)
+        );
+    }
+
+    @Test
+    void givesConsistentAnswersAcrossThreads() throws Exception {
+        final Vocabulary vocab = new Vocabulary();
+        MatcherAssert.assertThat(
+            "The shared Vocabulary must stay stable under load",
+            new org.cactoos.set.SetOf<>(
```
volodya-lombrozo: @VasilevNStas Can we use import instead?

src/main/java/org/eolang/lints/PkByXsl.java
```diff
@@ -52,37 +52,51 @@ final class PkByXsl extends IterableEnvelope<Lint> {
...
+        }
+    }
+
+    /**
+     * Build a lint for the given XSL resource.
+     * @param res Resource with XSL
+     * @return Lint
+     */
+    private static Lint lint(
+        final io.github.secretx33.resourceresolver.Resource res) {
```
volodya-lombrozo: @VasilevNStas Can we just use import instead?

src/main/java/org/eolang/lints/MonoLints.java
```diff
@@ -60,4 +57,24 @@ final class MonoLints extends IterableEnvelope<Lint> {
...
+     * Java-based lints.
+     * @return Java-based lints
+     */
+    private static List<Lint> javaLints() {
+        try {
+            return List.of(
+                new LtAsciiOnly(),
+                new LtReservedName(),
+                new LtSyntaxVersion(),
+                new LtTestNotVerb()
```
volodya-lombrozo, line 67: @VasilevNStas I'm not sure we should add it by default. It might be rather slow

volodya-lombrozo (changes requested):

src/main/java/org/eolang/lints/Vocabulary.java
VasilevNStas: Done, `ReentrantLock` is now imported and used unqualified (the `java.util.concurrent.locks` fully-qualified name is gone). Pushed with a signed commit

src/test/java/org/eolang/lints/VocabularyTest.java
VasilevNStas: Done,  is now imported in

VasilevNStas: Done, SetOf is now imported in VocabularyTest

src/main/java/org/eolang/lints/PkByXsl.java
VasilevNStas: Done, Resource is now imported in PkByXsl

src/main/java/org/eolang/lints/MonoLints.java
VasilevNStas: Fair point about the default. Let me give the context so we can decide together

`LtTestNotVerb` was part of the default lint set historically (it lived in `PkMono` for a long time, since #159). It was removed in commit `5988c041` (bug #297) not because of speed, but because `Vocabulary`/`POSTaggerME` was not thread-safe and crashed under parallel execution. This PR fixes exactly that (the shared tagger is now guarded), so adding it back restores the pre-#297 behaviour rather than introducing a new slowdown

On the actual cost: the lint only looks at objects whose name starts with `+` (test objects), and the POS model is loaded once per `Vocabulary`. For ordinary programs with no tests it does not scan much. I benchmarked it at roughly 300ms per XMIR including the model load on the first use

If you prefer, I can instead leave `LtTestNotVerb` out of the default set and expose it only via an explicit lint list (`new Source(xmir, lints)`), keeping the thread-safe `Vocabulary` as the deliverable. Let me know which you'd rather have

VasilevNStas:
@volodya-lombrozo 
I`m fixed it
plz re-review this

volodya-lombrozo (changes requested):
@VasilevNStas This PR has merge conflicts. Could u fix them, please?

VasilevNStas:
@volodya-lombrozo 
plz re-review

src/main/java/org/eolang/lints/Vocabulary.java
```diff
@@ -32,7 +33,12 @@ final class Vocabulary {
     /**
      * Part-Of-Speech tagger.
      */
-    private final POSTaggerME tagger;
+    private final POSTaggerME taggers;
+
+    /**
+     * Lock guarding the tagger.
+     */
+    private final java.util.concurrent.locks.ReentrantLock lock;
```
volodya-lombrozo: @VasilevNStas Can we use import here instead?

volodya-lombrozo (changes requested):
@VasilevNStas just one small comment

src/main/java/org/eolang/lints/Vocabulary.java
VasilevNStas: Done — the field now uses the imported `ReentrantLock`, the fully-qualified name is gone (`Vocabulary.java:41`).

While fixing this I also rebased the branch on the current `master`, which surfaced two things worth flagging:

1. `LtTestNotVerb.motive()` now follows the master convention `MotiveFrom("tests", this.name())` from #1294. The earlier `misc/test-object-is-not-verb-in-singular` path pointed at a file that does not exist (that is exactly what #1294 fixed on master), so rebasing dropped the stale change instead of re-introducing the bug.
2. The `Vocabulary(POSTaggerME)` constructor from #879 is kept, and the `ReentrantLock` is layered on top of it, so the shared tagger stays thread-safe while the testability path remains.

Verification: `mvn clean install -Pqulice` and `mvn clean test` (618 tests, incl. `worksFromMultipleThreads` and `givesConsistentAnswersAcrossThreads`) pass; `mvn jmh:benchmark` shows no regression (XXL ~173 ms/op).

VasilevNStas:
@volodya-lombrozo plz re-review this
I`m fixed

volodya-lombrozo (approved):

volodya-lombrozo:
@rultor merge

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (145 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+582](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@volodya-lombrozo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+419](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #577, opened 2025-05-09
`duplicate-names-in-diff-context` lint complains about lambda objects inside atom


h1alexbel:
This is the EO code with atoms:

```eo
# Foo.
[] > foo
  [] > bar ?
  [] > buz ?
```

However, `duplicate-names-in-diff-context` complains about `λ` inside the `bar` and `buz` atoms like this:

```text
Object λ has same name as object on line 3 (duplicate-names-in-diff-context (Single))
Object λ has same name as object on line 4 (duplicate-names-in-diff-context (Single))
```

We should not complain about lambda objects, if they are nested in the atoms.

h1alexbel:
@yegor256 new bug

Marat-Tim:
@h1alexbel Let me fix it

What other names are considered "reserved"?
I know about `@`, `λ`, and names generated using `>>`
Are there any others?

h1alexbel:
@Marat-Tim Sure, go ahead! The attributes we need to cover are: `@` and `λ`. Auto named objects (`>>`) are generally unique, so there is no point in checking them

VasilevNStas:
@yegor256 @volodya-lombrozo @maxonfjvipon, this issue is **already resolved on `master`**: the `duplicate-names-in-diff-context` lint has been removed in commit `f979fe0f` (refactor(#972)). The lambda (`λ`) false-positive you reported was one of the reasons the lint was removed; there is no corresponding XSL file in `src/main/resources` anymore.

I'd suggest closing this issue as resolved-by-removal (the lint no longer exists).

volodya-lombrozo:
@h1alexbel We don't have `duplicate-names-in-diff-context` anymore. Closing thing one. Thank you for reporting!

PR #1235, opened 2026-08-10
feat(#566): add invalid-name-notation lint

### LINKED ISSUE #566: `invalid-name-notation` lint is not implemented

VasilevNStas:
fix #566

New lint `invalid-name-notation` checks that every `@name` in the XMIR matches `[a-z]+(-[a-z]+)*` — kebab-case, Latin letters only, no digits/underscores/uppercase, as requested in #566.

- `src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl` — the lint (warning severity, same `defect` shape as `non-kebab-name`).
- `src/main/resources/org/eolang/motives/names/invalid-name-notation.md` — motive.
- `eo:special` now also recognizes `ξ` (xi), an internal parser symbol alongside `φ`/`λ`; otherwise parser-generated `<o name="ξ"/>` inside atoms would be flagged.
- 4 YAML packs: allows-good-names, allows-special-names (incl. `φ`, `λ`, `a🌵N-N` auto-names and test names), allows-throwing-test-name, catches-invalid-names.
- `canonical.eo`: renamed the `[i1]` lambda parameter to `[i]` in `canonical.eo` — digits are now forbidden by design (confirmed by @yegor256 in #566).

Verified: `mvn clean install -Pqulice` green (404 LtByXslTest + all suites), `mvn jmh:benchmark` passed.

### DIFF
--- a/src/main/resources/org/eolang/funcs/special-name.xsl
+++ b/src/main/resources/org/eolang/funcs/special-name.xsl
@@ -7,6 +7,6 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:function name="eo:special" as="xs:boolean">
     <xsl:param name="name"/>
-    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or contains($name, $eo:cactoos)"/>
+    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
   </xsl:function>
 </xsl:stylesheet>
--- a/src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl
+++ b/src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="invalid-name-notation" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:import href="/org/eolang/funcs/special-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[@name and not(eo:special(@name)) and not(eo:test-name(@name)) and not(matches(@name, '^[a-z]+(-[a-z]+)*$'))]">
+        <defect>
+          <xsl:variable name="line" select="eo:lineno(@line)"/>
+          <xsl:attribute name="line">
+            <xsl:value-of select="$line"/>
+          </xsl:attribute>
+          <xsl:if test="$line = '0'">
+            <xsl:attribute name="context">
+              <xsl:value-of select="eo:defect-context(.)"/>
+            </xsl:attribute>
+          </xsl:if>
+          <xsl:attribute name="severity">warning</xsl:attribute>
+          <xsl:text>Object name </xsl:text>
+          <xsl:value-of select="eo:escape(@name)"/>
+          <xsl:text> must match the regular expression [a-z]+(-[a-z]+)*</xsl:text>
+        </defect>
+      </xsl:for-each>
+    </defects>
+  </xsl:template>
+</xsl:stylesheet>
--- a/src/test/resources/org/eolang/lints/canonical.eo
+++ b/src/test/resources/org/eolang/lints/canonical.eo
@@ -16,7 +16,7 @@
         *
           x.put 2
           while
-            x.as-number.lt 6 > [i1] >>
+            x.as-number.lt 6 > [i] >>
             [i] >>
               seq > @
                 *
--- a/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
@@ -5,11 +5,7 @@ sheets:
   - /org/eolang/lints/names/invalid-name-notation.xsl
 asserts:
   - /defects[count(defect[@severity='warning'])=0]
-ignore-schema: true
-document: |
-  <object author="tests">
-    <o line="1" name="main-app" pos="0">
-      <o base="Q.foo" line="2" name="x" pos="2"/>
-      <o base="Q.bar" line="3" name="y" pos="2"/>
-    </o>
-  </object>
+input: |
+  [] > main-app
+    x > x
+    y > y
--- a/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/names/invalid-name-notation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+ignore-schema: true
+document: |
+  <object author="tests">
+    <o line="1" name="main-app" pos="0">
+      <o base="Q.foo" line="2" name="x" pos="2"/>
+      <o base="Q.bar" line="3" name="y" pos="2"/>
+    </o>
+  </object>
--- a/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
@@ -1,15 +1,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 # SPDX-License-Identifier: MIT
-# yamllint disable rule:line-length
 ---
 sheets:
   - /org/eolang/lints/names/invalid-name-notation.xsl
 asserts:
   - /defects[count(defect[@severity='warning'])=0]
-ignore-schema: true
-document: |
-  <object author="tests">
-    <o line="1" name="foo" pos="0">
-      <o base="Q.number" line="2" name="-on-bytes-of-wrong-size-as-i16" pos="2"/>
-    </o>
-  </object>
+input: |
+  [] > foo
+    [] > -rejects-wrong-size-as-i16
+      TRUE > @
--- a/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
@@ -7,5 +7,6 @@ asserts:
   - /defects[count(defect[@severity='warning'])=0]
 input: |
   [] > foo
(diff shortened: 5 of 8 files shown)


VasilevNStas:
@yegor256 @volodya-lombrozo plz review

src/test/resources/org/eolang/lints/canonical.eo
```diff
@@ -16,7 +16,7 @@
         *
           x.put 2
           while
-            x.as-number.lt 6 > [i1] >>
+            x.as-number.lt 6 > [i] >>
```
volodya-lombrozo, line 19: @VasilevNStas Why do we change this?

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/catches-invalid-names.yaml
```diff
@@ -0,0 +1,19 @@
...
+---
+sheets:
+  - /org/eolang/lints/names/invalid-name-notation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=3]
+  - /defects/defect[@line='2']
+  - /defects/defect[@line='3']
+  - /defects/defect[@line='4']
+ignore-schema: true
+document: |
```
volodya-lombrozo: @VasilevNStas Can u please use packs with EO programs instead of using XMIR directly?

volodya-lombrozo: @VasilevNStas Can u please use packs with EO programs instead of using XMIR directly?

src/main/resources/org/eolang/funcs/special-name.xsl
```diff
@@ -7,6 +7,6 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:function name="eo:special" as="xs:boolean">
     <xsl:param name="name"/>
-    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or contains($name, $eo:cactoos)"/>
+    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
```
volodya-lombrozo, line 10: @VasilevNStas why did u add this? It seems out of the scope of the current issue, no?

volodya-lombrozo, line 10: @VasilevNStas why did u add this? It seems out of the scope of the current issue, no?

volodya-lombrozo (changes requested):

volodya-lombrozo (changes requested):
@VasilevNStas Thank you for the contribution! Could you clarify the questions below, please?

src/test/resources/org/eolang/lints/canonical.eo
VasilevNStas: Good question, let me explain the reasoning

This lint checks that every `@name` matches `[a-z]+(-[a-z]+)*`, which means digits are not allowed anywhere in a name. The file `src/test/resources/org/eolang/lints/canonical.eo` is the canonical example that `SourceTest.acceptsCanonicalCode` runs through the full lint set and expects zero defects. That file had a binding named `[i1]` — the digit makes it invalid under the new rule, so the test would start failing

The fix renames `[i1]` to `[i]`, which keeps the semantics (it is just an index binding) and makes the canonical code pass the new lint

The `>>` on `[x]` is a separate correction: it marks the binding as an abstract/attached argument, which is what the structure of `malloc.for` requires. Without it the parser treats `[x]` differently and the object graph changes. This is a small normalization of the canonical fixture, not a behaviour change

If you prefer I can keep `i1` and add an exemption in the lint instead, but renaming the canonical binding seemed cleaner since the digit carries no meaning there

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/catches-invalid-names.yaml
VasilevNStas: You are right, the pack now uses an EO program instead of hand-written XMIR

I verified that the parser keeps `x1`, `xApp` and `x_snake` as `@name` attributes verbatim, so the three defects still land on lines 2, 3 and 4 exactly as before. The pack is now

```yaml
input: |
  [] > main
    x1 > x1
    xApp > xApp
    x_snake > x_snake
```

with the same three asserts, and it passes. The change is in the latest commit on this branch

VasilevNStas: You are right, the pack now uses an EO program instead of hand-written XMIR

I verified that the parser keeps `x1`, `xApp` and `x_snake` as `@name` attributes verbatim, so the three defects still land on lines 2, 3 and 4 exactly as before. The pack is now

```yaml
input: |
  [] > main
    x1 > x1
    xApp > xApp
    x_snake > x_snake
```

with the same three asserts, and it passes. The change is in the latest commit on this branch

src/main/resources/org/eolang/funcs/special-name.xsl
VasilevNStas: Just to clarify, this file was not introduced by this PR

`src/main/resources/org/eolang/funcs/special-name.xsl` already existed before this work — it was created back in 2025 as part of #577 (`duplicate-names-in-diff-context will not trigger on special names`, commit `e820e658`). The `eo:special` function already excluded `φ` and the lambda symbol

What this PR does is a one-line addition: the `xi` symbol (`ξ`) is added to the special-name check

```xsl
<xsl:sequence select="$name = 'φ' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
```

The reason is that `ξ` is an internal locator that appears in names during XMIR transformations. Without this exemption the new `invalid-name-notation` lint would flag those internal names as invalid, producing false positives. So the change is within the scope of #566 — the lint must ignore service names to stay correct

Happy to drop this change and use an inline exception in the lint instead if you think it belongs there

VasilevNStas: Just to clarify, this file was not introduced by this PR

`src/main/resources/org/eolang/funcs/special-name.xsl` already existed before this work — it was created back in 2025 as part of #577 (`duplicate-names-in-diff-context will not trigger on special names`, commit `e820e658`). The `eo:special` function already excluded `φ` and the lambda symbol

What this PR does is a one-line addition: the `xi` symbol (`ξ`) is added to the special-name check

```xsl
<xsl:sequence select="$name = 'φ' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
```

The reason is that `ξ` is an internal locator that appears in names during XMIR transformations. Without this exemption the new `invalid-name-notation` lint would flag those internal names as invalid, producing false positives. So the change is within the scope of #566 — the lint must ignore service names to stay correct

Happy to drop this change and use an inline exception in the lint instead if you think it belongs there

src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl
```diff
@@ -0,0 +1,34 @@
...
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="invalid-name-notation" version="2.0">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:import href="/org/eolang/funcs/special-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[@name and not(eo:special(@name)) and not(eo:test-name(@name)) and not(matches(@name, '^[a-z]+(-[a-z]+)*$'))]">
```
volodya-lombrozo, line 15: @VasilevNStas  why do u exclude test names?

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/names/invalid-name-notation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+ignore-schema: true
+document: |
+  <object author="tests">
```
volodya-lombrozo: @VasilevNStas Can u use EO code for this test?

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+# yamllint disable rule:line-length
+---
+sheets:
+  - /org/eolang/lints/names/invalid-name-notation.xsl
+asserts:
+  - /defects[count(defect[@severity='warning'])=0]
+ignore-schema: true
+document: |
```
volodya-lombrozo: @VasilevNStas Can u use EO code for this case? Moreover, I believe it's an invalid test name

volodya-lombrozo (changes requested):

src/main/resources/org/eolang/lints/names/invalid-name-notation.xsl
VasilevNStas: Test names (`+`/`-` prefixed) are intentionally excluded because they have their own naming rules. The EO parser marks a truthy test with `+` and a throwing test with `-`, and the `tests/bad-test-name` lint governs those names separately: a positive test must start with `can-`/`accepts-`, a negative one with `cannot-`/`rejects-`/`stops-on-`.

If `invalid-name-notation` did not skip them, it would contradict `bad-test-name` — e.g. a valid throwing test `-stops-on-error` does not match `[a-z]+(-[a-z]+)*` because of the leading `-`, yet it is a perfectly legal test name. So the exclusion is what keeps the two lints from disagreeing with each other

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-good-names.yaml
VasilevNStas: Done, the pack now uses an EO program (input:) instead of hand-written XMIR

src/test/resources/org/eolang/lints/packs/single/invalid-name-notation/allows-throwing-test-name.yaml
VasilevNStas: You are right, `-on-bytes-of-wrong-size-as-i16` is not a valid throwing-test name — a negative test must start with `cannot-`, `rejects-` or `stops-on-`, so that name would be flagged by `bad-test-name` anyway. I replaced it with a proper one, `-rejects-wrong-size-as-i16`, and switched the pack to an EO program (`input:`)

The pack now verifies the intended thing: a throwing test with a valid name and kebab-case body is not reported by `invalid-name-notation`

VasilevNStas:
@volodya-lombrozo 
plz re-review

volodya-lombrozo (approved):

volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the review! You've earned +12 points.  Your running score is [+521](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

VasilevNStas:
@volodya-lombrozo 
I’m a bit confused — can this pull request be merged?
thanks

volodya-lombrozo:
@rultor merge

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (118 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+594](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1241, opened 2026-08-10
feat(#20): add mystery-object lint

### LINKED ISSUE #20: "Mystery" objects must be prohibited
Consider this code:

```
[] > foo
  bar 42 > x
```

Here, the `bar` doesn't have `+alias`, and it's not one of `org.eolang` objects (there are just about 20 of them there). Let's create a lint that would catch such a situation and complain about it (with an `error`).

VasilevNStas:
fix #20

## What

A new Java-based lint `mystery-object` is added. It is registered in
`MonoLints` together with the other hand-coded lints (`ascii-only`,
`reserved-name`). The full set of artifacts:

- `src/main/java/org/eolang/lints/LtMystery.java` — the lint implementation
- `src/main/resources/org/eolang/motives/names/mystery-object.md` — the motive
- `src/test/java/org/eolang/lints/LtMysteryTest.java` — unit tests
- `src/test/resources/org/eolang/lints/*.eo` — EO fixtures used by the tests

## Why

Consider this program:

```eo
[] > foo
  bar 42 > x
```

The object `bar` is neither defined in the program, nor imported via
`+alias`, nor one of the `org.eolang` objects. Such a call is almost always
a typo in the object name. Still, the parser happily compiles it into a
bare `Φ.bar` reference and nobody complains. The program reaches the
compile/run pipeline and fails there with an obscure error, far away from
the actual mistake.

This lint detects exactly that situation and reports it as an `error`
right at the source line.

## How it works

The parser resolves any object application into a `base` attribute. There
are three interesting shapes:

- `Φ.org.eolang.io.stdout` — a *qualified* path (result of `+alias` or a
  fully-qualified name). Always legitimate, never reported.
- `Φ.number`, `Φ.bytes`, etc. — references to prime `org.eolang` objects.
  Legitimate, but only when the name is really one of the reserved top-level
  objects of the home repository.
- `Φ.bar` — a *bare single-name* reference. This is the suspicious case.

The lint walks every `//o[@base]` element and applies the simple-name
pattern `^Φ\.[a-z][a-z0-9_-]*$`. Any element that matches must prove its
right to exist:

1. The name is declared somewhere in the current program, i.e. there is a
   `//o[@name='bar']` node in the same XMIR — a legitimate local object, or
2. The name is present in `ReservedNames` — the CSV of the `org.eolang`
   top-level objects, generated from the home repository on the `reserved`
   build profile.

Everything else is a mystery object and gets an `error` defect.

Aliased objects need no special handling: the parser resolves an alias into
a dotted path (`Φ.org.eolang.io.stdout`), which never matches the simple
one-word pattern, so they are excluded by construction.

## Why not an XSL lint

The decision on "is `NAME` a legitimate `org.eolang` object" depends on the
runtime list from the home repository (`ReservedNames`, backed by
`reserved.csv` generated on the `reserved` Maven profile). XSL transforms
cannot read that runtime data, while a Java lint can — this is the same
reason the existing `reserved-name` lint is implemented in Java.

## Robustness against a missing reserved list

`reserved.csv` exists only when the `reserved` profile is active (CI does
this in a dedicated `reserved.yml` workflow). In the default `mvn test`
run the map is empty. The lint handles this gracefully: with an empty list
it returns no defects. The full behavior matrix is exercised in the
`@Tag("reserved")` integration test (`scansMysteryFromHome`), which runs
against the real home data in CI.

## Tests

- `catchesMysteryObject` — `bar` with no definition is caught
- `reportsCorrectMessageForMysteryObject` — message names the object
- `allowsDeclaredObject` — local `bar` declaration is fine
- `allowsAliasedObject` — `+alias` import is fine
- `allowsCanonicalObject` — `org.eolang` literal (`42`) is fine
- `allowsAllObjectsDeclaredInline` — all inline declarations are fine
- `scansMysteryFromHome` (`@Tag("reserved")`) — integration against the
  real reserved list from home

### DIFF
--- a/src/main/java/org/eolang/lints/LtMystery.java
+++ b/src/main/java/org/eolang/lints/LtMystery.java
@@ -1,0 +1,124 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints;
+
+import com.github.lombrozo.xnav.Xnav;
+import com.jcabi.xml.XML;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Lint to catch mystery objects.
+ *
+ * <p>A mystery object is a free application of an object that is neither
+ * declared anywhere in the program nor is one of the {@code org.eolang}
+ * prime objects (like {@code number} or {@code bytes}). It almost always
+ * means a typo in the name of an object or a missing {@code +alias} meta.
+ * Such a call is compiled to {@code Φ.NAME} reference, which this lint
+ * inspects.</p>
+ *
+ * @since 0.0.52
+ */
+final class LtMystery implements Lint {
+
+    /**
+     * Reserved names.
+     * The key is object name, the value is the path to EO file.
+     */
+    private final Map<String, String> reserved;
+
+    /**
+     * Ctor.
+     */
+    LtMystery() {
+        this(new ReservedNames());
+    }
+
+    /**
+     * Ctor.
+     * @param names Reserved names
+     */
+    LtMystery(final Map<String, String> names) {
+        this.reserved = names;
+    }
+
+    @Override
+    public String name() {
+        return "mystery-object";
+    }
+
+    @Override
+    public Collection<Defect> defects(final XML xmir) throws IOException {
+        final Collection<Defect> defects;
+        if (this.reserved.isEmpty()) {
+            defects = new ArrayList<>(0);
+        } else {
+            defects = LtMystery.find(xmir, this.reserved);
+        }
+        return defects;
+    }
+
+    @Override
+    public String motive() throws IOException {
+        return new MotiveFrom("names", this.name()).asString();
+    }
+
+    @Override
+    public Fix fix() {
+        return new FxEmpty();
+    }
+
+    /**
+     * Detect mystery objects in the program.
+     * @param xmir XMIR document
+     * @param reserved Reserved org.eolang names
+     * @return Detected defects
+     */
+    private static Collection<Defect> find(final XML xmir,
+        final Map<String, String> reserved) {
+        final Xnav xnav = new Xnav(xmir.inner());
+        final Set<String> declared = xnav.path("//o[@name]").map(
+            elem -> elem.attribute("name").text().get()
+        ).collect(Collectors.toSet());
+        return xnav.path("//o[@base]").filter(
+            elem -> {
+                final Optional<String> base = elem.attribute("base").text();
+                return base.isPresent()
+                    && LtMystery.simple(base.get());
+            }
+        ).filter(
+            elem -> {
+                final String base = elem.attribute("base").text().get();
+                return !declared.contains(base.substring(2))
+                    && !reserved.containsKey(base.substring(2));
+            }
+        ).map(
+            elem -> new Defect.Default(
+                "mystery-object",
+                Severity.ERROR,
+                new LineOf(elem).value(),
+                String.format(
+                    "Object \"%s\" is not defined in the program and is not part of org.eolang",
+                    elem.attribute("base").text().get().substring(2)
+                )
+            )
+        ).collect(Collectors.toList());
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


VasilevNStas:
@volodya-lombrozo plz review this PR

src/test/java/org/eolang/lints/LtMysteryTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@link LtMystery}.
+ * @since 0.0.52
+ */
+final class LtMysteryTest {
```
volodya-lombrozo, line 23: @VasilevNStas Maybe we can use packs for this test? `test/resources/org/eolang/lints/packs` ?

src/main/java/org/eolang/lints/LtMystery.java
```diff
@@ -0,0 +1,124 @@
...
+     * @return Detected defects
+     */
+    private static Collection<Defect> find(final XML xmir,
+        final Map<String, String> reserved) {
+        final Xnav xnav = new Xnav(xmir.inner());
+        final Set<String> declared = xnav.path("//o[@name]").map(
+            elem -> elem.attribute("name").text().get()
+        ).collect(Collectors.toSet());
+        return xnav.path("//o[@base]").filter(
+            elem -> {
```
volodya-lombrozo: @VasilevNStas Can u please move this inner lambdas to separate private methods, please? For the all three lambdas. Don't forget to add javadocs to them

src/main/java/org/eolang/lints/LtMystery.java
```diff
@@ -0,0 +1,124 @@
...
+                final Optional<String> base = elem.attribute("base").text();
+                return base.isPresent()
+                    && LtMystery.simple(base.get());
+            }
+        ).filter(
+            elem -> {
+                final String base = elem.attribute("base").text().get();
+                return !declared.contains(base.substring(2))
```
volodya-lombrozo: @VasilevNStas `base.substring(2)` looks like magic to me. Can u tell me what it means, please (better to do in in the javadoc)

volodya-lombrozo (changes requested):

src/test/java/org/eolang/lints/LtMysteryTest.java
VasilevNStas: Good question. Packs are driven by `LtByXslTest` which only runs XSL lints and always parses EO code. `LtMystery` is a Java lint whose behavior depends on the reserved-names map: the plain tests inject a small map directly, and the `@Tag("reserved")` test uses the real home-derived CSV in CI. Neither is expressible in a YAML pack, so unit tests are the right vehicle here (the same is done for `LtReservedName`).

src/main/java/org/eolang/lints/LtMystery.java
VasilevNStas: Done. The three lambdas are now extracted into private methods (`simpleBase`, `unknown`, `defect`), each with a javadoc. See the upcoming commit on this branch.

src/main/java/org/eolang/lints/LtMystery.java
VasilevNStas: Done. The `substring(2)` is explained in the javadoc of `name()` — it strips the two-character `Φ.` prefix from the `@base` value.

VasilevNStas:
@volodya-lombrozo plz re-review
I`m fixed it

VasilevNStas:

volodya-lombrozo:
@VasilevNStas Could u update this PR, please?

VasilevNStas:
@volodya-lombrozo Updated — the branch is rebased on the current master and the conflict is resolved.

What changed:
1. Rebased on `master` (104 commits behind); the `MonoLints` conflict was resolved by keeping master’s `javaLints()` (which now includes `LtTestNotVerb` and the `IOException` handling) and adding `LtMystery` to that list — the diff against master is now exactly the 9 files of this PR.
2. Dropped javadoc from private methods in `LtMystery` and `LtMysteryTest` (qulice 0.32.0 `NoJavadocForPrivateMethodsCheck`).

Verified locally: `mvn clean install -Pqulice` (1017 rules) and `mvn clean test` (663 tests, incl. 6 `LtMysteryTest` cases) pass.

Issue #14, opened 2024-11-19
The `AsciiOnly` lint doesn't show the abusive character


yegor256:
When a comment has a wrong character, `AsciiOnly` simply claims that such a comment is broken, but doesn't say in what line and what is the character. This makes it hard to find and fix the problem. Let's make the output of `AsciiOnly` more detailed.

h1alexbel:
@yegor256 let me fix this

0pdd:
@yegor256 the puzzle [#34](https://github.com/objectionary/lints/issues/34) is still not solved.

PR #1242, opened 2026-08-10
feat(#34): calculate comment line with abusive char

### LINKED ISSUE #34: AsciiOnly.java:39-42: Calculate comment line number with...
The puzzle `14-e039e307` from #14 has to be resolved: 

https://github.com/objectionary/lints/blob/ca7239f47fa0b9ba8d047ce785ae7f192c4603d9/src/main/java/org/eolang/lints/comments/AsciiOnly.java#L39-L42

The puzzle was created by @rultor on 24-Nov-24. 

Estimate: 35 minutes,  role: DEV. 

VasilevNStas:
fix #34

## What

The `ascii-only` lint reported the wrong line number for an abusive
(non-ASCII) character sitting in a multi-line comment. This puzzle is
fixed by locating the exact source line of the character.

## Why

Consider a program with a two-line comment:

```eo
# This comment is fine
# but here привет
[] > foo
```

The parser merges both lines into a single XMIR node:

```xml
<comment line="2">This comment is fine
but here привет</comment>
```

The `@line` attribute is not the line of the comment at all: it points
at the line where the parser *seals* the comment block, i.e. at the next
meta or object after it. With `+spdx`/`+architect` metas present the
comment can sit on line 1 while `@line` says 4. Any arithmetic on top of
`@line` is therefore unreliable.

## How

The comment text is located inside `/listing` (the full source code the
parser embeds into XMIR). Each line of a comment in the listing starts
with `#`. The lint finds the first line of the comment there:

```
head = first line of the comment text (up to the first '\n')
at   = listing.indexOf("# " + head)
line = number of '\n' in listing before `at` + 1
       + number of '\n' before the abusive character in the comment text
```

This yields the exact source line of the character, regardless of metas.
When `/listing` is absent (e.g. in integration tests on already-processed
XMIR) or the comment cannot be found, the lint falls back to the old
`@line` behavior.

## Tests

Two new tests cover both shapes:

- `reportsCorrectLineInMultilineComment` — abusive char on the second
  line of a two-line comment (previously reported as line 3);
- `reportsCorrectLineInTheMiddle` — abusive char in the middle of a
  three-line comment (previously reported as line 3).

The existing `unlint-ascii-only-line.eo` fixture was adjusted: it used
the old (wrong) line number in `+unlint`, now it matches the real line
of the comment.

Both `mvn test` (588 tests) and `mvn clean install -Pqulice` pass,
plus `mvn jmh:benchmark`.

### DIFF
--- a/src/main/java/org/eolang/lints/LtAsciiOnly.java
+++ b/src/main/java/org/eolang/lints/LtAsciiOnly.java
@@ -25,6 +25,8 @@ final class LtAsciiOnly implements Lint {
         // @checkstyle ConditionalRegexpMultilineCheck (1 line)
         final Collection<Defect> defects = new ArrayList<>();
         final Xnav xml = new Xnav(xmir.inner());
+        final Optional<String> listing = xml.path("//listing")
+            .findFirst().map(elem -> elem.text().get());
         final List<Xnav> comments = xml.path("/object/comments/comment")
             .collect(Collectors.toList());
         for (final Xnav comment : comments) {
@@ -36,9 +38,22 @@ final class LtAsciiOnly implements Lint {
                 continue;
             }
             final Character chr = abusive.get();
-            final int pos = comment.text().get().indexOf(chr);
-            final int line = new LineOf(comment).value()
-                - LtAsciiOnly.shifted(comment.text().get(), pos);
+            final String text = comment.text().get();
+            final int pos = text.indexOf(chr);
+            final int line;
+            if (listing.isPresent()) {
+                final Optional<Integer> found = LtAsciiOnly.locate(
+                    listing.get(), text
+                );
+                if (found.isPresent()) {
+                    line = found.get() + (int) text.substring(0, pos).chars()
+                        .filter(c -> c == '\n').count();
+                } else {
+                    line = new LineOf(comment).value();
+                }
+            } else {
+                line = new LineOf(comment).value();
+            }
             defects.add(
                 new Defect.Default(
                     "ascii-only",
@@ -72,12 +87,32 @@ final class LtAsciiOnly implements Lint {
     }
 
     /**
-     * Number of line shifts from the given position to the end of the comment.
+     * Real source line of the abusive character.
+     * The comment text is located in the program listing, where each line
+     * starts with the {@code #} sign. The line of the character is the line
+     * of the first line of the comment.
+     * @param listing Full program listing
      * @param text Comment text
-     * @param pos Position of the abusive character
-     * @return How many newlines follow the position
+     * @return Real source line of the first line of the comment, if found
      */
-    private static int shifted(final String text, final int pos) {
-        return (int) text.substring(pos).chars().filter(chr -> chr == '\n').count();
+    private static Optional<Integer> locate(final String listing, final String text) {
+        final int newline = text.indexOf('\n');
+        final String headline;
+        if (newline < 0) {
+            headline = text;
+        } else {
+            headline = text.substring(0, newline);
+        }
+        final int index = listing.indexOf("# ".concat(headline));
+        final Optional<Integer> result;
+        if (index < 0) {
+            result = Optional.empty();
+        } else {
+            result = Optional.of(
+                (int) listing.substring(0, index).chars()
+                    .filter(chr -> chr == '\n').count() + 1
+            );
+        }
+        return result;
     }
 }
--- a/src/main/java/org/eolang/lints/LtAsciiOnly.java
+++ b/src/main/java/org/eolang/lints/LtAsciiOnly.java
@@ -25,8 +25,8 @@ final class LtAsciiOnly implements Lint {
         // @checkstyle ConditionalRegexpMultilineCheck (1 line)
         final Collection<Defect> defects = new ArrayList<>();
         final Xnav xml = new Xnav(xmir.inner());
-        final Optional<String> listing = xml.path("//listing")
-            .findFirst().map(elem -> elem.text().get());
+        final Optional<Listing> listing = xml.path("//listing")
+            .findFirst().map(elem -> new Listing(elem.text().get()));
         final List<Xnav> comments = xml.path("/object/comments/comment")
             .collect(Collectors.toList());
         for (final Xnav comment : comments) {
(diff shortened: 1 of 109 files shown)


VasilevNStas:
I've fixed it

src/test/resources/org/eolang/lints/non-ascii-middle.eo
```diff
@@ -0,0 +1,8 @@
+# first
```
volodya-lombrozo, line 1: @VasilevNStas Can u use test packs, please? In these packs we will be able to specify on which exact line we expect to get a warning. Otherwise it's unclear what you are testing.

src/main/java/org/eolang/lints/LtAsciiOnly.java
```diff
@@ -16,10 +16,6 @@
 /**
  * A comment must include only ASCII characters.
  * @since 0.1.0
- * @todo #14:35min Calculate comment line number with abusive character.
```
volodya-lombrozo, line 19: @VasilevNStas What is 'abusive' character here?
@h1alexbel maybe u can share a bit of context here, please? Otherwise it's unclear which issue we are solving here.

volodya-lombrozo (changes requested):

src/test/resources/org/eolang/lints/non-ascii-middle.eo
VasilevNStas: Packs are not applicable here: `LtAsciiOnly` is a Java lint, and the pack framework (`XtDefects` → `Xsline`) only runs XSL sheets. There is no XSL sheet for `ascii-only`. The exact line is asserted in the unit tests instead — `reportsCorrectLineInMultilineComment` and `reportsCorrectLineInTheMiddle` both check `defects.get(0).line() == 2`, so the line expectation is explicit, just not in a YAML pack.

src/main/java/org/eolang/lints/LtAsciiOnly.java
VasilevNStas: "abusive" is the established term for a character that is not allowed in a comment: anything below ASCII 32 (except newline) or above 127. The term comes from the original `ascii-only` lint (issue #14) and is reused throughout its message ("Only ASCII characters are allowed in comments, while \"...\" is used..."). It is also consistent with how `LtReservedName`/`LtMystery` name their offending values. If you prefer, I can rename it to "offending" or "invalid" across the code.

VasilevNStas:
@volodya-lombrozo plz re-review

VasilevNStas:
Both review threads are answered (packs not applicable for a Java lint; "abusive" term explained). Commits are now SSH-signed and verified. Could you resolve the two conversations so the PR can be merged? Thank you!

VasilevNStas:

src/main/resources/org/eolang/motives/misc/redundant-attachment.md
```diff
@@ -62,16 +62,3 @@ written in the source, is left alone too. The `!` suffix on a nameless
 argument is such a case: `m.plus m!` makes the parser wrap `m` into a named
 `.as-bytes` over `Φ.dataized`, and no `>>` exists in the source to be
 removed.
-
-A void attribute declared as `? >> name` is left alone as well. The name
```
volodya-lombrozo, line 66: @VasilevNStas Why do u remove it in this PR? I guess, it should be solved in a separate issue

src/main/resources/org/eolang/lints/misc/redundant-attachment.xsl
```diff
@@ -21,13 +21,6 @@
   </xsl:function>
   <xsl:template match="/">
     <defects>
-      <!--
```
volodya-lombrozo, line 24: @VasilevNStas Why do u remove it in this PR? I guess, it should be solved in a separate issue

src/main/java/org/eolang/lints/LtAsciiOnly.java
```diff
@@ -42,8 +40,23 @@ public Collection<Defect> defects(final XML xmir) throws IOException {
             if (!abusive.isPresent()) {
                 continue;
             }
-            final int line = new LineOf(comment).value();
             final Character chr = abusive.get();
+            final String text = comment.text().get();
```
volodya-lombrozo, line 44: @VasilevNStas To be honest, I don't like how it looks. Can we create a separate clean classs `Listing` and place all of this logic into it?

volodya-lombrozo (changes requested):

src/main/java/org/eolang/lints/LtAsciiOnly.java
VasilevNStas: Done, the logic is now in a separate class `Listing` (`src/main/java/org/eolang/lints/Listing.java`)

It encapsulates everything about locating a comment inside the program source: reading the `<listing>` text, finding the comment line (which starts with `#`) and counting the newlines up to the character position. `LtAsciiOnly` now only consumes the result:

```java
line = listing.get().line(text, pos).orElse(new LineOf(comment).value());
```

`Listing` is package-private with its own unit test (`ListingTest`) covering the first line, a second line of a multi-line comment, and the not-found case. Commits are SSH-signed and pushed

VasilevNStas:
@volodya-lombrozo I`m fixed it
please re-review this

src/main/resources/org/eolang/lints/misc/redundant-attachment.xsl
volodya-lombrozo: @VasilevNStas The question still remains.

yegor256:
Please forgive us for closing this pull request. It has had merge conflicts with the `master` branch for over a month, so we're closing it as stale.

Please feel free to submit a new pull request, perhaps similar to this one, taking into account the changes that have since landed in `master`. Thank you for your contribution!

0crat:
@volodya-lombrozo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+441](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1244, opened 2026-08-11
fix(#1243): enforce prefixes per test kind in bad-test-name

### LINKED ISSUE #1243: `bad-test-name` lint fails to enforce specific prefixes for positive and negative test names
Consider this:

```
eq. ++> stops-on-zero
  recovered
    42.div 0
    "oops"
  "oops"
```

This is a valid test, technically speaking, but it's name is misleading. It starts with `stops-on`, while the test is positive. Should be `can-`.

Let's fix the `bad-test-name` lint. It should guarantee that names of positive tests start from `can-` and `accepts-` (only), while negative tests start from `cannot-`, `rejects-`, or `stops-on-`.

yegor256:
Closes #1243.

## Problem

`bad-test-name` accepted the same five prefixes (`can-`, `cannot-`, `accepts-`, `rejects-`, `stops-on-`) for every test, and it only inspected truthy tests (`@name` starting with `+`). So the example from the issue passed silently, even though the name claims the object fails while the test asserts that it works:

```eo
[] > foo
  eq. +> stops-on-zero
    recovered
      42.div 0
      "oops"
    "oops"
```

Throwing tests (`@name` starting with `-`) weren't checked at all.

## Fix

The allowed prefixes now depend on the kind of the test, as `Suffix.attribute` in `eo-parser` marks it:

| Kind | EO syntax | `@name` in XMIR | Allowed prefixes |
| --- | --- | --- | --- |
| positive | `+>` | `+the-name` | `can-`, `accepts-` |
| negative | `->` | `-the-name` | `cannot-`, `rejects-`, `stops-on-` |

`bad-test-name.xsl` iterates over `eo:test-name(@name)` instead of `starts-with(@name, '+')`, so throwing tests are linted too, and it picks the regexp and the list of prefixes reported in the defect message from the kind of the test. The message also names the kind now, e.g.:

```
The name of the positive test object "stops-on-zero" must start with one of the prefixes "can-,⌴accepts-"
```

## Tests

`allows-all-prefixes.yaml` was updated — the three negative prefixes moved to `->` tests, where they now belong. Four packs added:

- `catches-negative-prefix-in-positive-test.yaml` — the exact snippet from the issue
- `catches-all-negative-prefixes-in-positive-tests.yaml` — `+> cannot-…`, `+> rejects-…`, `+> stops-on-…`
- `catches-positive-prefix-in-negative-test.yaml` — `-> can-…`, `-> accepts-…`
- `catches-bad-name-in-negative-test.yaml` — `-> throws-on-empty`

The motive in `bad-test-name.md` documents both sets of prefixes.

`mvn test` passes (590 tests, 0 failures), as does `mvn test -Pdeep -Dtest=LtByXslTest#validatesEoPacksForErrors`, which confirms the new EO packs parse without errors. `xcop`, `vale` and `yamllint` are clean on the touched files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HaeFS4H5nK5HJtGyisEZKq)_

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
+++ b/src/main/resources/org/eolang/lints/tests/bad-test-name.xsl
@@ -3,17 +3,23 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="bad-test-name" version="2.0">
-  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" id="bad-test-name" version="2.0">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object//o[starts-with(@name, '+')]">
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (112 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2734](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1247, opened 2026-08-11
feat(#155): add too-deep-object lint

### LINKED ISSUE #155: new lint `too-deep-object` to catch very nested objects

VasilevNStas:
fix #155

## What

A new `design/too-deep-object` lint that warns when an object is nested
deeper than twelve levels.

## How it works

The lint counts the ancestors of every object in XMIR:

```
depth = count(ancestor::o)
```

If the depth exceeds 12, the object is reported as a `warning`.

Synthetic nodes introduced by the parser are excluded from the count:
`Φ.tuple` (tuples written as `*`), `Φ.bytes`, `Φ.number` and `Φ.string`
(literals). Without this, even the canonical example
(`src/test/resources/org/eolang/lints/canonical.eo`, which is only about
12 source levels deep) would be flagged, because the parser wraps every
literal and tuple in extra levels of nesting.

Only the deepest offending objects are reported (not every descendant),
so a single deep chain yields a single defect.

## Why

Deeply nested objects are hard to read and debug. The issue proposes a
threshold of 12, which this lint follows.

## Tests

- `catches-too-deep-object` — a 14-level chain is flagged
- `allows-normal-nesting` — a 7-level chain (from the issue) is clean

### DIFF
--- a/src/main/resources/org/eolang/lints/design/too-deep-object.xsl
+++ b/src/main/resources/org/eolang/lints/design/too-deep-object.xsl
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="too-deep-object">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@volodya-lombrozo plz review this pull request

yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (106 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+554](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1248, opened 2026-08-11
feat(#256): add roll-bases lint

### LINKED ISSUE #256: composite @base attributes must be mandatory, if possible
Consider this XMIR:

```
<o base=".foo">
  <o base=".bar">
    <o base="x"/>
  </o>
</o>
```

It may be written shorter:

```
<o base="x.bar.foo"/>
```

Let's make it mandatory to write it shorter, if possible.

VasilevNStas:
fix #256

## What

A new `misc/roll-bases` lint that reports composite `@base` attributes
which could be written in a shorter, rolled form.

Consider this XMIR:

```xml
<o base=".foo">
  <o base="x"/>
</o>
```

It can (and must) be written shorter:

```xml
<o base="x.foo"/>
```

The lint reports the unrolled form with a `warning`.

## Why

As @maxonfjvipon correctly pointed out, XMIR is not only produced by
`eo-parser` (which already rolls bases during parsing) — it may also come
from `phino`, `jeo-maven-plugin`, or be written by hand. In those inputs
unrolled composite bases are a real mistake, and nothing currently flags
them.

## How it works

The lint follows the collapse rules of `roll-bases.xsl` from `eo-parser`:
a node with `@base` starting with `.` and a single child whose `@base`
does not start with `.`, has no inner objects, no data and no name, must
be merged into `concat(child/@base, @base)`.

## Why unit tests instead of yaml packs

The yaml packs always parse EO code, and `eo-parser` rolls bases on that
path — so an unrolled `catches` case cannot be expressed in a pack. The
lint is therefore covered by unit tests in `LtByXslTest` that feed a
hand-built XMIR (via `Xembler`) straight to `LtByXsl`:

- `catchesUnrolledBases` — `.foo`/`x` is reported
- `allowsRolledBase` — `x.foo` is clean
- `allowsNestedBaseWithData` — a child with data must not be rolled
- `allowsDoubleDotBases` — `.foo`/`.bar` must not be rolled

Both `mvn test` (594 tests) and `mvn clean install -Pqulice` pass.


VasilevNStas:
@volodya-lombrozo @maxonfjvipon please review this PR

VasilevNStas:
@volodya-lombrozo 
I`m fixed conflict, plz re-review

src/test/java/org/eolang/lints/LtByXslTest.java
```diff
@@ -404,6 +404,76 @@ void validatesEoPacksForErrors() throws IOException {
         );
     }
 
+    @Test
+    void catchesUnrolledBases() throws ImpossibleModificationException {
```
volodya-lombrozo, line 408: @VasilevNStas Can u use test packs, please? If it's impossible, it's better to wait for https://github.com/objectionary/lints/issues/1271

volodya-lombrozo (changes requested):

src/test/java/org/eolang/lints/LtByXslTest.java
VasilevNStas: @volodya-lombrozo Unfortunately, test packs can't cover these scenarios — here is the concrete reason.

The roll-bases lint only fires on *unrolled* bases:

```xml
<o base=".foo">
  <o base="x"/>
</o>
```

But the EO parser collapses exactly this shape while parsing: `org/eolang/parser/parse/roll-bases.xsl` (a pipeline step of `eo-parser`) turns it into `base="x.foo"`, and a bare `.foo` continuation without an expression is a parse error. So an `input:` pack (an EO program) can never yield the XMIR the lint is about — I verified that `EoProgram.parse()` only emits rolled bases (`Φ.x.foo` and similar). The `allowsNestedBaseWithData` and `allowsDoubleDotBases` cases equally depend on the unrolled shape, so they can't be expressed in EO either.

That leaves hand-written XMIR (via `Directives`/`Xembler`) as the only way to exercise the lint, which is what the four JUnit tests do. I'll be happy to move them to packs as soon as packs can accept raw XMIR (or once #1271-style `lint:` references make this possible).

VasilevNStas:
@volodya-lombrozo 
plz re-review this

src/test/java/org/eolang/lints/LtByXslTest.java
VasilevNStas: @volodya-lombrozo My previous reply was wrong — thanks for pushing on this. The packs do support raw XMIR via the `document:` key (now used by line-is-absent, identity-object and others), which is exactly what the roll-bases scenarios need: the EO parser rolls bases, so `input:` can never express the unrolled form.

I converted all four JUnit tests to packs in `packs/single/roll-bases/`:
- `catches-unrolled-base.yaml` — `document:` with the unrolled `<o base=".foo"><o base="Φ.x"/></o>`, `defects: 1`;
- `allows-rolled-base.yaml`, `allows-nested-base-with-data.yaml`, `allows-double-dot-bases.yaml` — `defects: 0`.

The conversion also surfaced a real bug in the XSL: `eo:escape()` already wraps its argument in quotes, so the defect text was emitted with doubled quotes (`""x.foo""`), which the `GrammarMatcher` spell-check rejected. Removed the redundant quotes in `roll-bases.xsl`.

Verified: `mvn clean install -Pqulice` (1019 rules) and `mvn clean test` (664 tests) pass.

PR #1249, opened 2026-08-11
feat(#407): add test-without-assert lint

### LINKED ISSUE #407: unit tests without asserts should be prohibited
Consider this EO test:

```eo
+tests

# Test
[] > checks-something
  52 > @
```

This test has no assertions, so it does not test anything. Let's complain about such situations if test object don't have any assertions inside.

Correct test should be:

```eo
+test

# Test
[] > checks-something
  eq. > @
    42
    42
```

VasilevNStas:
fix #407

## What

A new `tests/test-without-assert` lint that reports unit tests whose `@`
attribute is just a literal — i.e. tests that assert nothing.

## Why

A test like this:

```eo
[] +> checks-something
  52 > @
```

passes no matter what the code does, so it can never catch a regression.
It tests nothing.

## How it works

The lint walks test objects (`eo:test-name`, i.e. `+`/`-` prefixed names)
and checks their `φ` (`@`) attribute. If `φ` is one of the literal bases
(`Φ.number`, `Φ.string`, `Φ.bytes`, `Φ.bool`), there is no assertion and
the test is reported as a `warning`.

The heuristic answers @yegor256's question about the formal definition of
an "assertion": the simplest provable case is a bare literal — a test that
just returns a value with no comparison, no application, no side effect.
Tests that call anything (an `eq`, a `stdout`, any application) have a
non-literal `φ` and are left alone. This keeps the check conservative:
no false positives, at the cost of only catching the clearest offenders.

## Tests

- `catches-test-without-assert` — `52 > @` is flagged
- `allows-test-with-assert` — `eq. > @` is clean
- `allows-side-effect-test` — `stdout > @` is clean (conservative)

Both `mvn test` (593 tests) and `mvn clean install -Pqulice` pass.

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/test-without-assert.xsl
+++ b/src/main/resources/org/eolang/lints/tests/test-without-assert.xsl
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="test-without-assert">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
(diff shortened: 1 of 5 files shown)


VasilevNStas:
@yegor256 @volodya-lombrozo please review

yegor256:
@VasilevNStas the #407 was never approved. Please, try NOT to fix issues that are not registered as bugs.

PR #1252, opened 2026-08-11
fix(#586): accumulate and validate without() names

### LINKED ISSUE #586: Strange behavior of `without` method
I’m not happy with how the `without()` method works:

1. If I call `without()` twice, only the last call has any effect - previous ones are ignored, which can be misleading
```java
new Source(
    new EoSyntax(
        """
        +architect yegor256@gmail.com
        +home https://www.eolang.org
        +package canonical
        +version 0.0.0
        +spdx SPDX-FileCopyrightText Copyright (c) 2016-2025 Objectionary.com
        
        # Times table, which is a canonical example of EO program.
        [start] > canonical
        """
    ).parsed()
).without("unused-void-attr").without("bipki").defects().forEach(System.out::println);
```
Output
```console
[canonical.canonical unused-void-attr (Single) WARNING]:9 The void attribute "start" is unused anywhere in the object "canonical"
```

2. The method takes strings, so if I pass an incorrect lint name, nothing happens and there's no feedback. I see two ways to address this:

    2.1. If we don't intend to support custom lints(as it currently stands - `Program` and `Source` don't accept `Lint` as an argument, even though `Lint` is a public interface), then we should make `Lint` `package-private` and generate enum of all existing lint names during build time(or write it manually). Then, `without()` would take this enum type, allowing compile-time checks to ensure the disables are valid

    2.2. If we do want to allow custom lints in the future, we should add a public constructor for `Program` and `Source` that accepts a list of custom linters. Then, `without()` should throw an exception if it tries to disable a non-existent lint

3. `without()` should accept one required argument followed by varargs, so that calling `.without()` with no arguments isn't possible.

VasilevNStas:
fix #586

## What

Fixes the strange behavior of `Source.without()`:

1. **Accumulation** — chained calls now combine:
   `without("a").without("b")` disables both `a` and `b`, instead of the
   last call silently overriding the previous one.
2. **Validation** — passing a lint name that does not exist now throws
   `IllegalArgumentException` instead of silently doing nothing.
3. **Signature** — `without(String first, String... rest)` requires at
   least one argument, so `without()` with no arguments is impossible.

## How

`Source` now keeps a `disabled` collection of lint names. Each `without()`
call appends to it, so the accumulated set is passed to `MonoWithout`.
Before constructing the new `Source`, every requested name is checked
against the full set of known lint names (from `PkMono`); an unknown name
throws.

Note: the existing test `createsSourceWithoutMultipleLints` referenced the
lint `object-does-not-match-filename`, which no longer exists in the
codebase — the validation rightfully rejects it, so the test now uses a
real lint name (`sparse-seq`) instead.

## Tests

- `accumulatesWithoutCalls` — two chained `without()` calls both take effect
- `rejectsUnknownLintName` — an unknown name throws `IllegalArgumentException`
- existing `without`-related tests updated for the new signature

### DIFF
--- a/src/main/java/org/eolang/lints/Source.java
+++ b/src/main/java/org/eolang/lints/Source.java
@@ -10,7 +10,10 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
 import org.cactoos.iterable.Sticky;
 import org.cactoos.iterable.Synced;
 
@@ -41,6 +44,11 @@ public final class Source {
     private final Iterable<Lint> lints;
 
     /**
+     * Disabled lint names.
+     */
+    private final Collection<String> disabled;
+
+    /**
      * Ctor.
      * @param file The absolute path of the XMIR file
      * @throws FileNotFoundException If file isn't found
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@volodya-lombrozo @yegor256 please review this

volodya-lombrozo:
@VasilevNStas before implementing an issue it's better to wait until it it's accepted by maintainers. We aren't planning to implement this one. That said, thank you for the attempt!

PR #752, opened 2025-09-13
Added lint for fix of comments in test code duplicate test names

kislayykumar:
Fix for https://github.com/objectionary/lints/issues/748#issuecomment-3287589378

## Summary by CodeRabbit

- **New Features**
  - Added a lint that warns when test objects contain comments and recommends concise, self‑explanatory test names.

- **Tests**
  - Added unit tests validating the lint’s behavior, name, and guidance.
  - Increased a test timeout to 6 minutes to accommodate longer validation runs.

PR #1255, opened 2026-08-12
feat(#748): add test-with-comment lint

### LINKED ISSUE #748: comments in test code duplicate test names
I suggest prohibiting this:

```
# This test checks how my object works:
[] +> the-object-works
```

Such a comment usually is a duplication of the test name. It's better to create short and easy to understand tests, instead of writing documentation for them.

VasilevNStas:
fix #748

## What

A new `tests/test-with-comment` lint that warns when a test object has a
comment before it. A comment that duplicates the test name adds nothing —
the name should be self-explanatory.

## How it works

The lint looks at top-level test objects (`/object/o[eo:test-name(@name)]`,
i.e. `+`/`-` prefixed names). The parser puts the comment that precedes a
top-level object into `/object/comments/comment`. If a test object has any
such comment, a `warning` is reported.

Note: this only applies to top-level tests. A comment placed inside an
object right before a nested `+>` test is already a parse error
("comment is allowed only on top of the file, before metas"), so there is
no nested case to handle.

## Why not the earlier attempt (PR #752)

The earlier PR #752 looked for `meta[@key='comment']`, but comments are not
stored as metas in XMIR — they live in `/object/comments/comment`. This
implementation uses the correct location.

## Tests

- `catches-test-with-comment` — a comment before a top-level test is flagged
- `allows-test-without-comment` — no comment is clean

### DIFF
--- a/src/main/resources/org/eolang/lints/tests/test-with-comment.xsl
+++ b/src/main/resources/org/eolang/lints/tests/test-with-comment.xsl
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="test-with-comment">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@volodya-lombrozo @yegor256 please review this pull request

yegor256:
@VasilevNStas comments are disabled in XMIR now, except the top-level single comment

VasilevNStas:
@yegor256 
Thanks — that is exactly the case this lint targets. Since nested comments are rejected by the parser, the only place a comment can appear is the top-level comment block (`/object/comments/comment`), right before the single top-level object. The lint flags precisely that: when the top-level object is a test (`eo:test-name`, i.e. `+`/`-` prefixed) and a top-level comment exists, it reports a warning. So it is aligned with the current XMIR behavior, not relying on any nested-comment support.

yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+570](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1257, opened 2026-08-12
feat(#796): add long-live-object lint

### LINKED ISSUE #796: `long-live-object` lint is not implemented
How about we create a lint that would catch objects, located too far from their usages. In this program, `f` can be moved closer to its usage:

```eo
[] > app
  foo 42 > f
  … # some statements here
  52.plus f > r
```

The lint should be able to accept `maxObjectLifetime` parameter. If its bigger than predefined threshold, the defect with warning severity should be generated.

VasilevNStas:
fix #796

## What

A new `design/long-live-object` lint that warns when an object is defined
too far from its first usage.

## Why the naive approach fails

The issue's original idea was a simple distance in source lines. As
@yegor256 pointed out, that breaks on this case:

```eo
[] > app
  [] > a
    # 100 lines
  [] > b
    # 100 lines
  a.plus b > r
```

Here `a` is used many lines after its definition, yet moving it closer
does not help — `b` is equally far in the other direction. A line-based
check would produce a useless warning.

## The approach

Instead of counting source lines, the lint counts **sibling attributes of
the same parent** between the definition and the first usage:

- `f` in the issue's example is `6` attributes away from `52.plus f` →
  reported.
- In the counter-example above, `a` is only `1` attribute away from
  `a.plus b` (only `b` sits between) → not reported, because the bodies
  of `a` and `b` are nested and do not count.

So the "lifetime" of an object is measured in structural neighbors, not
physical lines — which resolves @yegor256's concern.

The threshold is hard-coded to `5` attributes.

## How it works

For every named object, the lint finds the first later reference
(`//o[@base='ξ.NAME']`), then counts the parent's children whose
`@line` lies between the definition and that usage. If more than 5, a
`warning` is emitted at the definition line.

Note: `@line` comparisons are numeric (`number(@line)`); string
comparison silently breaks on two-digit line numbers.

## Tests

- `catches-far-usage` — `f` used 6 attributes later is flagged
- `allows-close-usage` — the @yegor256 counter-example is clean

### DIFF
--- a/src/main/resources/org/eolang/lints/design/long-live-object.xsl
+++ b/src/main/resources/org/eolang/lints/design/long-live-object.xsl
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" version="2.0" id="long-live-object">
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@yegor256  @volodya-lombrozo 
please review this

volodya-lombrozo:
@VasilevNStas The issue isn't accepted yet. Please, wait a bit when maintainers will come up with the solution or agree with the necessity to add this lint. Thank you for the attempt!

Issue #1101, opened 2026-07-26
`redundant-attachment` lint fails to target unnecessary private attachments in specific instances


yegor256:
Look at this code:

```
[] > foo
  test > @
    [] >>
      2.plus 2 > @
```

Here, the `>>` at the third line is redundant, because the body of the object doesn't refer to any objects outside it. It may be safely refactored to this:

```
[] > foo
  test > @
    []
      2.plus 2 > @
```

Let's create a lint to catch such situations.

0pdd:
@yegor256 the puzzle [#1236](https://github.com/objectionary/lints/issues/1236) is still not solved.

0pdd:
@yegor256 the only puzzle [#1236](https://github.com/objectionary/lints/issues/1236) is solved here.

volodya-lombrozo:
This one was solved in the #1229

0crat:
@volodya-lombrozo Thanks for closing this issue! You've earned +4 points for this: +8 as a basis; -7 for 15 days of delay; +3 to give you at least something. Your running score is [+517](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1262, opened 2026-08-13
fix(#1261): stop redundant-attachment from contradicting anonymous-formation

### LINKED ISSUE #1261: `redundant-attachment` and `anonymous-formation` rules clash causing unavoidable warnings
The `redundant-attachment` and `anonymous-formation` rules give opposite
instructions for the same object, so a formation that reads its enclosing
scope cannot satisfy both. `misc/redundant-attachment.xsl:16` flags every
`a🌵N-M` auto-name that has no referrer, asking for the `>>` to be deleted,
while `names/anonymous-formation.xsl:30` flags every unnamed formation that
reaches out to `ξ.ρ.*`, asking for the name back. `tuple/mapped.eo` in
objectionary/eo is a plain example, since `[item idx]` reads both `func` and
`item` from outside itself:

```eo
func item > [item idx] >>   # redundant-attachment: "doesn't need to be named"
func item > [item idx]      # anonymous-formation: "accessing undefined object func"
```

Whichever form is written, one of the two warnings stands, so the project
cannot reach zero warnings. In eo-runtime the rule flags 120 such handles,
and dropping them raises 196 complaints of the other kind, which is why
`redundant-attachment` had to be switched off there through
`skipSourceLints`. The name is not in fact redundant: `redundant-attachment.md`
says a generated name matters when something refers to it, and #1101 framed
the removable case as the one where "the body doesn't refer to any objects
outside it" — a qualifier the XSL never checks.

The smallest fix is to add that qualifier as a guard on the condition at
`redundant-attachment.xsl:16`, skipping the object when its body holds a
`ξ.ρ.`-based reference that none of its own `∅` attributes cover, which is
the node set `anonymous-formation.xsl:30` already selects. Note also that the
same condition fires where the source has no `>>` at all: `m.plus m!` makes
the parser synthesise `<o name="a🌵9-9" base=".as-bytes">` over `Φ.dataized`,
and that warning cannot be acted upon from the source.

yegor256:
Closes #1261.

## The deadlock

`misc/redundant-attachment.xsl:16` flagged every `a🌵N-M` auto-name with no
referrer, asking for the `>>` to go. `names/anonymous-formation.xsl:30` flagged
every unnamed formation reaching into `ξ.ρ.*`, asking for the name back. A
formation that reads its enclosing scope satisfied neither form:

```eo
func item > [item idx] >>   # redundant-attachment: "doesn't need to be named"
func item > [item idx]      # anonymous-formation: "accessing undefined object func"
```

## The fix

The set of objects a formation reads from outside itself — every `ξ.ρ.`-based
reference in its body that neither its own `∅` voids nor its own bound
attributes cover — is now a shared function, `eo:outer-refs`, in the new
`funcs/outer-refs.xsl`. Both lints judge a formation by that same set, from
opposite sides, so they can no longer drift apart:

- `anonymous-formation` iterates it (behaviour unchanged — the inline
  `eo:matches-any-void`/`eo:base-to-name` pair and the nested loop it drove
  collapse into one `for-each`).
- `redundant-attachment` guards on it being empty, which is the qualifier
  `redundant-attachment.md` and #1101 always implied but the XSL never checked:
  a name is removable only when "the body doesn't refer to any objects outside
  it".

The issue's second note is fixed too. The rule fired where the source has no
`>>` at all: `m.plus m!` makes the parser synthesise
`<o name="a🌵N-M" base=".as-bytes">` over `Φ.dataized` for the nameless const
argument, and no suffix existed to delete. `eo:const-wrapper` now recognises
that wrapper and leaves it alone.

## Consequences

`canonical.eo` no longer needs `+unlint redundant-attachment`, which resolves
the `@todo #1101:30min` on `SourceTest.acceptsCanonicalCode` in favour of the
documented exception: the `[i]`/`[i1]` closures over `^.x` keep their `>>`,
and the self-contained `[x]` drops the one it never needed. Verified that
`[x]`'s subtree reaches no further out than `ξ.ρ.x` (its own void), so
inlining it changes nothing it references.

## Tests

Three new packs under `packs/single/redundant-attachment/`, covering the
`mapped.eo` shape, a nested formation reading an outer `t`, and the `m!` const
wrapper. Existing packs for both lints keep their expectations unchanged.

Full suite green (594 tests, 0 failures), plus `qulice`, `xcop`, `reuse` and
`yamllint` locally.

https://claude.ai/code/session_01Hw5qR8Jk8bJN2EsjPdMWSK

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hw5qR8Jk8bJN2EsjPdMWSK)_

### DIFF
--- a/src/main/resources/org/eolang/funcs/outer-refs.xsl
+++ b/src/main/resources/org/eolang/funcs/outer-refs.xsl
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" id="outer-refs" version="2.0">
+  <!--
+  The name a "ξ.ρ."-based reference resolves to, with the "ξ" and the
+  chain of "ρ" hops that lead out of the formation stripped away.
+  -->
+  <xsl:function name="eo:base-to-name" as="xs:string">
+    <xsl:param name="base" as="xs:string"/>
+    <xsl:value-of select="replace($base, '^ξ(\.ρ)+\.', '')"/>
+  </xsl:function>
+  <!--
+  The objects a formation reads from the scope that encloses it: every
+  "ξ.ρ."-based reference in its body that neither its own voids nor its own
+  bound attributes cover. When this is empty, the formation stands on its
+  own and makes sense without a name and without an enclosing scope. Both
+  "anonymous-formation" and "redundant-attachment" judge a formation by
+  this very set, from opposite sides, so they share it here.
+  -->
+  <xsl:function name="eo:outer-refs" as="element()*">
+    <xsl:param name="formation" as="element()"/>
+    <xsl:variable name="allowed" as="xs:string*">
+      <xsl:for-each select="$formation/o[@base='∅']">
+        <xsl:sequence select="concat('ξ(\.ρ)+\.', @name, '(?:\.\w+)?')"/>
+      </xsl:for-each>
+    </xsl:variable>
+    <xsl:for-each select="$formation//o[starts-with(@base, 'ξ.ρ.') and not(some $pattern in $allowed satisfies matches(@base, $pattern))]">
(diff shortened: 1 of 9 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (183 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2924](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1267, opened 2026-08-13
fix(#1266): exempt void auto-names from `redundant-attachment`

### LINKED ISSUE #1266: `redundant-attachment` warning incorrectly triggered by void alias which cannot be removed
The selector at `src/main/resources/org/eolang/lints/misc/redundant-attachment.xsl:24` takes every object with an `a🌵`-auto-name except the parser's const wrapper, so it also takes void attributes. In 0.4.3 this source draws `WARNING redundant-attachment/S:11` on the `? >> right` line:

```
# Picks one of the two.

+architect yegor256@gmail.com
+home https://github.com/objectionary/eo
+package foo
+version 0.0.0

[] > choice
  ? >> left
  ? >> right
  left > @
```

The warning cannot be acted upon, because the grammar has no nameless void. Dropping the alias makes eo-parser refuse the file: `a void attribute must be written as ? > name or ? >> name`. The only other spelling, `? > right`, publishes the attribute under that name instead of `a🌵11-2`, and that changes dispatch. In objectionary/eo exactly this rewrite of `true.eo` and `false.eo` — whose inner formation is `bool`'s `if` — made `i64.eo`'s `if. > magnitude!` resolve `magnitude.left` and `magnitude.right` to the voids instead of letting them fall through the chosen branch to `bytes.left` and `bytes.right`, which is what `bool.eo` pins with `can-answer-the-left-method-of-a-chosen-branch`. About ninety eo-runtime tests failed with `Φ.bytes.right ... the 'x' attribute must be a number`, so eo-runtime now carries `+unlint redundant-attachment:16` in `true.eo` and `+unlint redundant-attachment:15` in `false.eo`.

The fix is to add `@base='∅'` to the exclusions in that selector, next to `not(eo:const-wrapper(.))`. It is the same exemption the rule already grants the `.as-bytes` const wrapper at lines 12-21, for the same reason given in that comment: the auto-name comes from the parser and there is no suffix in the source to delete.

These two voids are two of the three complaints left in eo-runtime after #1261; the third one was a genuine leftover and is fixed on our side.

yegor256:
Closes #1266.

## The problem

The selector in `redundant-attachment.xsl` took every object with an `a🌵`-auto-name except the parser's `.as-bytes` const wrapper, so it also took void attributes declared as `? >> name`:

```eo
[] > choice
  ? >> left
  ? >> right
  left > @
```

That warning could not be acted upon. The grammar has no nameless void, so the `>>` cannot be deleted — eo-parser refuses the file with `a void attribute must be written as ? > name or ? >> name`. The only other spelling, `? > right`, publishes the attribute under `right` instead of the generated `a🌵11-2`, which changes dispatch. In objectionary/eo exactly that rewrite of `true.eo` and `false.eo` made `i64.eo`'s `if. > magnitude!` resolve `magnitude.left`/`magnitude.right` to the voids instead of letting them fall through the chosen branch to `bytes.left`/`bytes.right`, breaking ~90 eo-runtime tests.

## The fix

`src/main/resources/org/eolang/lints/misc/redundant-attachment.xsl` — add `not(@base='∅')` to the selector's exclusions, next to `not(eo:const-wrapper(.))`. It is the same exemption the rule already grants the const wrapper, for the same reason: the auto-name comes from the parser, not from a `>>` written in the source, so there is no suffix to remove.

## Tests

- New pack `src/test/resources/org/eolang/lints/packs/single/redundant-attachment/allows-void-auto-name.yaml` reproduces the issue's `choice` source and expects `defects: 0`.
- Verified the pack is not vacuous: with the selector reverted it fails (`Pack 'redundant-attachment/allows-void-auto-name.yaml' doesn't tell the story as expected`), and with the fix in place `LtByXslTest` passes — 416 tests, 0 failures, 2 skipped.
- The motive doc gained a section on why `? >> name` is left alone.

Once released, eo-runtime can drop `+unlint redundant-attachment:16` from `true.eo` and `+unlint redundant-attachment:15` from `false.eo`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EkEZZTN7jSGSS9QNqF4ihg)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2784](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1275, opened 2026-08-14
`duplicate-as-attribute`: group `@as` siblings instead of a global key

### LINKED ISSUE #1274: `duplicate-as-attribute` XSL transformation is too slow on large XMIR files
While compiling `eo-runtime` (in [objectionary/eo](https://github.com/objectionary/eo)), this shows up:

```text
XSL transformation 'duplicate-as-attribute' took 1s, whereas threshold is 100ms
```

The same warning is reproducible right here, in `LtByXslTest.checksDuplicateAsAttributeLintOnLargeXmirInReasonableTime`.

The stylesheet is `org/eolang/lints/names/duplicate-as-attribute.xsl`. It finds duplicates through a key:

```xml
<xsl:key name="as-key" match="o[@as]" use="concat(generate-id(..), '|', @as)"/>
...
<xsl:for-each select="//o[o[@as]]">
  <xsl:variable name="pid" select="generate-id()"/>
  ...
  <xsl:for-each select="o[@as][count(key('as-key', concat($pid, '|', @as))) &gt; 1 and generate-id(.) = generate-id(key('as-key', concat($pid, '|', @as))[1])]">
```

Two things make this expensive:

  * The key is global, so it is built over **every** `o[@as]` in the document,
    calling `generate-id()` on each parent and building a concatenated string
    for each one. All of that work happens even for documents that have no
    duplicates at all.

  * The inner predicate does **two** `key()` lookups and **two** `generate-id()`
    calls per `@as` child, on top of a `concat()`. Since the duplicates we look
    for are always siblings, none of this global indexing is needed.

`xsl:for-each-group` does the same grouping locally, in one pass over the
children, and needs no ids and no keys. Adding a `[2]` positional predicate to
the outer selector also lets us skip, before any grouping happens, every object
that has fewer than two `@as` children — which is the majority of them in a
real XMIR:

```xml
<xsl:for-each select="//o[o[@as][2]]">
  <xsl:variable name="parent" select="."/>
  ...
  <xsl:for-each-group select="o[@as]" group-by="string(@as)">
    <xsl:if test="count(current-group()) &gt; 1">
```

Measured on this machine, through `XSLDocument` + `ClasspathSources`, exactly
as `LtByXsl` applies it:

| Input | Now | With grouping |
| --- | --- | --- |
| The 500×500 XMIR from `LtByXslTest` | 1297 ms | 280 ms |
| 171 XMIR files of `eo-runtime`, parsed by `eo-parser` | 398 ms | 186 ms |

The output is unchanged. I ran both stylesheets over 5000 randomly shaped XMIRs
(varying nesting, presence of `@name`/`@line`, and a tiny alphabet of `@as`
values that included the empty string and values containing `|`, to try to
break the concatenated key) and the two produced byte-identical `<defects>`
documents in every case.

I'm sending a pull request.

yegor256:
Closes #1274.

`duplicate-as-attribute.xsl` built a document-wide `xsl:key` over every
`o[@as]`, keyed by `generate-id()` of the parent, and then spent two `key()`
lookups plus two `generate-id()` calls per `@as` child to decide whether that
child was the first of a duplicate group. All of that indexing was paid even by
documents with no duplicates at all.

Duplicated `@as` attributes are always siblings, so no global index is needed.
`xsl:for-each-group` finds them locally, in a single pass over the children,
with no ids and no keys:

```xml
<xsl:for-each select="//o[o[@as][2]]">
  <xsl:variable name="parent" select="."/>
  ...
  <xsl:for-each-group select="o[@as]" group-by="string(@as)">
    <xsl:if test="count(current-group()) &gt; 1">
```

The `[2]` predicate on the outer selector is the second half of it: it skips,
before any grouping happens, every object that has fewer than two `@as`
children — the majority of them in a real XMIR.

Since `eo:defect-context()` is now called from inside the `xsl:if`, it can only
ever run for an object that really has a duplicate, instead of being bound as a
variable for every candidate parent.

## Measurements

Through `XSLDocument` + `ClasspathSources`, exactly as `LtByXsl` applies it,
best of several runs:

| Input | Before | After |
| --- | --- | --- |
| The 500×500 XMIR from `LtByXslTest` | 1297 ms | 280 ms |
| 171 XMIR files of `eo-runtime`, parsed by `eo-parser` | 398 ms | 186 ms |

## Correctness

The reported defects are unchanged. I ran the old and the new stylesheet over
8000 randomly shaped XMIRs — varying nesting depth, presence of `@name` and
`@line`, and a deliberately tiny alphabet of `@as` values that included the
empty string and values containing `|` (to try to break the concatenated key) —
and they produced byte-identical `<defects>` documents in every case, `@context`
attributes included.

`LtByXslTest` passes: 417 tests, 0 failures.

I also added `reports-each-duplicate-once.yaml`, which pins the behaviour the
grouping has to preserve: an object with three `foo` children and two `bar`
children yields exactly two defects, not five.

---
_Generated by [Claude Code](https://claude.ai/code/session_013RN4eq44VcauNbKJ2LVnX9)_

### DIFF
--- a/src/main/resources/org/eolang/lints/names/duplicate-as-attribute.xsl
+++ b/src/main/resources/org/eolang/lints/names/duplicate-as-attribute.xsl
@@ -8,39 +8,44 @@
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:key name="as-key" match="o[@as]" use="concat(generate-id(..), '|', @as)"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[o[@as]]">
-        <xsl:variable name="pid" select="generate-id()"/>
+      <!--
+      The "[2]" predicate lets us skip, without any grouping at all, every
+      object that has less than two "as" children, which is the majority
+      of them.
+      -->
+      <xsl:for-each select="//o[o[@as][2]]">
+        <xsl:variable name="parent" select="."/>
         <xsl:variable name="oname" select="@name"/>
         <xsl:variable name="line" select="eo:lineno(@line)"/>
-        <xsl:variable name="context" select="eo:defect-context(.)"/>
-        <xsl:for-each select="o[@as][count(key('as-key', concat($pid, '|', @as))) &gt; 1 and generate-id(.) = generate-id(key('as-key', concat($pid, '|', @as))[1])]">
-          <defect>
-            <xsl:attribute name="line">
-              <xsl:value-of select="$line"/>
-            </xsl:attribute>
-            <xsl:if test="$line = '0'">
-              <xsl:attribute name="context">
-                <xsl:value-of select="$context"/>
+        <xsl:for-each-group select="o[@as]" group-by="string(@as)">
+          <xsl:if test="count(current-group()) &gt; 1">
+            <defect>
+              <xsl:attribute name="line">
+                <xsl:value-of select="$line"/>
               </xsl:attribute>
-            </xsl:if>
-            <xsl:attribute name="severity">warning</xsl:attribute>
-            <xsl:text>The </xsl:text>
-            <xsl:choose>
-              <xsl:when test="$oname">
-                <xsl:text>object </xsl:text>
-                <xsl:value-of select="eo:escape($oname)"/>
-              </xsl:when>
-              <xsl:otherwise>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2564](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1233, opened 2026-08-10
fix(#364): run only referenced lints in unlint-non-existing-defect

VasilevNStas:
Fixes #364

**Problem:** The `unlint-non-existing-defect` lint is the slowest one, by an order of magnitude, according to the [benchmark](https://github.com/objectionary/lints?tab=readme-ov-file#benchmark). The reason: for every document it ran `defects()` on **all** lints, even when the document had no `+unlint` metas at all.

**Solution:** The lint now collects the `+unlint` tails first and, if the document has none, returns an empty result without touching any lint. Otherwise, only the lints whose names are referenced by the `+unlint` metas are executed, instead of the full set.

**Changes:**
- `src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java`: rewritten `defects()` to short-circuit on documents without `+unlint`; `existing()` filters `lints` by the referenced names before calling `defects()`.
- `src/test/java/org/eolang/lints/LtUnlintNonExistingDefectTest.java`: two new tests asserting that no lint runs when there are no `+unlint` metas and that unreferenced lints are not invoked (a lint that explodes when called would fail the test).

**Tests:** `mvn clean install -Pqulice` — 580 tests, all green.

PR #1238, opened 2026-08-10
feat(#1100): add broad-scope lint

VasilevNStas:
Fixes #1100.

## Problem
A private attribute (`local`) may have an excessively broad scope: it is used only inside a single nested object, while it is declared at a much higher level. This should lead to a lint warning.

## Solution / What
New XSLT lint `broad-scope` modeled after `redundant-object`: it finds private attributes referenced (via `^ξ(?:\.ρ)*\.` base prefixes) exclusively within a single object different from the declaration site, and warns that the scope is too broad.

## Changes
- `src/main/resources/org/eolang/lints/design/broad-scope.xsl` — the new lint.
- `src/main/resources/org/eolang/motives/design/broad-scope.md` — motivation.
- `src/test/resources/org/eolang/lints/packs/single/broad-scope/` — 3 test packs.

## Tests
- `catches-broad-scope`
- `allows-used-on-top-level`
- `allows-used-in-multiple-objects`

PR #1285, opened 2026-08-19
feat(#1284): benchmark unlint-non-existing-defect with +unlint metas

### LINKED ISSUE #1284: `unlint-non-existing-defect` benchmark lacks coverage for documents with `+unlint` metas, potentially missing regressions
I cannot tell whether objectionary/lints#1233 (or any future change to `LtUnlintNonExistingDefect`) is safe, because the only benchmark we have for that lint — the `unlint-non-existing-defect` entries in the README benchmark table — is driven by real-world bytecode-derived XMIR (`com.sun.jna.*`, `org.apache.hadoop...FSNamesystem`), none of which contain a single `+unlint` meta. That benchmark only measures the "no `+unlint` at all" fast path.

objectionary/lints#1233 changes the *other* path too: when `+unlint` metas are present, it now filters `this.lints` down to only the lints referenced by name before computing `existingDefects()`, instead of running the full list. We have no fixture with `+unlint` metas referencing defects that actually exist, so we have zero timing signal for that code path — no way to tell if it got faster, stayed the same, or regressed.

VasilevNStas:
fix #1284

## What

Adds a JMH benchmark that measures `LtUnlintNonExistingDefect` on an XMIR that actually contains `+unlint` metas.

## Why

The existing `scansXmir` benchmark is driven by real-world bytecode-derived XMIR (`com.sun.jna.*`, `org.apache.hadoop...FSNamesystem`), none of which contains a single `+unlint` meta. So it only measures the "no `+unlint` at all" fast path of `LtUnlintNonExistingDefect`.

#1233 changed the other path too: when `+unlint` metas are present, the lint now filters `this.lints` down to the referenced names before computing `existing()`. Until now there was no fixture exercising that path, so a regression there would go unnoticed by the benchmark.

## How

New benchmark method `scansXmirWithUnlints` in `SourceBench`, which parses the existing fixture `unlint-ascii-only.eo` (it has `+unlint ascii-only` and a non-ASCII comment that produces a real defect) and runs the full lint set on it.

Measured result (JMH, average time):

```
SourceBench.scansXmirWithUnlints  N/A  avgt  3  9.097 ± 31.283  ms/op
```

This gives a timing signal for the `+unlint` code path, so future changes to `LtUnlintNonExistingDefect` (faster, same or slower) become visible.

Both `mvn test` (606 tests) and `mvn clean install -Pqulice` pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,42 +124,44 @@ Here is the result of linting XMIRs:
 <!-- benchmark_begin -->
 ```text
 Input: com/sun/jna/PointerType.class (S source)
-Lint time: 10s (10054 ms)
+Lint time: 6s (6315 ms)
 
 Input: com/sun/jna/Memory.class (M source)
-Lint time: 7s (7167 ms)
+Lint time: 2s (2251 ms)
 
 Input: com/sun/jna/Pointer.class (L source)
-Lint time: 9s (8715 ms)
+Lint time: 4s (3776 ms)
 
 Input: com/sun/jna/Structure.class (XL source)
-Lint time: 11s (10952 ms)
+Lint time: 5s (5443 ms)
 
 Input: org/apache/hadoop/hdfs/server/namenode/FSNamesystem.class (XXL source)
-Lint time: 38s (37742 ms)
+Lint time: 26s (25826 ms)
+
+Input: org/eolang/lints/unlint-ascii-only.eo (UNLINT source)
+Lint time: 296ms (296 ms)
 
 
-
-unlint-non-existing-defect (XXL) (14513 ms)
-application-without-as-attributes (XXL) (3157 ms)
-unlint-non-existing-defect (XL) (2790 ms)
-unlint-non-existing-defect (L) (1559 ms)
-object-has-data (XXL) (1413 ms)
-unlint-non-existing-defect (M) (1256 ms)
-empty-object (XXL) (1065 ms)
-duplicate-as-attribute (XXL) (797 ms)
-redundant-object (XXL) (686 ms)
-reserved-name (XXL) (611 ms)
-application-without-as-attributes (XL) (597 ms)
-line-is-absent (XXL) (569 ms)
-incorrect-bytes-format (XXL) (548 ms)
-application-without-as-attributes (M) (507 ms)
-compound-name (XXL) (469 ms)
-bytes-without-data (XXL) (376 ms)
+application-without-as-attributes (XXL) (5405 ms)
+object-has-data (XXL) (2065 ms)
+duplicate-as-attribute (XXL) (1929 ms)
+empty-object (XXL) (1596 ms)
+redundant-object (XXL) (1233 ms)
+incorrect-bytes-format (XXL) (1226 ms)
+application-without-as-attributes (XL) (1025 ms)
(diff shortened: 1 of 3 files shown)


VasilevNStas:

volodya-lombrozo (changes requested):
@VasilevNStas Currently we create `S` ... `XXL` benchmarks using `jeo-maven-plugin` which works pretty well, but it generates XMIR that is a bit of artifical. Maybe we can reconsider our strategy to generating these files? Maybe it's better to generate them by parsing real EO programs? At least, we can substitute `S`,  `M` examples. What do u think?

VasilevNStas (commented):
@volodya-lombrozo 

Good suggestion about using real EO programs instead of bytecode-derived XMIR for the S/M benchmarks. That would make the benchmark data closer to what the lints actually process in production. I see two ways to do it and would like your call on the scope:

1. A focused follow-up that only adds this `scansXmirWithUnlints` benchmark for the `+unlint` path (what this PR does), and
2. A bigger refactor that replaces `jeo-maven-plugin` fixtures for S/M (or all sizes) with parsed EO programs.

I kept this PR to option 1 because switching the existing S..XXL fixtures touches `SourceSize`, `BenchmarkState` and the README table, and it is a separate concern from #1284 (which is specifically about the `+unlint` path having no coverage at all).

If you prefer, I can prepare option 2 as another PR after this one lands. Meanwhile, note this PR also surfaced a pre-existing infra issue: the `proguard` profile (active on JDK < 22) fails on the rultor merge (JDK 21) and locally on JDK 21 with `Obfuscation failed`, unrelated to the diff — I described it in #1238 with the `MAVEN_OPTS`/`-P!proguard` suggestion.

Happy to proceed either way.

VasilevNStas:
@volodya-lombrozo 
plz re-review

volodya-lombrozo (changes requested):
@VasilevNStas I don't have the new benchmark results in the list above. That said, I should.

VasilevNStas:
@volodya-lombrozo 
plz re-re-review

VasilevNStas:
@volodya-lombrozo Good point — the new `+unlint` results are now part of the standard benchmark, so they show up both in the `/benchmark` comparison list and in the README table.

What changed on the branch:

1. `SourceTest.benchmarkResults()` now also lints the `+unlint` fixture `unlint-ascii-only.eo` as a benchmark source, marked `UNLINT`. It produces 8 defects (including the `unlint-non-existing-defect` one), and `TimedLint` writes the per-lint rows into `lint-timings.csv` (e.g. `unlint-non-existing-defect (UNLINT)`, 8 ms). So a re-run of `/benchmark` will list them, and future changes to the `+unlint` path in `LtUnlintNonExistingDefect` become visible.
2. The README benchmark section now includes the new source (`Input: org/eolang/lints/unlint-ascii-only.eo (UNLINT source)`, `Lint time: 296 ms`) and the refreshed top-16. It will be regenerated by the GHA `benchmark` job after merge anyway.
3. The JMH `scansXmirWithUnlints` method stays as the controlled measurement of the whole scan.

Verified locally: `mvn clean install -Pqulice` (1016 rules) and `mvn clean test` (628 tests) pass.

PR #1291, opened 2026-08-21
fix(#1270): drop dead mode="dups" template in duplicate-aliases

### LINKED ISSUE #1270: Dead 'dups' mode template in duplicate-aliases.xsl emitting outdated 'error' element
## Problem

`duplicate-aliases.xsl` contains a dead template (`mode="dups"`) that is never invoked, and it emits an outdated `<error>` element instead of a `<defect>`.

## Details

`src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl:37-54`:

```xsl
<xsl:template match="meta" mode="dups">
  <xsl:for-each select="/object/metas/meta[head='alias']">
    ...
    <error>
      ...
    </error>
  </xsl:for-each>
</xsl:template>
```

Two issues:

1. **Dead code.** The `mode="dups"` template is declared but never invoked — there is no `<xsl:apply-templates ... mode="dups">` anywhere in the file (confirmed by grep: the only occurrence of `dups` is the template declaration). So this template never runs and produces nothing.

2. **Outdated element name.** Even if it were invoked, it emits `<error>` (the old defect format used by eo-parser), while the project's defect contract is `<defect>` (see `LtByXsl` / the existing packs for this lint, which assert against `/defects/defect`). The real, active logic of the lint lives in the default template above it, which correctly produces `<defect>`.

## Impact

- Dead code adds confusion: a reader may think the `mode="dups"` path is part of the lint's behavior, but it is unreachable.
- If someone "fixes" the invocation later, the lint would start emitting `<error>` elements that break the `/defects/defect` contract and would not be picked up by `LtByXsl`.
- Minor, but it is the kind of leftover that the `errors/empty-object` or `comment-without-dot` lints got cleaned up of before.

## Expected behaviour

Remove the dead `mode="dups"` template (and any `<error>` usage) so the file contains only the working `<defect>`-based logic. A simple code cleanup with no behavior change.

yegor256:
Closes #1270.

The `mode="dups"` template in `src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl` was dead code:

* Nothing invoked it — there is no `<xsl:apply-templates ... mode="dups"/>` in the stylesheet, and no other stylesheet imports `duplicate-aliases.xsl` (the only other `mode="dups"` in the repo is in `critical/duplicate-names.xsl`, which is a separate, self-contained stylesheet that does apply its own).
* It was a copy-paste leftover from `duplicate-names.xsl`: it matched `meta` yet tested `preceding-sibling::o/@name`, so it could not have produced anything even had it been invoked.
* It emitted `<error>` rather than the `<defect>` element that `LtByXsl` and the packs expect, so "fixing" the invocation later would have broken the `/defects/defect` contract.

The working logic stays untouched in the default `match="/"` template. All three `xsl:import`s are still used by it (`eo:lineno`, `eo:escape`, `eo:defect-context`), so they remain.

No behaviour change — pure removal of 18 unreachable lines.

## Verification

`mvn test` — 606 tests, 0 failures, 0 errors (including all 416 `LtByXslTest` pack assertions, among them the two `duplicate-aliases` packs).

https://claude.ai/code/session_017Axukgt3yd3R6V1cfWA446

---
_Generated by [Claude Code](https://claude.ai/code/session_017Axukgt3yd3R6V1cfWA446)_

### DIFF
--- a/src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl
+++ b/src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl
@@ -34,22 +34,4 @@
       </xsl:for-each>
     </defects>
   </xsl:template>
-  <xsl:template match="meta" mode="dups">
-    <xsl:for-each select="/object/metas/meta[head='alias']">
-      <xsl:variable name="x" select="."/>
-      <xsl:if test="preceding-sibling::o/@name = $x/@name">
-        <error>
-          <xsl:attribute name="line">
-            <xsl:value-of select="if (@line) then @line else '0'"/>
-          </xsl:attribute>
-          <xsl:attribute name="severity">
-            <xsl:text>error</xsl:text>
-          </xsl:attribute>
-          <xsl:text>The name </xsl:text>
-          <xsl:value-of select="eo:escape(@name)"/>
-          <xsl:text> is already in use</xsl:text>
-        </error>
-      </xsl:if>
-    </xsl:for-each>
-  </xsl:template>
 </xsl:stylesheet>


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2137](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1293, opened 2026-08-21
fix(#1269): align comment-is-too-wide motive with the 100-char limit

### LINKED ISSUE #1269: `comment-is-too-wide` lint and corresponding documentation mismatch in character limit
## Problem

The `comment-is-too-wide` lint enforces a width of **100 characters**, but its motive documents a limit of **80 characters**.

## Details

`src/main/resources/org/eolang/lints/comments/comment-is-too-wide.xsl:11`:

```xsl
<xsl:variable name="max" select="100"/>
```

But `src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md` says:

> Each comment should not be wider than 80 characters.

and its incorrect example is "a very long comment that contains more than eighty characters and should be flagged by the lint as too wide."

## Impact

- Code and documentation contradict each other: a comment of 90 characters is flagged by the lint but is fine according to the motive, and vice versa.
- Anyone relying on the motive to know the rule gets wrong guidance.
- The same confusion affects `comment-without-dot`/`comment-not-capitalized` docs only if they mention a width — this one explicitly does.

## Expected behaviour

Make the code and the motive agree. Either:
- change `$max` to `80` to match the documented rule, or
- update the motive (and its example) to say `100` if 100 is the intended limit.

A pack test asserting the actual threshold would prevent future drift.

yegor256:
Closes #1269.

The contradiction reported in the issue is real: `comment-is-too-wide.xsl` uses `$max = 100`, while `motives/comments/comment-is-too-wide.md` documented 80.

Git history says the docs are the stale side, not the code: commit 11d6464 ("100 chars per line in comments", 2025-01-22) deliberately changed `$max` from `80` to `100` in the XSL and left the motive untouched. So this PR fixes the documentation, and does not touch the lint's behaviour.

Changes:

* `motives/comments/comment-is-too-wide.md` — now says 100 characters; the "incorrect" example no longer claims "more than eighty characters" (it is 113 characters long, so it is still a genuine violation).
* Two new boundary packs pin the actual threshold, as the issue suggested, so the code and the docs cannot drift again silently:
  * `allows-comment-of-exactly-max-width.yaml` — a comment of exactly 100 characters produces no defect.
  * `catches-comment-one-char-wider-than-max.yaml` — a comment of 101 characters produces one warning, and the assertion pins the message text `The comment width is 101, while 100 is max allowed`.
* The two pre-existing `catches-*` packs said "more than eighty characters" in their EO input; reworded to "more than 100 characters" (lengths unchanged in effect — both inputs are still well over 100).

Verification: `mvn -Dtest=LtByXslTest test` → 418 tests, 0 failures (both new packs run, confirming the 100/101 boundary and the exact defect message).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LtRC58UjpmCN3yncfPAmhA)_

### DIFF
--- a/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
+++ b/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
@@ -1,11 +1,11 @@
 # Comment Is Too Wide
 
-Each comment should not be wider than 80 characters.
+Each comment should not be wider than 100 characters.
 
 Incorrect:
 
 ```eo
-# This is a very long comment that contains more than eighty characters and should be flagged by the lint as too wide.
+# This is a very long comment that contains more than 100 characters and should be flagged by the lint as too wide.
 [] > foo
 ```
 
--- a/src/test/resources/org/eolang/lints/packs/single/comment-is-too-wide/allows-comment-of-exactly-max-width.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/comment-is-too-wide/allows-comment-of-exactly-max-width.yaml
@@ -5,10 +5,9 @@
 sheets:
   - /org/eolang/lints/comments/comment-is-too-wide.xsl
 asserts:
-  - /defects[count(defect[@severity='warning'])=1]
-  - /defects/defect[@line='3']
+  - /defects[count(defect[@severity='warning'])=0]
 input: |
-  # This is a very long comment that contains more than 100 characters and should be flagged by the lint as too wide.
+  # This comment is exactly 100 characters wide, therefore the lint must not report it as a defect here.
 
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2145](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1294, opened 2026-08-21
fix(#1268): point LtTestNotVerb.motive() at a real motive document

### LINKED ISSUE #1268: `LtTestNotVerb.motive()` fails due to reference to a missing motive file
## Problem

`LtTestNotVerb.motive()` throws an exception because it references a motive file that does not exist.

## Details

`src/main/java/org/eolang/lints/LtTestNotVerb.java:59`:

```java
@Override
public String motive() throws IOException {
    return new MotiveFrom("misc", "test-object-is-not-verb-in-singular").asString();
}
```

The resource `org/eolang/motives/misc/test-object-is-not-verb-in-singular.md` does not exist in `src/main/resources/org/eolang/motives/misc/` — the only file there is `incorrect-test-object-name.md`. Since `MotiveFrom.asString()` reads the resource via `TextOf`, the missing file makes `motive()` fail with:

```
The resource "org/eolang/motives/misc/test-object-is-not-verb-in-singular.md" was not found ...
```

(I reproduced this locally by calling `new LtTestNotVerb().motive()`.)

## Impact

- The lint's own `name()` returns `unit-test-is-not-verb` (line 45), so the motive reference is doubly wrong: it uses a different name than the lint, and that name has no file.
- `motive()` is part of the public `Lint` contract and is delegated by wrappers (`LtUnlint.motive()`, `LtDfSticky.motive()`), so any tool that shows lint motives (e.g. IDE integrations, the motives workflow, or a future "explain this lint" feature) will crash on this lint.
- There is no unit test covering `LtTestNotVerb.motive()`, which is why this went unnoticed.

## Expected behaviour

`LtTestNotVerb.motive()` should return the content of a real motive file. Either:
- create `src/main/resources/org/eolang/motives/misc/unit-test-is-not-verb.md` (matching `name()`), or
- point `motive()` at the correct existing file.

A test asserting `motive()` succeeds (like `LtAsciiOnlyTest.explainsMotive`) should be added.

### LINKED ISSUE #1270: Dead 'dups' mode template in duplicate-aliases.xsl emitting outdated 'error' element

yegor256:
Closes #1268.

**Rescoped.** This PR originally covered #1268, #1269 and #1270. While it was in CI, master landed #1293 (for #1269) and #1291 (for #1270) independently. I synced with master and resolved every overlap in master's favour, so this branch now carries only the #1268 work. See the "Sync with master" section below.

### The bug

`LtTestNotVerb.motive()` loaded `org/eolang/motives/misc/test-object-is-not-verb-in-singular.md`, which does not exist anywhere in the resources, so the method threw on every call. The name it used also disagreed with `name()`, which returns `unit-test-is-not-verb`.

### The fix

- `motive()` now uses `new MotiveFrom("tests", this.name())`, the same `name()`-derived convention as `LtAsciiOnly`, `LtReservedName`, `LtSyntaxVersion` and `LtUnlintNonExistingDefect`.
- Added `src/main/resources/org/eolang/motives/tests/unit-test-is-not-verb.md`. The `tests` dimension is where the other unit-test lints keep their motives (`bad-test-name`, `unit-test-missing`, `unit-test-without-phi`).
- Added `LtTestNotVerbTest.explainsMotive`, in the shape of `LtAsciiOnlyTest.explainsMotive`.
- `PkMonoTest.linksEveryMotiveFileToALint` rejects any motive `.md` whose basename is not the `name()` of a lint produced by `PkMono`. `LtTestNotVerb` is currently detached from `PkMono` (see the `@todo #297` there), so the new file would have been reported as an orphan. The check now also tolerates that one detached name, and the assertion message says so.

One correction to the issue's impact section: because `LtTestNotVerb` is not registered in `MonoLints`, no public API path reaches it today, so nothing crashes on it right now. The broken resource reference is real all the same, and would fire the moment the lint is returned to `PkMono`.

### Sync with master

- `motives/comments/comment-is-too-wide.md` and the two boundary packs — conflicted with #1293. Took master's side. Master's `catches-comment-one-char-wider-than-max.yaml` also asserts the line number and the full defect message, so it is strictly stronger than what this branch had.
- `aliases/duplicate-aliases.xsl` — both sides deleted the same dead `mode="dups"` template, so it merged with no conflict.

The diff against master is now four files: `LtTestNotVerb.java`, the new motive document, `LtTestNotVerbTest.java` and `PkMonoTest.java`.

I dropped `closes #1270` from this description: the change that fixes it is #1291, already in master. That issue is still open even though its fix merged, so it likely just needs closing by hand.

### Verification (on the merged tree)

- Full `mvn surefire:test` → **609 tests, 0 failures, 0 errors**, 2 skipped, `BUILD SUCCESS`.
- `mvn -Pqulice verify -DskipTests` → `jtcop:check` and `qulice:check` both pass. (The run then fails in `proguard`, whose profile activates only on JDK `[,22)`; CI uses JDK 23, so it does not apply there.)
- `vale` 3.17.0 over `motives/` → 0 errors, 0 warnings; the single remaining suggestion is pre-existing in `metas/syntax-version.md`.
- `typos` and `yamllint` → clean.

### DIFF
--- a/src/main/java/org/eolang/lints/LtTestNotVerb.java
+++ b/src/main/java/org/eolang/lints/LtTestNotVerb.java
@@ -56,7 +56,7 @@ final class LtTestNotVerb implements Lint {
 
     @Override
     public String motive() throws IOException {
-        return new MotiveFrom("misc", "test-object-is-not-verb-in-singular").asString();
+        return new MotiveFrom("tests", this.name()).asString();
     }
 
     @Override
--- a/src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl
+++ b/src/main/resources/org/eolang/lints/aliases/duplicate-aliases.xsl
@@ -34,22 +34,4 @@
       </xsl:for-each>
     </defects>
   </xsl:template>
-  <xsl:template match="meta" mode="dups">
-    <xsl:for-each select="/object/metas/meta[head='alias']">
-      <xsl:variable name="x" select="."/>
-      <xsl:if test="preceding-sibling::o/@name = $x/@name">
-        <error>
-          <xsl:attribute name="line">
-            <xsl:value-of select="if (@line) then @line else '0'"/>
-          </xsl:attribute>
-          <xsl:attribute name="severity">
-            <xsl:text>error</xsl:text>
-          </xsl:attribute>
-          <xsl:text>The name </xsl:text>
-          <xsl:value-of select="eo:escape(@name)"/>
-          <xsl:text> is already in use</xsl:text>
-        </error>
-      </xsl:if>
-    </xsl:for-each>
-  </xsl:template>
 </xsl:stylesheet>
--- a/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
+++ b/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
@@ -1,11 +1,11 @@
 # Comment Is Too Wide
 
-Each line of a comment should not be wider than 100 characters.
+Each comment should not be wider than 100 characters.
 
 Incorrect:
 
 ```eo
-# This is a very long comment that contains more than one hundred characters and should be flagged by the lint as too wide.
+# This is a very long comment that contains more than 100 characters and should be flagged by the lint as too wide.
 [] > foo
 ```
 
--- a/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
+++ b/src/main/resources/org/eolang/motives/comments/comment-is-too-wide.md
@@ -1,11 +1,11 @@
 # Comment Is Too Wide
 
-Each comment should not be wider than 80 characters.
+Each line of a comment should not be wider than 100 characters.
 
 Incorrect:
 
 ```eo
-# This is a very long comment that contains more than eighty characters and should be flagged by the lint as too wide.
+# This is a very long comment that contains more than one hundred characters and should be flagged by the lint as too wide.
 [] > foo
 ```
 
--- a/src/main/resources/org/eolang/motives/metas/unsorted-metas.md
+++ b/src/main/resources/org/eolang/motives/metas/unsorted-metas.md
@@ -19,3 +19,9 @@ Correct:
 
 [] > foo
 ```
+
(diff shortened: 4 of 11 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1951](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1296, opened 2026-08-21
fix(#1295): catch bare ξ/ρ renames in anemic-getter

### LINKED ISSUE #1295: `anemic-getter` lint ignores redundant `ξ`/`ρ` chains in EO code
Consider this code:

```
[] > foo
  ^ > f
```

Here, `f` is an anemic getter of `^`: it adds nothing, since `^` is reachable
by its own symbol from the very place where `f` sits. However, the
`anemic-getter` lint stays silent about it.

The reason is in `anemic-getter.xsl`: the lint only looks for
`ξ.<name>`-based references whose `<name>` is a sibling attribute of the same
formation:

```xml
<xsl:for-each select="//o[@name and @name!='φ' and @base and matches(@base, '^ξ\.[^.]+$') and not(o)]">
  <xsl:variable name="ref" select="substring-after(@base, 'ξ.')"/>
  <xsl:if test="../o[@name=$ref]">
```

The parser turns `^ > f` into `<o name="f" base="ξ.ρ"/>`. That base does match
`^ξ\.[^.]+$`, so `$ref` becomes `ρ` — and since no sibling is named `ρ`, the
`xsl:if` never fires and no defect is reported.

The same holds for the rest of that family, none of which is caught today:

| EO | XMIR base | caught now? |
| --- | --- | --- |
| `x > y` (`x` is a sibling) | `ξ.x` | yes |
| `^ > f` | `ξ.ρ` | no |
| `^.^ > f` | `ξ.ρ.ρ` | no |
| `$ > f` | `ξ` | no |

All of them are the very same redundancy the lint exists to catch: a named
object whose whole body is nothing but a name for something that is already
nameable, unchanged, right where the getter is written.

Note that `^.bar > f` (base `ξ.ρ.bar`) is a different story — it reaches out of
the formation for an attribute of the parent, and `redundant-attachment`
already reasons about such references — so it should stay out of scope here.

Let's teach `anemic-getter` to complain about bare `ξ`/`ρ` chains too.

yegor256:
Closes #1295.

## The bug

```
[] > foo
  ^ > f
```

`f` is an anemic getter of `^`, but the lint said nothing.

The parser turns `^ > f` into `<o name="f" base="ξ.ρ"/>`. That base *did* match
the lint's `^ξ\.[^.]+$` filter, so the object reached the body of the
`xsl:for-each` — but `$ref` then became `ρ`, and the `../o[@name=$ref]` guard
looked for a sibling named `ρ`, which never exists. So the `xsl:if` never fired.

## The fix

The lint now recognizes the whole family of bare hop chains alongside the
sibling case, and names the hop in EO syntax in the defect message:

| EO | XMIR base | before | after |
| --- | --- | --- | --- |
| `x > y` (`x` is a sibling) | `ξ.x` | caught | caught |
| `^ > f` | `ξ.ρ` | silent | caught, renames `^` |
| `^.^ > f` | `ξ.ρ.ρ` | silent | caught, renames `^.^` |
| `$ > f` | `ξ` | silent | caught, renames `$` |
| `^.bar > f` | `ξ.ρ.bar` | silent | silent (out of scope) |

Each hop chain is the same redundancy the lint exists to catch: the hop is
already reachable by its own symbol right where the getter sits, so the second
name adds nothing.

Deliberately left alone:

- `^.bar > f` (base `ξ.ρ.bar`) — this reaches *past* the hops for an attribute
  of the parent, so `f` is a name for something other than the hop itself, not
  a rename of it. `redundant-attachment` already reasons about such references.
- `^ > @` — the `@name!='φ'` guard already covers decoration.
- `<o base="ξ" name="xi🌵"/>` — the parser synthesizes this marker, so the `ξ`
  branch excludes it, the same way `redundant-object` and friends do.

The per-branch logic moved out of the `for-each` predicate into an
`xsl:choose` that computes the rename target, which is `''` when the body is
not a rename at all.

## Tests

Six new packs under `packs/single/anemic-getter/`: `catches-rho-rename`,
`catches-double-rho-rename`, `catches-xi-rename`, `allows-rho-attribute`,
`allows-rho-application`, and `allows-synthetic-xi`.

`mvn test` is green — 614 tests, 0 failures, 0 errors, 2 skipped. `xcop`,
`markdownlint`, `yamllint`, and `vale` (against the updated motive) all pass
locally too.

The motive gained a section on `^` and `$`, and on why `^.bar` is not a rename.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RwCDozUQGr5K6PFbVDTgDh)_

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/anemic-getter.xsl
+++ b/src/main/resources/org/eolang/lints/misc/anemic-getter.xsl
@@ -9,17 +9,37 @@
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <!--
-  A named object whose whole body is a reference to a sibling attribute of
-  the same formation is an "anemic getter". For example, "title > t" inside
-  a "[title] > book" formation just renames "title" without adding anything.
-  Such renaming is redundant, since the original attribute may be used
-  directly, so we complain about it here.
+  A named object whose whole body is a reference to something the code can
+  already reach, unchanged, right where the object sits is an "anemic getter".
+  Two shapes of body qualify: a sibling attribute of the same formation, as in
+  "title > t" inside a "[title] > book" formation, and a bare chain of "ξ" and
+  "ρ" hops, as in "^ > f" or "$ > f", which "^" and "$" already say. Such
+  renaming is redundant, since the original may be used directly, so we
+  complain about it here. A reference that reaches past the hops for an
+  attribute, such as "^.bar", names something other than the hop itself, so it
+  is not a rename and stays out of scope.
   -->
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[@name and @name!='φ' and @base and matches(@base, '^ξ\.[^.]+$') and not(o)]">
+      <xsl:for-each select="//o[@name and @name!='φ' and not(@base='ξ' and @name='xi🌵') and @base and starts-with(@base, 'ξ') and not(o)]">
         <xsl:variable name="ref" select="substring-after(@base, 'ξ.')"/>
-        <xsl:if test="../o[@name=$ref]">
+        <xsl:variable name="target">
+          <xsl:choose>
+            <!-- "$", the formation itself -->
+            <xsl:when test="@base='ξ'">
+              <xsl:text>$</xsl:text>
+            </xsl:when>
+            <!-- "^", "^.^", and so on: one "^" per "ρ" hop -->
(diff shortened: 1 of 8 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (137 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1955](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1303, opened 2026-08-24
anemic-getter flags ^ > name, but naming the receiver is not a rename

### LINKED ISSUE #1300: `anemic-getter` lint incorrectly flags explicit receiver as a redundant rename
In `eo-runtime` we are making the receiver explicit: every formation head declares `^`, so the receiver becomes an ordinary void filled by dispatch rather than arriving through `AtWithRho` at load time. That turns eighty-four existing bindings into lint failures.

The idiom is a formation giving its receiver a name:

```
[^ x] > xor
  or. > @
    b.and x.not
    b.not.and x
  ^ > b
```

Once `ρ` is a declared void, `^ > b` reads as a getter renaming an attribute, and the lint says `The object "b" is a redundant getter, it just renames "ρ" without adding anything, use "ρ" directly instead`.

For a flat body the advice is fine. It stops being fine as soon as an inner formation uses the name, because the replacement is not `^` but a chain. In `chunk/to-output.eo` the body reads `^.^.^.allocated`, and dropping the binding makes that `^.^.^.^`.

Those two are equal only for the nesting that exists today. `allocated` names one thing no matter where it is written; `^.^.^.^` names whatever happens to be four hops up, so re-indenting a block or introducing a wrapper silently retargets it. We tried the rewrite and CI found the breakage — a `StackOverflowError` in the README snippets and errors across `map`, `while`, `range` and `input` — which is a fair summary of how easy it is to get the hop count wrong.

So the binding does add something: a name that survives nesting. That is the opposite of anemic.

Skip a getter whose source is `ρ`. Or, if the flat case is still worth reporting, only report when no descendant formation refers to the name, since that is exactly the case where the replacement is a plain `^`.

morphqdd:
Fixes #1300. `anemic-getter` flagged `^ > f` (and `^.^ > f`, and so on) as a
redundant rename of `^`, but naming the receiver this way adds something a
hop count does not: a name that survives nesting. `^.^` retargets silently
if a block using it is re-indented or wrapped in another formation; a name
written once as `f` does not.

Removed the branch that treated a bare chain of `^` hops as a rename
target. `$ > f`, a bare reference to the formation itself, is still
flagged, since `$` is already the shortest possible spelling.

Converted the two packs that exercised this shape, `catches-rho-rename` and
`catches-double-rho-rename`, into `allows-*` packs asserting zero defects,
and updated the motive doc to match.

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/anemic-getter.xsl
+++ b/src/main/resources/org/eolang/lints/misc/anemic-getter.xsl
@@ -11,13 +11,19 @@
   <!--
   A named object whose whole body is a reference to something the code can
   already reach, unchanged, right where the object sits is an "anemic getter".
-  Two shapes of body qualify: a sibling attribute of the same formation, as in
-  "title > t" inside a "[title] > book" formation, and a bare chain of "ξ" and
-  "ρ" hops, as in "^ > f" or "$ > f", which "^" and "$" already say. Such
-  renaming is redundant, since the original may be used directly, so we
-  complain about it here. A reference that reaches past the hops for an
-  attribute, such as "^.bar", names something other than the hop itself, so it
-  is not a rename and stays out of scope.
+  One shape of body qualifies: a sibling attribute of the same formation, as
+  in "title > t" inside a "[title] > book" formation, which is redundant
+  since the original may be used directly, so we complain about it here. A
+  bare chain of "ρ" hops, as in "^ > f" or "^.^ > f", stays out of scope even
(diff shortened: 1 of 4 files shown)


volodya-lombrozo:
Thanks for the PR! I can't merge this yet since the commits aren't signed, and signed commits are required for rultor to merge. Would you mind signing your commits and reopening? Docs here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #1304, opened 2026-08-24
many-void-attributes counts rho toward the five-void cap

### LINKED ISSUE #1299: `many-void-attributes` erroneously includes receiver `ρ` in void count
In `eo-runtime` we are making the receiver an explicit void: every formation head declares `^`, so the receiver becomes an ordinary attribute filled by dispatch instead of arriving through `AtWithRho` at load time. That works everywhere except one place, where `many-void-attributes` stops it.

`string/regex.eo` has a formation with five voids:

```
[position start from to groups] > matched
```

Adding the receiver makes it six, and the lint fires: `The object "matched" has more than 5 void attributes, it's too many`. The formation did not grow a parameter — it wrote down one it already had. Nothing about how hard `matched` is to read or call changed.

The cap is measuring argument count, and `ρ` is not an argument. It is the receiver, and every formation has exactly one whether or not the head spells it. Counting it means the real cap is four for any object that declares its receiver and five for one that leaves it implicit, which rewards leaving it out.

Skip a void named `ρ` when counting. That keeps the rule measuring what it means to measure, and the same reasoning likely applies to any other lint that counts voids.

For now `matched` keeps its implicit receiver as the single exception, which is a wart we would like to remove.

morphqdd:
Fixes #1299. `many-void-attributes` counts a void named `ρ` toward the
five-void cap, but `ρ` is the receiver, not an argument: every formation
has one whether or not the head spells it out. Counting it means the real
cap is four for a formation that declares its receiver and five for one
that leaves it implicit, which rewards leaving it out.

Excluded a void named `ρ` from the count. Added a regression pack with
five real voids plus a named receiver, asserting zero defects.

### DIFF
--- a/src/main/resources/org/eolang/lints/errors/many-void-attributes.xsl
+++ b/src/main/resources/org/eolang/lints/errors/many-void-attributes.xsl
@@ -9,9 +9,14 @@
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:variable name="max" select="5"/>
+  <!--
+  A void named "ρ" is the receiver, not an argument: every formation has one
+  whether or not the head spells it out, so counting it toward the cap
+  rewards leaving it implicit (objectionary/lints#1299).
+  -->
   <xsl:template match="/">
     <defects>
-      <xsl:for-each-group select="//o[@name and @base='∅' and not(o)]" group-by="generate-id(..)">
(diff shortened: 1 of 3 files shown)


volodya-lombrozo:
Thanks for the contribution! Unfortunately I can't merge this as-is because the commits aren't signed, and rultor requires signed commits to merge. Could you sign your commits and reopen? Here's a guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #1313, opened 2026-08-26
feat(#1207): add excessive-visibility lint

### LINKED ISSUE #1207: `excessive-visibility` lint doesn't flag unused public methods
Consider this object:

```
[t] > phrase
  pos.gte 0 > multi-words
  t.index-of " " > pos

  (book "Object Thinking").multi-words ++> can-find-space-in-name
```

Here, the `pos` is not used by any tests. It can and should be made private:

```
[t] > phrase
  pos.gte 0 > multi-words
  t.index-of " " >> pos
                ^^^^
                here
```

Let's create a lint that would catch this. If no tests refer to a method, it must be private/obfuscated (`>>`).

yegor256:
fix #1207

## What

A new XSL-based lint `excessive-visibility` is added:

- `src/main/resources/org/eolang/lints/design/excessive-visibility.xsl` — the lint
- `src/main/resources/org/eolang/motives/design/excessive-visibility.md` — the motive
- `src/test/resources/org/eolang/lints/packs/single/excessive-visibility/*.yaml` — test packs

## Why

Consider this object:

```eo
[t] > phrase
  pos.gte 0 > multi-words
  t.index-of " " > pos

  [] +> checks-phrase
    (phrase "Object Thinking").multi-words > @
```

Here, `pos` is never referred to by any test, so it isn't really part of the
object's public interface. It should be declared private instead:

```eo
[t] > phrase
  pos.gte 0 > multi-words
  t.index-of " " >> pos

  [] +> checks-phrase
    (phrase "Object Thinking").multi-words > @
```

## How it works

For every abstract (named) formation that contains at least one unit test
(a nested attribute whose name starts with `+` or `-`, per the parser's own
test-attribute marker), the lint looks at each of its other direct, public,
non-void attributes (no `@local`, i.e. not already declared with `>>`, and
not `φ`). A method counts as "used by a test" when a test refers to it
either:

- by calling it on itself, which the parser resolves into a
  `ξ.ρ...ρ.<name>`-prefixed chain of self-references, or
- by chaining a dotted call onto some other expression, e.g.
  `(phrase "x").multi-words`, which the parser resolves into a
  `.<name>`-prefixed postfix base.

A public method that no test refers to either way is flagged, with a
message suggesting `>>` instead of `>`.

Objects without any unit tests at all are skipped entirely, to avoid
flagging every public attribute of code that simply isn't unit-tested yet —
this lint is only about attributes whose *tested* surface doesn't match
their declared visibility.

The self/postfix reference patterns were verified against the actual XMIR
produced by the current `eo-parser` (0.62.1) for both the self-call and the
chained-call cases, rather than assumed.

## Tests

- `catches-unused-method` — a public method untouched by the object's test
  is caught
- `allows-method-used-directly-by-test` — a bare self-call from the test
  clears the method
- `allows-method-used-via-qualified-call` — a dotted call onto a fresh
  instance clears the method
- `allows-object-without-tests` — no defects when the formation has no
  tests at all
- `allows-already-private-method` — a method already declared with `>>` is
  never a candidate
- `allows-decoratee-attribute` — the `@` decoratee itself is never flagged

`mvn test` (all `LtByXslTest` packs, 437 tests) and `mvn -Pqulice verify`
both pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GJEhUWEihLeSuKoHCyPX3w)_

### DIFF
--- a/src/main/resources/org/eolang/lints/design/excessive-visibility.xsl
+++ b/src/main/resources/org/eolang/lints/design/excessive-visibility.xsl
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" id="excessive-visibility" version="2.0">
+  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/test-name.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[eo:abstract(.) and @name]">
+        <xsl:variable name="tests" select="o[eo:test-name(@name)]"/>
+        <xsl:if test="exists($tests)">
+          <xsl:for-each select="o[@name and @base and @base != '∅' and @name != 'φ' and not(@local) and not(eo:test-name(@name))]">
+            <xsl:variable name="method" select="@name"/>
+            <!--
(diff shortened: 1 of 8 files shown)


yegor256:
@rultor merge

0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (159 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1510](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1320, opened 2026-08-28
feat(#1317): add identity-object lint

### LINKED ISSUE #1317: Linter doesn't flag usage of manual identity object instead of `I` glyph
The language has a glyph for the identity object — `I`, which is
`[x] > (φ ↦ x)`. Still, we keep writing that object out by hand. In
objectionary/eo both `printf.eo` and `as-decimal.eo` carried a hand-made
one-attribute formation that a glyph could spell, and nothing complained
(objectionary/eo#7822). I'd like a lint that catches an identity object
wherever it exists and suggests replacing it with `I`.

Incorrect:

```eo
# Jeff.
[] > jeff
  x > [x] >> id
  foo id > @
```

```eo
# Jeff.
[] > jeff
  [value] > pass
    value > @
  foo pass > @
```

Correct:

```eo
# Jeff.
[] > jeff
  foo I > @
```

## What to catch

A formation is the identity object when it binds exactly one void, decorates
that very void, and holds nothing else. In XMIR that is:

```xml
<o name="id">
  <o name="x" base="∅"/>
  <o name="φ" base="ξ.x"/>
</o>
```

The predicate already exists, in `eo:identity` of eo's
`eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`
(objectionary/eo#6834), where the printer uses it to print any such formation
back as the glyph:

```xpath
eo:abstract($o) and not(eo:has-data($o)) and count($o/o) = 2
  and eo:void($o/o[1]) and empty($o/o[1]/(@local, @type, @args))
  and $o/o[2]/@name = $eo:phi and empty($o/o[2]/o)
  and empty($o/o[2]/@const)
  and $o/o[2]/@base = concat($eo:xi, '.', $o/o[1]/@name)
```

The tight tests there are load-bearing and worth keeping: exactly two children,
because a third is an argument applied to the formation and the glyph has
nowhere to put it; and the φ base rooted at `ξ`, because an unrooted name
denotes some other object entirely.

## The one catch

`I` desugars to precisely the same XMIR — eo's `Emissions.identity` emits a
baseless object, a void named `x`, and a φ bound to it — so a lint that matches
the shape alone would flag every use of the glyph, `I > sanitized` included.
Two ways out, and I have no strong preference:

1. Have eo's parser mark the desugared glyph (an attribute on the `<o>`), and
   let the lint report the shape only where the mark is absent. Cleanest, but it
   needs a change in eo first.
2. Tell them apart by position: for the desugared glyph the formation, its void
   and its φ all carry the same `@line` and `@pos`, since all three come from
   the one character; a hand-written formation spreads over more than one line
   or more than one column. No change in eo needed, at the price of a heuristic.

## Deliverables

* `src/main/resources/org/eolang/lints/misc/identity-object.xsl` — `warning`
  severity, and maybe `experimental="true"` at first, the way
  `design/no-attribute-formation.xsl` does it
* `src/main/resources/org/eolang/motives/misc/identity-object.md`
* a test pack under
  `src/test/resources/org/eolang/lints/packs/single/identity-object/`, with a
  catching case and a no-catch case — the no-catch one guarding against the
  false positive above, plus the near misses the predicate must let through: a
  formation with two voids, one whose φ is applied to something, and one whose φ
  reaches a name that isn't its void

yegor256:
## Summary

- Adds a new `identity-object` lint that flags a handwritten identity formation (`[x] > (φ ↦ x)`) that could be replaced with the `I` glyph.
- A formation is only matched when it binds exactly one void and decorates that same void via `ξ`; the glyph's own desugared shape is told apart by its formation, void and `φ` all sharing the same `@line`/`@pos`, since all three come from the single glyph character (per the heuristic proposed in the issue).
- Adds the lint's motive doc and a YAML test pack covering the catching case, the desugared-glyph false positive, the three near-miss shapes called out in the issue (two voids, `φ` applied to something, `φ` reaching a name that isn't its own void), an anonymous-formation case, and a line-0/context case.

Closes #1317

## Test plan

- [x] `mvn test -Dtest=LtByXslTest#testsAllLintsByEo` — all 427 YAML packs pass, including the new `identity-object` pack
- [x] `mvn test` — full test suite passes
- [x] `mvn verify -DskipTests` — qulice static analysis passes
- [x] `xcop`, `yamllint`, `reuse lint`, `typos`, `markdownlint-cli2`, `vale` — all pass on the new files (no errors)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KicXH9j9HV3JxU57JHJr33)_

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/identity-object.xsl
+++ b/src/main/resources/org/eolang/lints/misc/identity-object.xsl
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="identity-object" version="2.0">
+  <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
+  <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
+  <xsl:import href="/org/eolang/funcs/escape.xsl"/>
+  <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="/">
+    <defects>
+      <xsl:for-each select="//o[eo:abstract(.) and count(o) = 2 and o[1][@base='∅'][@name][not(o)] and o[2][@name='φ'][not(@const)][not(o)][@base=concat('ξ.', o[1]/@name)] and not(@line = o[1]/@line and @line = o[2]/@line and @pos = o[1]/@pos and @pos = o[2]/@pos)]">
+        <xsl:element name="defect">
+          <xsl:variable name="line" select="eo:lineno(@line)"/>
+          <xsl:attribute name="line">
+            <xsl:value-of select="$line"/>
+          </xsl:attribute>
+          <xsl:if test="$line = '0'">
+            <xsl:attribute name="context">
+              <xsl:value-of select="eo:defect-context(.)"/>
+            </xsl:attribute>
+          </xsl:if>
+          <xsl:attribute name="severity">
+            <xsl:text>warning</xsl:text>
+          </xsl:attribute>
+          <xsl:attribute name="experimental">true</xsl:attribute>
+          <xsl:text>The </xsl:text>
+          <xsl:choose>
+            <xsl:when test="@name">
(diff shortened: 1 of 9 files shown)


yegor256:
Thanks for the contribution!

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (168 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1704](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1322, opened 2026-08-29
fix(#640): recognize chained-dot usage of $-aliased attributes

### LINKED ISSUE #640: `redundant-object` lint rule fails to recognize object usage above its declaration level

yegor256:
Fixes #640.

## The bug

`redundant-object` flags a `$`-aliased attribute (e.g. `$ > self`) as redundant when its only usage in the program is a trailing dot-attribute access on the result of an unrelated call chain, such as:

```eo
[bts] > bytes-as-input
  [size] > read
    ((input-block bts --).read size).self > @

    [data buffer] > input-block
      buffer > @
      $ > self
```

Here `self` (declared via `$ > self`) is used through `.self` at the end of the `((input-block bts --).read size).self` chain. The lint's `referenced-by-name` key only indexes usages whose `@base` starts with `ξ(.ρ)*.` or `Φ.pkg.`, so this chained-dot access is invisible to it, the usage count stays at `<= 1`, and the lint incorrectly reports `self` as redundant — even though `self` cannot actually be inlined (`(...).$` is not valid EO).

## Fix

`src/main/resources/org/eolang/lints/misc/redundant-object.xsl`: add a `self-used-via-chained-dot` check that looks for any `o[@base]` in the document matching `^\.NAME(?:\.[\w-]+)*$`, and skip the warning for `$`-aliased candidates (`@base = 'ξ'`) when such a usage exists. This mirrors the existing `self-alias-crosses-formation` guard already in the file.

## Test

New pack `redundant-object/allows-self-used-via-chained-dot.yaml` reproduces the exact example from the issue and asserts zero defects. The full `LtByXslTest#testsAllLintsByEo` suite (438 tests) and the complete `mvn test` run pass locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ELh6FCEHCn5UN8GYxAYqr7)_

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
+++ b/src/main/resources/org/eolang/lints/misc/redundant-object.xsl
@@ -15,7 +15,8 @@
       <xsl:for-each select="//o[generate-id() != $top and @name and @name != 'φ' and @base and @base != '∅' and not(@base='ξ' and @name='xi🌵')]">
         <xsl:variable name="in-recursive" select="some $r in key('referenced-by-name', @name), $f in $r/ancestor::o[@name and not(@base)] satisfies exists(key('referenced-by-name', $f/@name) intersect $f/descendant::o)"/>
         <xsl:variable name="self-alias-crosses-formation" select="@base = 'ξ' and (some $r in key('referenced-by-name', @name) satisfies matches($r/@base, '^ξ\.ρ\.'))"/>
-        <xsl:if test="count(key('referenced-by-name', @name))&lt;=1 and not(@name and o[1]/@base = 'Φ.dataized') and not($in-recursive) and not($self-alias-crosses-formation)">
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for the contribution!

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1321, opened 2026-08-28
Compiler fails to convert certain self-recursion shapes into Java loops


yegor256:
The EO compiler turns the self-recursion of a nested formation into a Java loop at transpile time (objectionary/eo#5783): `recursion-to-loop.xsl` handles a tail self-call in an `if.` branch or in the last step of a `seq *`, and `recursion-to-cps.xsl`, coming with the same issue, rewrites the remaining self-calls into continuation-passing style first. Both sheets bail out silently on a handful of shapes, and the program then runs as plain recursion, one Java stack block per step, which is exactly what #5783 set out to remove. Nothing tells the programmer which of the two happened.

The shapes that are skipped are all syntactic, so a lint could name them at the site: a formation with a `λ`; a self-call whose result is read through `.φ` or `.@`; a self-call inside a nested formation or inside the arguments of another self-call; a formation referred to other than by a plain application `^.F args` (as a decoratee, a dispatch receiver, or with a named binding in its `@`); an attribute holding a self-call that is read from outside the formation, as in `range` where the next step is a lazily read attribute; mutual recursion between two formations; and a self-call spelled as a top-level `Φ.name` rather than `^.name`. A warning such as "the recursion in `rec-contains` cannot be turned into a loop, the self-call is the receiver of a dispatch" would let the author rewrite the object, as objectionary/eo#6239 did for `repeated` by hand.

A new stylesheet next to `src/main/resources/org/eolang/lints/design/no-attribute-formation.xsl`, say `design/unoptimizable-recursion.xsl`, can find each nested formation `F` whose `φ` reaches an `o[@base='ξ.ρ.F']` and report the first of the conditions above that holds; the tail test from `_recursion.xsl` in objectionary/eo (an `if.` branch is `@as` α0 or α1, a `seq *` step is `@as` α1 of a `Φ.tuple` under `Φ.seq`) is small enough to inline. A warning, not an error, since the program is still correct, only slower and hungrier for stack.

yegor256:
@yegor256 could you take a look at this one when you have a moment? Happy to clarify any of the skipped shapes if the list above is unclear.

0pdd:
@yegor256 the puzzle [#1332](https://github.com/objectionary/lints/issues/1332) is still not solved.

0pdd:
@yegor256 the puzzle [#1416](https://github.com/objectionary/lints/issues/1416) is still not solved; solved: [#unknown]().

0pdd:
[...]

PR #1328, opened 2026-08-30
add unoptimizable-recursion lint

morphqdd:
The compiler turns a tail self-call of a nested formation into a Java loop, and stays silent when it cannot. The program then runs as plain recursion, one stack block per step, and nothing at the site says why.

This adds `design/unoptimizable-recursion.xsl`. For every nested formation with a bound `φ` it collects the self-calls, and when none of them sits in a tail position it reports one warning naming the reason: the formation is an atom, its body reads its own `φ`, the self-call is the receiver of a dispatch, or no self-call is in a tail position at all.

The tail test is the one the compiler runs in `_recursion.xsl`: a branch of an `if.`, or the last element of the tuple a `seq` is applied to. Both shapes already carry `@as` in parsed XMIR, so the test ports over as is.

Five packs: a tail call in an `if.` branch, a tail call as the last step of a `seq *`, a formation that never calls itself, a self-call under an operation, and a dispatch on the self-call.

The lint is `experimental`, like `no-attribute-formation` next to it.

The remaining shapes from #1321, a `Φ.name` self-call, a formation reached other than by a plain application, an attribute with a self-call read from outside, and mutual recursion, are left as a `@todo` in the stylesheet.

### DIFF
--- a/.github/styles/config/vocabularies/Custom/accept.txt
+++ b/.github/styles/config/vocabularies/Custom/accept.txt
@@ -19,3 +19,5 @@ inlined
 [Ss]ubgraph
 [Aa]pplication
 [Ff]orma(e)?
+[Tt]ranspile(d|s|r)?
+[Uu]noptimizable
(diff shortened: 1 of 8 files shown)


morphqdd:
@volodya-lombrozo, please review

yegor256:
Thanks for the contribution!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (216 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6947](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1330, opened 2026-08-31
#1307: give atom-in-atom its packs and check for the next one

### LINKED ISSUE #1307: `atom-in-atom` lint lacks corresponding pack and misses testing
`LtByXslTest.catchesLostXsls` checks that every YAML pack has an XSL beside it.
Nothing checks the other direction, and one lint has no pack at all.

What happens.

```
lints:    81 XSL files under src/main/resources/org/eolang/lints
packs:    80 directories under src/test/resources/org/eolang/lints/packs/single
missing:  atom-in-atom
```

There is no `atom-in-atom` directory and no YAML anywhere that names it, so the
rule is never run by the test suite.

The rule itself works. Handing it a document with an atom inside an atom:

```xml
<object author='tests'>
  <o name='outer' line='1'>
    <o name='λ' atom='Φ.bytes'/>
    <o name='inner' line='2'>
      <o name='λ' atom='Φ.bytes'/>
    </o>
  </o>
</object>
```

gives

```xml
<defects>
   <defect line="1" severity="error">Atom "outer" may not have any attributes,
   even if they are atoms, which however exist: "inner"</defect>
</defects>
```

Worth knowing while writing the pack: that shape cannot come from EO source. The
parser refuses both ways of writing it,

```
[] > outer /Q.bytes
  [] > inner /Q.bytes
```

with `Atom cannot contain inner objects; only '+>' and '->' test attributes are
allowed in an atom body`, and

```
[] > outer /Q.bytes

  [] +> t /Q.bytes
```

with `trailing garbage after name suffix`. XMIR reaches the linter from other
tools as well, so the rule still has a job to do.

What should happen. `atom-in-atom` should have a pack, the way the other eighty
lints do, and the test that guards packs should guard this direction too, so
that a lint added without one is caught.

morphqdd:
`atom-in-atom` was the one lint with no pack directory, so nothing ever ran it. The rule itself works: handed an atom that carries another atom it reports the error, and handed an atom with nothing inside it stays quiet, which is what the two packs here assert.

The existing check only looks one way, from a pack to its XSL. The new one looks the other way, so a lint added without packs fails the build instead of sitting unexercised.

Closes #1307


yegor256:
This branch was stacked on top of #1328 (now merged), so this diff still shows the same unoptimizable-recursion.xsl addition plus the atom-in-atom pack. Needs a rebase onto master to isolate just the #1307 fix, and a manual look either way. Not merging as-is.

---

PR #1361, opened 2026-09-05
Scope last() to each formation in phi-is-not-first

### LINKED ISSUE #1359: `phi-is-not-first` only reports the last misplaced `φ` in the document, ignoring multiple formations per file
`phi-is-not-first` takes the last misplaced `φ` in the **whole document**, so when two different formations each have one, only the second is reported.

`src/main/resources/org/eolang/lints/names/phi-is-not-first.xsl:13`:

```xml
<xsl:apply-templates select="(//o[@name='φ'][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
```

`//o[...]` gathers every misplaced `φ` in the file; the parentheses make `[last()]` apply to that whole sequence, not to each formation's own. Whatever the file contains, exactly one defect comes out.

Reporting one per formation is deliberate — `packs/single/phi-is-not-first/catches-only-last-phi.yaml` pins it, with two misplaced `φ` inside one formation and `count(defect)=1` expected. Every pack in that directory has a single formation, so the behaviour that is pinned is the within-formation one.

Across formations the same expression does something nobody asked for:

```
# No comments.
[] > foo
  y > bar
  x > @

# No comments.
[] > bar
  y > baz
  x > @
```

`foo`'s misplaced `φ` is dropped and only `bar`'s is reported. A file with ten such formations yields one defect, and each fix reveals the next one on the following run.

Scoping `last()` to the formation rather than to the document keeps the pinned behaviour and restores the rest:

```xml
<xsl:for-each select="//o[o[@name='φ'][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]]]">
  <xsl:apply-templates select="(o[@name='φ'][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
</xsl:for-each>
```

so `[last()]` picks the last offender within each parent, and every parent gets its own defect.

A pack with two formations, each carrying a misplaced `φ`, expecting `count(defect)=2`, would keep the two cases apart — the existing packs cannot tell the difference, since they never have a second formation to lose.

yegor256:
Fixes #1359.

`phi-is-not-first.xsl` selected `(//o[@name='φ']...)[last()]`. The
parentheses made `[last()]` apply to the sequence of misplaced `φ`
nodes across the *entire document*, not to each formation's own
sequence. A file with two formations, each with one misplaced `φ`,
only reported the second formation's defect — the first was silently
dropped.

The fix iterates over every formation that contains a misplaced `φ`
and applies `[last()]` scoped to that formation's own children, so
each formation gets its own defect (still reporting only the last
offender within a single formation, preserving the previously pinned
behavior).

Added `catches-phi-in-each-formation.yaml`, a pack with two sibling
formations each carrying a misplaced `φ`, asserting `2` defects
(one per formation, at their respective lines).

## Test plan

- [x] `mvn test -Dtest=LtByXslTest` — all 489 tests pass, including
      the new pack and the existing `validatesPacksAgainstXsdSchema`
      / `phi-is-not-first` packs.

https://claude.ai/code/session_01UzCPMT23UNH1YL28ZDJLQc

---
_Generated by [Claude Code](https://claude.ai/code/session_01UzCPMT23UNH1YL28ZDJLQc)_

### DIFF
--- a/src/main/resources/org/eolang/lints/names/phi-is-not-first.xsl
+++ b/src/main/resources/org/eolang/lints/names/phi-is-not-first.xsl
@@ -10,7 +10,9 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:apply-templates select="(//o[@name='φ'][not(@pipe)][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
+      <xsl:for-each select="//o[o[@name='φ'][not(@pipe)][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]]]">
+        <xsl:apply-templates select="(o[@name='φ'][not(@pipe)][preceding-sibling::o[not(@base='∅') and not(@base='ξ' and @name='xi🌵')]])[last()]" mode="unordered"/>
+      </xsl:for-each>
     </defects>
   </xsl:template>
   <xsl:template match="o" mode="unordered">
--- a/src/test/resources/org/eolang/lints/packs/single/phi-is-not-first/catches-phi-in-each-formation.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/phi-is-not-first/catches-phi-in-each-formation.yaml
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/names/phi-is-not-first.xsl
+defects:
+  - severity: warning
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks for the fix!

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2115](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1362, opened 2026-09-05
fix(#1353): report the physical line of a wide line inside a comment

### LINKED ISSUE #1353: `comment-is-too-wide` incorrectly reports over-wide lines at the start of multi-line comments
`comment-is-too-wide` splits a multi-line comment into its lines and reports each over-wide line at the comment's **first** line, so the number in the defect points at a line that is very likely fine.

`src/main/resources/org/eolang/lints/comments/comment-is-too-wide.xsl`:

```xml
<xsl:variable name="line" select="if (@line) then @line else '0'"/>
<xsl:variable name="lines" select="tokenize(replace(., '\n', '&#10;'), '&#10;')"/>
<xsl:choose>
  <xsl:when test="count($lines) &gt; 1">
    <xsl:for-each select="$lines[string-length(.) &gt; $max]">
      <xsl:element name="defect">
        <xsl:attribute name="line">
          <xsl:value-of select="$line"/>
        </xsl:attribute>
```

`$line` is read once, before the loop, from the comment element. Every defect the loop emits carries that same value. A comment whose fourth line is 120 characters wide is reported on the line the comment *starts* on, and a comment with three over-wide lines produces three defects that all name that one line.

The rule is the only one in this directory that walks inside a comment, so it is the only one where the comment's own line and the offending line differ. Its single-line branch, further down, is correct by construction — there the two are the same line.

The offset is available where the defect is built, but not from `position()`: the predicate `[string-length(.) > $max]` filters first, so inside that `for-each` the position is the index among the *offending* lines, not among all of them. Walking every line and testing inside the loop keeps the index meaningful:

```xml
<xsl:for-each select="$lines">
  <xsl:if test="string-length(.) &gt; $max">
    <xsl:element name="defect">
      <xsl:attribute name="line">
        <xsl:value-of select="number($line) + position() - 1"/>
      </xsl:attribute>
```

with `number($line)` guarding against the `'0'` fallback being concatenated rather than added.

Worth checking the `'0'` path in the same change. When the comment carries no `@line`, every defect gets line `0` and a `context` attribute, which is the intended fallback; with the offset added, `0 + position() - 1` would produce line numbers counting from zero for a comment whose real position is unknown. Keeping the fallback at `0` when `@line` is absent, and offsetting only when it is present, avoids inventing a location.

gemshrine:
Fixes #1353.

Every over-wide line inside a multi-line comment was reported at the line the comment starts on, so a comment with three wide lines produced three defects all naming the same line.

The loop now walks every line and tests inside it, so `position()` is the index among all lines rather than among the offending ones, and the defect carries `@line + position() - 1`.

The `'0'` fallback is left alone: when the comment carries no `@line` there is no location to offset from, and counting from zero would invent one.

This closes the `@todo #1046` in the same file, and `catches-multiline-wide-comment.yaml` now asserts line 2, the line that is actually too wide. That pack fails on master and passes here.

### DIFF
--- a/src/main/resources/org/eolang/lints/comments/comment-is-too-wide.xsl
+++ b/src/main/resources/org/eolang/lints/comments/comment-is-too-wide.xsl
@@ -7,15 +7,6 @@
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <!--
-  @todo #1046:30min Report the exact physical line of the too-wide line
-  inside a multi-line comment. Right now every offending line is reported
-  with the comment block's own start @line, since the XSL has no way to
-  compute the offset of a line within the tokenized comment text. Add a
-  $pos-based offset (comment @line + position() - 1) so
-  catches-multiline-wide-comment.yaml can assert on line 2 (the actual
-  offending line) instead of line 1.
-  -->
   <xsl:template match="/">
     <xsl:variable name="max" select="100"/>
     <defects>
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks for closing out that old todo!

---

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4022](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1363, opened 2026-09-05
fix(#1358): do not count the piece after a trailing newline as a line

### LINKED ISSUE #1358: `*-line-out-of-listing` rules incorrectly count lines in source files ending with a newline
The three `*-line-out-of-listing` rules compute the listing's length by counting the pieces `tokenize` returns, and a source file ending with a newline yields one piece more than it has lines. So each rule's bound is one too high and a line exactly one past the end of the file is never reported.

The same expression is in all three, on line 9 of each:

```xml
<xsl:variable name="max" select="count(tokenize(/object/listing, '&#10;'))"/>
```

`lines/error-line-out-of-listing.xsl`, `lines/meta-line-out-of-listing.xsl` and `lines/object-line-out-of-listing.xsl`, with the check on line 12 in each:

```xml
<xsl:for-each select="/object/errors/error[number(@line) and @line &gt; $max]">
```

`tokenize("a\nb\n", "\n")` gives `("a", "b", "")` — three items for two lines. EO sources end with a newline, so the trailing empty piece is always there. Measured on a parsed XMIR from `eo-runtime`:

```
listing tail                         : '    a false &gt; closed\n'
count(tokenize(listing, LF))  ($max) : 233
real line count of bool.eo           : 232
```

An `@line` of 233 on that file is out of the listing and passes the `@line > $max` test unreported. The rules do catch anything further out, so the effect is narrow — exactly one line — but it is the most likely wrong value: a line number produced by counting from the wrong end, or by adding one to the last line, lands precisely there.

Subtracting the trailing piece fixes all three:

```xml
<xsl:variable name="max" select="count(tokenize(/object/listing, '&#10;')[. != '' or position() != last()])"/>
```

or, more plainly, by tokenizing the listing with its trailing newline removed:

```xml
<xsl:variable name="max" select="count(tokenize(replace(/object/listing, '&#10;$', ''), '&#10;'))"/>
```

The second form also behaves correctly for a listing that does not end with a newline, where the current expression is already right and must not lose a line.

Worth checking the empty case while the expression is being touched: for an empty listing, `tokenize('', '\n')` returns one empty piece, so `$max` is 1 and a defect on line 1 of a file with no content goes unreported.

gemshrine:
Fixes #1358.

`tokenize("a\nb\n", "\n")` returns three pieces for two lines, and EO sources end with a newline, so `$max` was one too high in all three rules. A line exactly one past the end of the file passed the `@line > $max` test unreported, which is the most likely wrong value: counting from the wrong end, or adding one to the last line, lands precisely there.

The listing is tokenized with its trailing newline removed, which also leaves a listing that does not end with one counted correctly.

One pack per rule, each with a listing that ends in a newline and a defect on the line just past it. All three fail on master and pass here.

### DIFF
--- a/src/main/resources/org/eolang/lints/lines/error-line-out-of-listing.xsl
+++ b/src/main/resources/org/eolang/lints/lines/error-line-out-of-listing.xsl
@@ -6,7 +6,7 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="error-line-out-of-listing" version="2.0">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:variable name="max" select="count(tokenize(/object/listing, '&#10;'))"/>
+  <xsl:variable name="max" select="count(tokenize(replace(/object/listing, '&#10;$', ''), '&#10;'))"/>
   <xsl:template match="/">
     <defects>
       <xsl:for-each select="/object/errors/error[number(@line) and @line &gt; $max]">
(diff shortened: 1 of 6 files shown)


yegor256:
Merged, nice catch on the off-by-one, thanks!

---

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4038](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1364, opened 2026-09-05
fix(#1357): match the printf dispatch that EO actually emits

### LINKED ISSUE #1357: `sprintf-without-formatters` rule ineffective due to removed `Φ.txt.sprintf` object, failing to match `base='.printf'`
`sprintf-without-formatters` selects the same object `wrong-sprintf-arguments` does, and that object no longer exists in EO, so this rule cannot fire either.

`src/main/resources/org/eolang/lints/misc/sprintf-without-formatters.xsl:20`:

```xml
<xsl:for-each select="//o[@base='Φ.txt.sprintf']">
```

There is no `txt` package in `objectionary/eo` any more and no `sprintf` object anywhere in the runtime. Formatting is `eo-runtime/src/main/eo/string/printf.eo`, called as a method on the format string, so what a current XMIR carries is `base=".printf"`:

```
(nothing)

$ grep -rlo "\.printf" eo-runtime/target/eo/1-parse/*.xmir | wc -l
13
```

The check it performs is still worth having. It reports a format template with none of the supported conversions, on the grounds that formatting without a formatter is pointless — a real mistake that is easy to make and invisible at runtime, since the template simply comes back unchanged.

Adapting it means matching the dispatch rather than the base:

```xml
<xsl:for-each select="//o[@base='.printf']">
```

and reading the template from the receiver, which for a dispatch is emitted as a preceding sibling rather than as a child, so the `$text` extraction on lines 21-38 has to follow it there.

Two things are worth revisiting in the same change. The message on line 60 names `"%s", "%d", "%f", "%x", "%b"` as the supported set, and current `printf` handles more than that — flags, widths and precisions such as `%5d`, `%08.2f`, `%.0f` all work — so a template using only those would now be reported as having no formatters at all. And the message text names `"Φ.txt.sprintf"` three times, which will read oddly once the rule matches something else.

The sibling rule `wrong-sprintf-arguments` has the same dead selector and is reported separately, since the two need different adaptations: that one is about arity and tuple shape, this one about the template's content.

gemshrine:
Fixes #1357.

The rule selected `Φ.txt.sprintf`, which current EO does not emit at all: there is no `txt` package any more, and formatting is a method on the format string, so an XMIR carries `base=".printf"`. The rule could not fire on anything.

It matches the dispatch now. The template is read from the receiver, which in a real XMIR is the child without an `as` attribute:

```xml
<o base=".printf">
  <o base="Φ.string">…the format…</o>
  <o as="α0" base="Φ.tuple">…the arguments…</o>
</o>
```

Checked against `eo-runtime/target/eo/1-parse/bytes/as-i16.xmir` rather than assumed.

The detection of a formatter changed with it. Counting the byte pairs of `%s`, `%d`, `%f`, `%x` and `%b` misses everything current `printf` accepts around the conversion letter — `%08.2f`, `%5d`, `%1$s` — and would report a template using only those as having no formatters. The template is decoded first and then matched against the specifier grammar `printf.eo` documents: `%[N$][flags][width][.precision]conversion`, with a doubled percent taken out first, since it is a literal one.

The message names `.printf` instead of the object that no longer exists.

The existing packs are rewritten to the current call form, and two more cover a padded specifier that must pass and a template whose only percent is an escaped one.

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/sprintf-without-formatters.xsl
+++ b/src/main/resources/org/eolang/lints/misc/sprintf-without-formatters.xsl
@@ -17,31 +17,16 @@
   </xsl:function>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[@base='Φ.txt.sprintf']">
-        <xsl:variable name="text" select="o[1][@base='Φ.string']/o[1][@base='Φ.bytes']/o/text()"/>
-        <xsl:variable name="txt" select="translate($text, '-', '')"/>
-        <xsl:variable name="formatters">
-          <xsl:variable name="txt" select="translate($text, '-', '')"/>
-          <!-- First, replace %% with a unique placeholder to avoid counting it as a formatter -->
-          <xsl:variable name="escaped" select="replace($txt, '(25)(25)', 'ESCAPED_PERCENT')"/>
-          <!-- %s -->
-          <xsl:variable name="strings" select="count(tokenize($escaped, '2573'))"/>
-          <!-- %d -->
-          <xsl:variable name="numbers" select="count(tokenize($escaped, '2564'))"/>
-          <!-- %f -->
-          <xsl:variable name="floats" select="count(tokenize($escaped, '2566'))"/>
-          <!-- %x -->
-          <xsl:variable name="bytes" select="count(tokenize($escaped, '2578'))"/>
-          <!-- %b -->
-          <xsl:variable name="bools" select="count(tokenize($escaped, '2562'))"/>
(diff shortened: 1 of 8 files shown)


yegor256:
Merged, thanks for tracking down the actual dispatch shape and fixing this properly!

---

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4054](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1365, opened 2026-09-05
fix(#1355): check the arguments of the printf dispatch EO emits

### LINKED ISSUE #1355: `wrong-sprintf-arguments` rule ineffective due to outdated `Φ.txt.sprintf` reference
`wrong-sprintf-arguments` looks for an object that current EO does not have, so the rule can never fire.

`src/main/resources/org/eolang/lints/misc/wrong-sprintf-arguments.xsl:36`:

```xml
<xsl:for-each select="//o[@base='Φ.txt.sprintf']">
```

There is no `txt` package in `objectionary/eo` any more, and no object called `sprintf` anywhere in the runtime. Formatting lives in `eo-runtime/src/main/eo/string/printf.eo` and is reached as a method on the format string, so in XMIR it appears as a dispatch rather than as a package-level base:

```
(nothing)

$ grep -o 'base="[^"]*printf[^"]*"' eo-runtime/target/eo/1-parse/string/printf.xmir
base=".printf"
```

Thirteen files in the runtime alone use `.printf`, and none of them can be checked by this rule.

The three defects it knows how to report — more than two arguments on lines 53-56, a second argument that is not a well-formed `Φ.tuple` on line 93, and a tuple whose length disagrees with the format template on line 111 — are all still worth reporting. They simply have no subject: the rule matches a base string that stopped being emitted when `txt.sprintf` was replaced.

Adapting it means matching the dispatch instead of the base, and moving the format string from the first argument to the receiver:

```xml
<xsl:for-each select="//o[@base='.printf']">
```

with the argument indices shifted by one, since `"fmt".printf (* args)` puts the template on the receiver and leaves exactly one argument — the tuple — where the old form had two. That also simplifies the arity check: the old "only 2 arguments" rule becomes "only 1".

Worth confirming against a current XMIR before the change lands, since the receiver of a `.printf` dispatch is emitted as a preceding sibling rather than as a child, so the format string has to be reached through that sibling rather than through `o[1]`.

gemshrine:
Fixes #1355.

The rule selected `Φ.txt.sprintf`, an object current EO does not emit: there is no `txt` package any more, and formatting is a method on the format string, so an XMIR carries `base=".printf"`. None of the three checks the rule performs could fire on anything.

The dispatch puts the template on the receiver and leaves exactly one argument, so:

- the arity check becomes "only 1 argument", counted over the children that carry `@as`;
- the template is read from the child without `@as`, which is where a real XMIR keeps the receiver (checked against `eo-runtime/target/eo/1-parse/bytes/as-i16.xmir`, not assumed);
- the tuple is the first `@as` child rather than `o[2]`.

Counting the specifiers changed with it. Looking for the byte pairs of `%s`, `%d`, `%f`, `%x` and `%b` misses everything current `printf` accepts around the conversion letter, so `"%08.2f"` counted as zero specifiers and the rule would have reported a correct call as passing one argument too many. The template bytes are matched against the specifier grammar `printf.eo` documents: `%[N$][flags][width][.precision]conversion`.

A template using positional specifiers is left alone by the count check, since `%1$s` twice needs one argument, not two.

The packs are rewritten to the current call form, with three added: padded specifiers, a positional template, and the renamed arity pack. The eight `catches-*` packs fail without the change and pass with it.

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/wrong-sprintf-arguments.xsl
+++ b/src/main/resources/org/eolang/lints/misc/wrong-sprintf-arguments.xsl
@@ -7,6 +7,10 @@
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
+  <!-- One format specifier, matched over the hex bytes of the template:
+  a percent, an optional N$ position, the "-" and "0" flags, a width,
+  an optional .precision, and one of the five conversion letters -->
+  <xsl:variable name="eo:specifier" select="'25((3[0-9])+24)?(2D|30)*(3[0-9])*(2E(3[0-9])+)?(73|64|66|78|62)'"/>
   <!-- Find arguments in tuple -->
   <xsl:template match="o" mode="arguments" as="xs:integer">
     <xsl:choose>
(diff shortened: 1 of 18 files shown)


yegor256:
Merged, thanks for the thorough fix and test coverage!

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (133 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4066](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1395, opened 2026-09-06
fix(#1377): anchor the second alternative of the package regex

### LINKED ISSUE #1377: incorrect-package regex is not anchored and lets invalid package names pass
## Problem

`incorrect-package` lets invalid package names pass because the second alternative of its regex is not anchored to the start of the string.

## Root cause

`src/main/resources/org/eolang/lints/metas/incorrect-package.xsl:17`:

```xml
<xsl:if test="not(matches($tail, $eo:package-regexp))">
```

with the pattern

```xml
<xsl:variable name="eo:package-regexp" select="concat('[a-z][a-z0-9]*(\.[a-z0-9]+)*|[\p{L}\p{M}\p{N}_]+(\.[...])*$', ...)"/>
```

The second alternative `[\p{L}\p{M}\p{N}_]+(\.[...])*$` has no leading `^`, and `matches()` in XPath is not anchored by default. Verified: `-bad` and `org..example` are accepted without a defect.

## Minimal example

```eo
+package -bad
+architect yegor@256.com
+version 0.0.1

[] > main
```

`+package -bad` does not trigger `incorrect-package`, although the name is clearly invalid.

## Expected behavior

The whole regex should be anchored (a single `^...$` expression, or `^` added to the second alternative), so that invalid package names are always caught.

## Related

- `metas/prohibited-package.xsl` — sibling rule for runtime-repo packages.

gemshrine:
Fixes #1377.

`matches()` in XPath is not anchored, and the second alternative of the `+package` pattern ended with `$` but never started with `^`. Any tail whose last segments look like a name therefore passed, whatever stood in front of them: `-bad` matched on `bad`, and `org..example` matched on `example`.

Anchoring alone is not enough. The alternative is also the one that lets a package carry `-` and `@`, which the existing packs rely on (`c.d-test.x-yz`, `universe@obj`, `a.b-привет.c-大家好`); once it is anchored, its character class has to hold those two characters itself instead of skipping past them. A segment still has to start with something other than a hyphen, so `-bad` stays a defect while `d-test` does not.

Checked every name the packs of this lint carry, plus the two from the issue, against the sheet directly:

| `+package` | before | after |
| --- | --- | --- |
| `-bad` | passes | defect |
| `org..example` | passes | defect |
| `test.` | defect | defect |
| `привет, как дела?` | defect | defect |
| `a.b.c`, `X.Y.Z`, `Foo`, `hello.world123`, `test_xyz` | passes | passes |
| `j$org.j$eolang.j$jeo`, `com.foo$bar` | passes | passes |
| `universe@obj`, `c.d-test.x-yz`, `a.b-привет.c-大家好`, `foo.бар` | passes | passes |

The two new cases are in a pack of their own.

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/incorrect-package.xsl
+++ b/src/main/resources/org/eolang/lints/metas/incorrect-package.xsl
@@ -14,7 +14,7 @@
         <xsl:variable name="meta-head" select="head"/>
         <xsl:variable name="meta-tail" select="tail"/>
         <xsl:variable name="first" select="normalize-space(substring-before(concat($meta-tail, ' '), ' '))"/>
-        <xsl:if test="$meta-head='package' and not(matches($first, '^[a-zA-Z0-9_]+(\.[a-zA-Z0-9_]+|\$[a-zA-Z0-9_]*)*$|[\p{L}\p{M}\p{N}_]+(\.[\p{L}\p{M}\p{N}_]+)*$'))">
+        <xsl:if test="$meta-head='package' and not(matches($first, '^[a-zA-Z0-9_]+(\.[a-zA-Z0-9_]+|\$[a-zA-Z0-9_]*)*$|^[\p{L}\p{M}\p{N}_@][\p{L}\p{M}\p{N}_@-]*(\.[\p{L}\p{M}\p{N}_@][\p{L}\p{M}\p{N}_@-]*)*$'))">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3962](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1396, opened 2026-09-06
fix(#1380): let the also meta be known

### LINKED ISSUE #1380: unknown-metas reports the new 'also' meta as unknown
## Problem

`unknown-metas` reports `+also` as an unknown meta, although the modern parser emits it (see `add-default-package.xsl` in eo-parser).

## Root cause

`src/main/resources/org/eolang/lints/metas/unknown-metas.xsl:15` whitelists a fixed set of meta heads (`package`, `architect`, `version`, `home`, `spdx`, `rt`, `lint`, `tests`, `unlint`, `import`, `alias`, `compile`, `repo`, `register`, `system`, ...) — `also` is not among them, so any `+also ...` is flagged as unknown.

## Minimal example

```eo
+package x
+architect yegor@256.com
+version 0.0.1
+also app

[] > main
```

`+also app` is reported as an unknown meta.

## Expected behavior

`also` should be added to the list of known metas (and kept in sync with the metas the parser can emit).

## Related

- eo-parser `add-default-package.xsl` (introduces `+also`).
- Existing packs under `packs/single/unknown-metas/`.

gemshrine:
Fixes #1380.

`also` is a meta the parser knows: `add-default-package.xsl` and `resolve-aliases.xsl` in eo-parser both match `meta[head='also']` and rewrite its parts the way they rewrite an alias. The whitelist here never got it, so every `+also` is reported as a meta whose "usage has no effect", which is the opposite of what happens to it.

Added `also` to the list, to the pack, and to the motive, next to `alias`, since the two are read by the same sheets.

Run against the sheet directly, `+also app` no longer reports, while an actually unknown meta still does.

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/unknown-metas.xsl
+++ b/src/main/resources/org/eolang/lints/metas/unknown-metas.xsl
@@ -12,7 +12,7 @@
     <defects>
       <xsl:for-each select="/object/metas/meta">
         <xsl:variable name="meta-head" select="head"/>
-        <xsl:variable name="predefined" select="('package', 'alias', 'version', 'rt', 'architect', 'home', 'unlint', 'probe', 'spdx', 'syntax')"/>
+        <xsl:variable name="predefined" select="('package', 'alias', 'also', 'version', 'rt', 'architect', 'home', 'unlint', 'probe', 'spdx', 'syntax')"/>
         <xsl:if test="not($meta-head = $predefined)">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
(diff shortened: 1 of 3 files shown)


yegor256:

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3974](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1397, opened 2026-09-06
fix(#1379): compare the home URL without its trailing slash

### LINKED ISSUE #1379: prohibited-package false positive when +home URL has a trailing slash
## Problem

A `+home` URL with a trailing slash (`https://github.com/objectionary/eo/`) causes false positives in `prohibited-package` (and breaks `mandatory-package`), because the home-repo comparison is an exact string match.

## Root cause

`src/main/resources/org/eolang/funcs/home-object.xsl:27`:

```xml
<xsl:if test="... and $tail = $eo:home-repo and ...">
```

`eo:home-object` returns `true` only when the `+home` tail equals `$eo:home-repo` exactly. `$eo:home-repo` is built without a trailing slash, so a legitimate `+home https://github.com/objectionary/eo/` (with slash) does not match → `prohibited-package.xsl:17` treats an ordinary runtime object as prohibited.

## Minimal example

```eo
+package x
+architect yegor@256.com
+version 0.0.1
+home https://github.com/objectionary/eo/

[] > main
```

An ordinary object is flagged by `prohibited-package` because the trailing slash breaks the exact match.

## Expected behavior

Normalize the home URL (strip a trailing `/`) or use `starts-with`, so that URLs with and without a trailing slash behave identically.

## Related

- `metas/prohibited-package.xsl:17`
- `metas/mandatory-package.xsl` — affected by the same comparison.

gemshrine:
Fixes #1379.

`eo:home-object` asks whether the `+home` tail equals `https://github.com/objectionary/eo` character for character. A tail that names the same repository with a trailing slash does not, so both lints that share this function get the answer wrong: `prohibited-package` reports an ordinary runtime object as one that must not sit in `org.eolang`, and `mandatory-package` asks a runtime object for the `+package` it is not supposed to carry.

The tail is now compared with its trailing slashes cut off. Run against the sheets directly, with `+home https://github.com/objectionary/eo/`:

| sheet | object | before | after |
| --- | --- | --- | --- |
| `prohibited-package` | `bytes`, `+package org.eolang` | 1 defect | none |
| `mandatory-package` | `win32`, no `+package` | 1 defect | none |
| `prohibited-package` | `foo`, `+package org.eolang` | 1 defect | 1 defect |

A pack for each of the two lints.

### DIFF
--- a/src/main/resources/org/eolang/funcs/home-object.xsl
+++ b/src/main/resources/org/eolang/funcs/home-object.xsl
@@ -11,13 +11,16 @@
   neither "+package" (mandatory-package) nor "+package org.eolang"
   (prohibited-package) should be flagged on them. Keep this list in sync
   with the home repository; both lints share it from here so it never goes
-  stale in two places at once.
+  stale in two places at once. The "+home" tail is compared with its trailing
+  slashes cut off, because "https://github.com/objectionary/eo/" names the
+  same repository as the URL without the slash, and an exact match would let
+  a slash turn a runtime object into a defect of both lints (#1379).
   -->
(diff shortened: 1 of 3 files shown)


yegor256:

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3986](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1398, opened 2026-09-06
fix(#1382): check an spdx meta that carries one part

### LINKED ISSUE #1382: incorrect-spdx skips single-segment +spdx values like 'MIT'
## Problem

`incorrect-spdx` does not check single-segment `+spdx` values at all — e.g. `+spdx MIT` (with no space-separated `SPDX-...:` header) is silently accepted.

## Root cause

`src/main/resources/org/eolang/lints/metas/incorrect-spdx.xsl:13`:

```xml
<xsl:for-each select="/object/metas/meta[head='spdx' and count(part) > 1]">
```

The `count(part) > 1` guard skips metas whose tail has a single whitespace-separated part, so `+spdx MIT` never reaches the validation logic.

## Minimal example

```eo
+package x
+architect yegor@256.com
+version 0.0.1
+spdx MIT

[] > main
```

`+spdx MIT` is not flagged, even though a valid SPDX meta must start with `SPDX-...:`.

## Expected behavior

Single-part `+spdx` values (e.g. `MIT`, `Apache-2.0`) should be flagged as incorrect, or the guard should be reworked so the validation covers them.

## Related

- `metas/mandatory-spdx.xsl` — sibling rule.
- Existing packs under `packs/single/incorrect-spdx/`.

gemshrine:
Fixes #1382.

The loop skipped any `+spdx` whose tail is a single word, so `+spdx MIT` was never looked at. A compliant tail always has at least two parts, the `SPDX-...:` header and what follows it, which means a one-part tail can only ever be wrong, and it was the one shape the lint let through.

The guard now stands at `count(part) > 0`: a tail is checked as soon as there is one, and a bare `+spdx` with nothing after it is left where it was, since there is no header there to name in the message.

Against the sheet:

| `+spdx` tail | before | after |
| --- | --- | --- |
| `MIT` | passes | defect |
| `Hello world` | defect | defect |
| `SPDX-License-Identifier: MIT` | passes | passes |
| `SPDX-FileCopyrightText: Copyright (c) 2026 Objectionary.com` | passes | passes |
| nothing | passes | passes |

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/incorrect-spdx.xsl
+++ b/src/main/resources/org/eolang/lints/metas/incorrect-spdx.xsl
@@ -10,7 +10,7 @@
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="/object/metas/meta[head/text()='spdx' and count(part) &gt; 1]">
+      <xsl:for-each select="/object/metas/meta[head/text()='spdx' and count(part) &gt; 0]">
         <xsl:variable name="meta-tail" select="tail"/>
         <xsl:variable name="header" select="normalize-space(substring-before(concat($meta-tail, ' '), ' '))"/>
         <xsl:if test="$header!='SPDX-FileCopyrightText:' and $header!='SPDX-License-Identifier:'">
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3998](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1400, opened 2026-09-06
fix(#1372): use anyMatch for range unlints covering part of the defects

### LINKED ISSUE #1372: LtUnlintNonExistingDefect falsely warns for range +unlint that covers only part of the defects
## Problem

`LtUnlintNonExistingDefect` produces a false **"doesn't make sense"** warning when a line-range `+unlint` covers only *part* of the defects of a lint.

## Root cause

`src/main/java/org/eolang/lints/DefectMissing.java:50` uses `allMatch`:

```java
missing = !lines.stream().allMatch(new UnlintInRange(unlint));
```

For `+unlint foo:10-20` with `foo` defects on lines `[5, 15]`, `allMatch` returns `false` → the whole unlint is reported as "doesn't make sense", even though line `15` is legitimately suppressed by the range.

## Minimal example

EO program with `+unlint ascii-only:1-10` where the `ascii-only` lint reports defects both inside and outside the range (e.g. two non-ASCII comments on lines 5 and 15).

Expected: no warning — line 5 is suppressed by the range, line 15 is not and is reported by the lint itself.

Actual: `unlint-non-existing-defect` warns *"Suppressing 'ascii-only:1-10' doesn't make sense, because there are no defects with it"* (well — a defect exists, on line 15, but the warning fires anyway).

## Expected behavior

A range unlint makes sense if **any** defect of that lint falls into the range → use `anyMatch` (the same semantics `LtUnlint` uses via `removeIf(inRange)`, see `src/main/java/org/eolang/lints/LtUnlint.java:63`).

## Related

- #1232/#950 (line-specific `+unlint` handling)
- The pack `unlint/allows-unlint-for-defects-in-the-line-range.yaml` only covers the all-inside case, so the partial case is untested.

VasilevNStas:
Fixes #1372.

## Problem
`DefectMissing` reported a range `+unlint foo:10-20` as "doesn't make sense" whenever *any* defect of `foo` was outside the range, even if the range covered another defect. E.g. defects on lines `[5, 15]` with `foo:10-20` produced a false warning, although line `15` is legitimately suppressed.

## Root cause
`src/main/java/org/eolang/lints/DefectMissing.java:50` used `allMatch(new UnlintInRange(unlint))`, so the presence of a single out-of-range defect made the whole suppression "missing".

## Solution
Use `anyMatch` — the suppression makes sense as long as at least one defect falls into the range (the same semantics `LtUnlint` applies via `removeIf(inRange)`).

## Changes
- `src/main/java/org/eolang/lints/DefectMissing.java` — `allMatch` → `anyMatch`.
- `src/test/java/org/eolang/lints/DefectMissingTest.java` — `returnsTrueIfSomeLineIsOutOfRange` (which encoded the buggy behavior) replaced with `returnsFalseWhenSomeLineIsInTheRange`.

## Tests
- `mvn clean install -Pqulice` (1019 rules) — pass
- `mvn clean test` (668 tests) — pass

### DIFF
--- a/src/main/java/org/eolang/lints/DefectMissing.java
+++ b/src/main/java/org/eolang/lints/DefectMissing.java
@@ -47,7 +47,7 @@ final class DefectMissing implements Function<String, Boolean> {
             if (lines == null) {
                 missing = !this.excluded.contains(name);
             } else {
-                missing = !lines.stream().allMatch(new UnlintInRange(unlint));
+                missing = !lines.stream().anyMatch(new UnlintInRange(unlint));
             }
         } else {
             final Set<String> names;
--- a/src/test/java/org/eolang/lints/DefectMissingTest.java
+++ b/src/test/java/org/eolang/lints/DefectMissingTest.java
@@ -81,12 +81,12 @@ final class DefectMissingTest {
     }
 
     @Test
-    void returnsTrueIfSomeLineIsOutOfRange() {
+    void returnsFalseWhenSomeLineIsInTheRange() {
         MatcherAssert.assertThat(
-            "Defect should be missing, but it was not",
+            "Defect should not be missing, since at least one defect is covered by the range",
             new DefectMissing(new MapOf<>("noisy-lint", new ListOf<>(4, 6)), new ListOf<>())
                 .apply("noisy-lint:6-10"),
-            Matchers.equalTo(true)
+            Matchers.equalTo(false)
         );
     }
 


VasilevNStas:
@yegor256 
please review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1934](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1408, opened 2026-09-06
perf(#1389): reuse defect cache in unlint-non-existing-defect

### LINKED ISSUE #1389: `unlint-non-existing-defect` excessively re-runs XSL transformations, causing major performance issue
## Problem

`unlint-non-existing-defect` is by far the slowest lint (14513 ms on the XXL benchmark, ~10x the next one), because it re-runs every lint's XSL transformation for every `+unlint` name instead of reusing the defects already computed by the main pass.

## Root cause

`src/main/java/org/eolang/lints/LtUnlintNonExistingDefect.java:114-123`:

```java
private Set<String> existing(final String name) {
    return this.lints.stream()
        .filter(l -> l.name().equals(name))
        .flatMap(l -> l.defects(this.xml).stream())   // raw lint, no cache
        .map(Defect::line)
        ...
}
```

`existing()` calls `lint.defects(xmir)` on the **raw** lints (not wrapped in `LtDfSticky`), so for every `+unlint foo` the full XSL transformation for `foo` runs again — duplicating the work already done in the main pipeline where the same lint is wrapped in `LtDfSticky`. The results are thrown away right after the `DefectMissing` check.

## Expected behavior

Reuse the defects from the main pass (or share the `LtDfSticky`-wrapped instances), so `+unlint` validation does not double the XSL workload. Estimated impact: the XXL timing should drop to the same order as the other lints.

## Related

- `PkMono.java` — where lints get their `LtUnlint`/`LtDfSticky` decorations.
- #1349 (volodya) — coverage of this lint; complementary.
- README benchmark table shows the current 14513 ms figure.

VasilevNStas:
Fixes #1389.

## Problem
`unlint-non-existing-defect` was the slowest lint (14513 ms on XXL, ~10x the next). Its `existing()` re-ran every referenced lint's XSL transformation for every `+unlint` name, because it operated on raw lints — duplicating work already done by the main pass (where lints are wrapped in `LtDfSticky`).

## Solution
Rebuild the lint decorations so the `LtDfSticky`-wrapped instances are shared between the main pass and `existing()`. A single `Sticky` iterable materializes each `LtDfSticky` wrapper once; `existing()` reads from the same cached instances (raw defects, before `LtUnlint` suppression), so the XSL workload is not duplicated.

## Changes
- `src/main/java/org/eolang/lints/PkMono.java` — extract `build()`; wrap all lints in `LtDfSticky` into one `Sticky` iterable; the main pass wraps it with `LtUnlint`, while `LtUnlintNonExistingDefect` (wrapped as `LtDfSticky(LtUnlint(...))`) receives the shared cached iterable. Removed the now-unused `Chained`.

## Tests
- `mvn clean install -Pqulice` (668 tests) — pass

### DIFF
--- a/src/main/java/org/eolang/lints/PkMono.java
+++ b/src/main/java/org/eolang/lints/PkMono.java
@@ -5,10 +5,10 @@
 package org.eolang.lints;
 
 import javax.annotation.concurrent.ThreadSafe;
-import org.cactoos.func.Chained;
 import org.cactoos.iterable.IterableEnvelope;
 import org.cactoos.iterable.Joined;
 import org.cactoos.iterable.Mapped;
+import org.cactoos.iterable.Sticky;
 import org.cactoos.list.ListOf;
 
 /**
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
please review this PR

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2082](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1409, opened 2026-09-06
fix(#1390): bound the LtDfSticky cache by content

### LINKED ISSUE #1390: `LtDfSticky` cache grows infinitely due to unbounded limit and faulty equal check in key
## Problem

The `LtDfSticky` defect cache (backed by cactoos `StickyBiFunc`) never evicts entries, and its key has broken equality semantics, so in a long-lived JVM the cache grows without bounds while rarely hitting.

## Root cause

1. `StickyBiFunc` (cactoos 0.61.1) stores results in a `LinkedHashMap` with capacity `Integer.MAX_VALUE` — no eviction. Because `Source.MONO` is a static singleton (`src/main/java/org/eolang/lints/Source.java:27-31`), every XMIR document ever analyzed by the JVM stays in memory forever → OOM under batch/daemon workloads.

2. The cache key is the `XML` object itself. `XMLDocument.equals` compares by `toString()` (content) but `hashCode` is node-identity based (`Node.hashCode()` in jcabi-xml) — so two documents with identical content produce different hash codes, and the cache almost never hits for different `Source` instances, while still growing per new document.

## Expected behavior

- Bound the cache (capacity or time-based eviction), or scope it per `Source` instance instead of a static singleton.
- Make the key content-based (hash by `toString()`), so identical documents share cache entries.

## Related

- `src/main/java/org/eolang/lints/LtDfSticky.java:38-43` — the wrapper.
- `P1` (#1389) — the related `unlint-non-existing-defect` performance issue.

VasilevNStas:
Fixes #1390.

## Problem
The `LtDfSticky` defect cache (cactoos `StickyFunc`) never evicted entries and keyed on the `XML` object, whose `hashCode` is node-identity based while `equals` compares content. In a long-lived JVM (static `Source.MONO`), every analyzed XMIR stayed in memory forever (OOM under batch), and identical documents almost never hit the cache.

## Solution
Replace the unbounded `StickyFunc` with a bounded Guava `Cache` keyed by the document content (`xmir.toString()`):
- `maximumSize(1_000)` bounds memory in long-lived JVMs.
- The content-based key makes identical documents share a cache entry, fixing the broken `equals`/`hashCode` contract of `XMLDocument`.

## Changes
- `src/main/java/org/eolang/lints/LtDfSticky.java` — bounded Guava cache keyed by content; dropped the unused `LtDfSticky(Lint, Func)` ctor and cactoos `Func` wrappers.
- `src/test/java/org/eolang/lints/LtDfStickyTest.java` — new `reusesCacheForIdenticalContent`: two distinct `XML` objects with identical content yield a single origin call.

## Tests
- `mvn clean install -Pqulice` (669 tests) — pass

### DIFF
--- a/src/main/java/org/eolang/lints/LtDfSticky.java
+++ b/src/main/java/org/eolang/lints/LtDfSticky.java
@@ -4,16 +4,18 @@
  */
 package org.eolang.lints;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.jcabi.xml.XML;
 import java.io.IOException;
 import java.util.Collection;
-import org.cactoos.Func;
-import org.cactoos.func.IoCheckedFunc;
-import org.cactoos.func.StickyFunc;
-import org.cactoos.func.SyncFunc;
+import java.util.concurrent.ExecutionException;
 
 /**
  * Lint caching decorator that calls defects method only once. Uses in memory storage for caching.
+ *
+ * <p>The cache is bounded and keyed by the XMIR content, so a long-lived JVM does not run
+ * out of memory while repeatedly analyzing the same documents.</p>
  *
  * <p>This class is thread-safe.</p>
  *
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2098](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1412, opened 2026-09-06
fix(#1381): detect self-recursion in all three spellings

### LINKED ISSUE #1381: unoptimizable-recursion misses self-recursion written by plain name (Φ.<name>)
## Problem

`unoptimizable-recursion` misses the most common way to write a self-recursive call — by plain name — because the modern parser resolves such calls to `Φ.<name>`, while the lint only looks for `ξ.ρ.<name>` (the `^`-style form).

## Root cause

`src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl:64-66`:

```xml
<xsl:variable name="self" select="concat('ξ.ρ.', @name)"/>
<xsl:variable name="calls" select="... //o[@base = $self] ..."/>
```

Verified with a real program:

```eo
[] > calc
  if. > @
    n.eq 0
    1
    n.mul (calc (n.sub 1))
```

The inner call `calc (...)` is resolved by eo-parser 0.63.0 to `@base='Φ.calc'`, so the lint finds no `ξ.ρ.calc` and reports nothing. The same recursion written as `^.calc` *does* produce `ξ.ρ.calc` and is caught — so the lint's behavior depends on how the recursion is spelled, not on the recursion itself.

## Expected behavior

Self-calls should be detected for both `Φ.<name>` and `ξ.ρ.<name>` forms (and ideally `$<name>`), so the lint is consistent regardless of the referencing syntax.

## Related

- #1321 (compiler fails to convert certain self-recursion shapes into Java loops) — the underlying design issue.
- `packs/single/unoptimizable-recursion/` — existing packs use the `^` form only.

VasilevNStas:
Fixes #1381.

## Problem
`unoptimizable-recursion` missed self-recursion written by plain name. The modern parser resolves such calls to `Φ.<name>`, while the lint only looked for `ξ.ρ.<name>` (the `^`-style form), so the recursion went unreported depending on how it was spelled.

## Solution
Detect self-calls in all three forms the parser emits:
- plain name → `Φ.<name>`
- `^` → `ξ.ρ.<name>`
- `$` → `ξ.<name>`

Three helper functions were added: `eo:is-self-call` (self-call, including a dispatch on it), `eo:is-self-dispatch` (the self-call used as the receiver of a dispatch), and `eo:is-self-tail` (a plain self-call, with no method dispatched — only these can sit in a tail position). The selector no longer requires `parent::o`, so a top-level formation recursing by plain name is also reported.

## Changes
- `src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl` — new helper functions; `$calls`/`$loops` use them; dropped `parent::o` from the selector.
- `src/test/resources/org/eolang/lints/packs/single/unoptimizable-recursion/catches-plain-name-self-call.yaml` — plain-name recursion under an operation is caught.
- `src/test/resources/org/eolang/lints/packs/single/unoptimizable-recursion/catches-dollar-self-call.yaml` — `$`-form recursion is caught.
- `src/test/resources/org/eolang/lints/packs/single/unoptimizable-recursion/allows-plain-name-tail-call.yaml` — plain-name recursion in a tail position stays allowed.

## Tests
- `mvn clean install -Pqulice` (671 tests) — pass

### DIFF
--- a/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
+++ b/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
@@ -52,6 +52,36 @@
   a reason of its own, the way the four below are.
   -->
   <!--
+  TRUE when the @base is a self-call of the object with the given name, in
+  any of the three forms the parser emits: the plain name resolves to
+  "Phi.<name>", the "^" form to "xi.rho.<name>", and the "$" form to
+  "xi.<name>". A dispatch on the self-call is also a self-call.
+  -->
+  <xsl:function name="eo:is-self-call" as="xs:boolean">
+    <xsl:param name="base" as="xs:string"/>
+    <xsl:param name="name" as="xs:string"/>
+    <xsl:sequence select="$base = concat('Φ.', $name) or starts-with($base, concat('Φ.', $name, '.')) or $base = concat('ξ.ρ.', $name) or starts-with($base, concat('ξ.ρ.', $name, '.')) or $base = concat('ξ.', $name) or starts-with($base, concat('ξ.', $name, '.'))"/>
+  </xsl:function>
+  <!--
+  TRUE when the @base is a dispatch on a self-call, i.e. the self-call is
+  the receiver and a method is dispatched on it.
+  -->
+  <xsl:function name="eo:is-self-dispatch" as="xs:boolean">
+    <xsl:param name="base" as="xs:string"/>
+    <xsl:param name="name" as="xs:string"/>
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@yegor256 
please review this pull request

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2102](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1413, opened 2026-09-06
fix(#1384): align rule texts with the conditions

### LINKED ISSUE #1384: Discrepancy between rule descriptions and implementations in `comment-without-dot`, `incorrect-version`, and `not-empty-atom` lints
## Problem

In three lints the defect text promises a rule that the XPath condition does not implement (or vice versa).

## 1. `comment-without-dot` — text allows `:`, condition does not

`src/main/resources/org/eolang/lints/comments/comment-without-dot.xsl:27`:

> "The comment for the object doesn't end with a dot (or '?', '!', ':')"

but the condition (`:13`) only checks `.`, `?`, `!`:

```xml
<xsl:for-each select="/object/comments/comment[not(ends-with(., '.')) and not(ends-with(., '?')) and not(ends-with(., '!'))]">
```

A comment ending with `:` is flagged although the message explicitly permits it.

## 2. `incorrect-version` — text promises strict SemVer, regex allows `-SNAPSHOT`

`src/main/resources/org/eolang/lints/metas/incorrect-version.xsl:16` (regex) vs `:32` (message "SemVer expected"). The regex additionally accepts `1.0-SNAPSHOT`, which is not a SemVer version — a user is told SemVer is required, but a non-SemVer value passes.

## 3. `not-empty-atom` — message lists void attributes the rule excludes

`src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl:31` builds the message from `o[@base and not(@name='λ')]` (which includes void attributes `@base='∅'`), while the selector (`:14`) excludes them (`@base!='∅'`). For an atom with both a void and a non-void attribute, the message names the void attribute as the violation.

## Expected behavior

Each lint's message and its condition must describe the same rule (align the condition with the text, or vice versa), and add packs for the boundary cases.

## Related

- `funcs/escape.xsl` — used by these lints to quote names in messages.

VasilevNStas:
Fixes #1384.

## Problem
Three lints had a discrepancy between their defect message and the XPath condition:

1. `comment-without-dot` — the message allows `:` at the end, the condition did not.
2. `incorrect-version` — the message promises strict SemVer, while the condition also accepts the Maven `X.Y-SNAPSHOT` convention (deliberately allowed by existing packs).
3. `not-empty-atom` — the message enumerated attributes that the selector excludes (void `@base='∅'` and `ξ`/`xi🌵`).

## Solution
Align each message with its condition:
- `comment-without-dot` — the condition now also accepts a trailing `:`.
- `incorrect-version` — the message (and the motive) now mention the Maven `X.Y-SNAPSHOT` convention alongside SemVer.
- `not-empty-atom` — the enumeration in the message now uses the same predicate as the selector.

## Changes
- `src/main/resources/org/eolang/lints/comments/comment-without-dot.xsl` — added `and not(ends-with(., ':'))`.
- `src/main/resources/org/eolang/lints/metas/incorrect-version.xsl` — new message text; `src/main/resources/org/eolang/motives/metas/incorrect-version.md` — motive updated.
- `src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl` — message enumeration matches the selector.
- New packs: `comment-without-dot/allows-comment-with-colon.yaml`, `not-empty-atom/catches-atom-with-void-and-real-attrs.yaml` (asserts the void attribute is not named).

## Tests
- `mvn clean install -Pqulice` (670 tests) — pass

### DIFF
--- a/src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl
+++ b/src/main/resources/org/eolang/lints/atoms/not-empty-atom.xsl
@@ -28,7 +28,7 @@
           <xsl:text>The atom </xsl:text>
           <xsl:value-of select="eo:escape(@name)"/>
           <xsl:text> may not have any attributes, which however exist: </xsl:text>
-          <xsl:for-each select="o[@base and not(@name='λ')]">
+          <xsl:for-each select="o[@base and @base!='∅' and not(@name='λ') and not(@base='ξ' and @name='xi🌵')]">
             <xsl:if test="position() &gt; 1">
               <xsl:text>, </xsl:text>
             </xsl:if>
--- a/src/main/resources/org/eolang/lints/comments/comment-without-dot.xsl
+++ b/src/main/resources/org/eolang/lints/comments/comment-without-dot.xsl
@@ -10,7 +10,7 @@
   <xsl:template match="/">
     <xsl:variable name="min" select="32"/>
     <defects>
-      <xsl:for-each select="/object/comments/comment[not(ends-with(., '.')) and not(ends-with(., '?')) and not(ends-with(., '!'))]">
+      <xsl:for-each select="/object/comments/comment[not(ends-with(., '.')) and not(ends-with(., '?')) and not(ends-with(., '!')) and not(ends-with(., ':'))]">
         <xsl:element name="defect">
           <xsl:variable name="line" select="eo:lineno(@line)"/>
           <xsl:attribute name="line">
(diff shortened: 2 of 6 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2114](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1414, opened 2026-09-06
fix(#1385): report one defect with the filtered depth

### LINKED ISSUE #1385: `too-deep-object` lint tool reports multiple defects for single nesting chain and emits incorrect depth number
## Problem

`too-deep-object` reports several defects along a single nesting chain (a cascade) and prints a depth number that does not match the rule's threshold, confusing the user.

## Root cause

`src/main/resources/org/eolang/lints/design/too-deep-object.xsl:13` (selector) and `:28` (message):

- The selector excludes certain wrapper objects (`Φ.tuple`, `Φ.bytes`, `Φ.number`, `Φ.string`) from the depth count, while the message prints the raw `count(ancestor::o)` — so the number in the text (e.g. "17 levels") differs from the filtered depth that triggered the rule.
- Because of the counting difference, more than one node per chain can match the `not(o[count(ancestor::o[...]) > 12])` guard, producing a cascade of defects for one chain (observed: defects at depths 13/14/15/17 for a 15-level chain).
- For data objects the message prints `eo:escape(@base)` which yields an empty/`∅` name.

## Minimal example

```eo
+package x
+architect yegor@256.com
+version 0.0.1

[] > main
  [] > a
    [] > b
      [] > c
        [] > d
          [] > e
            [] > f
              [] > g
                [] > h
                  [] > i
                    [] > j
                      [] > k
                        [] > l
                          [] > m
                            1 > @
```

## Expected behavior

- The depth number in the message must use the same (filtered) metric as the selector.
- Only the deepest offending node of a branch should be reported.
- Use `@name` (or a clear label) instead of `@base` for formations.

## Related

- #155 (the lint itself, merged); the cascade/off-by-N behavior is a new regression on top of it.
- `design/` category siblings (`broad-scope`, `excessive-visibility`).

VasilevNStas:
Fixes #1385.

## Problem
`too-deep-object` reported several defects along a single nesting chain (a cascade), printed a raw `count(ancestor::o)` depth that could exceed the rule's threshold (which excludes `Φ.tuple`/`Φ.bytes`/`Φ.number`/`Φ.string` wrappers), and named objects by `@base` (empty for formations, `ξ.φ` for data).

## Solution
- Extracted the filtered depth into `eo:depth()` and use it in both the selector and the message, so the reported number matches the threshold metric.
- The selector now excludes a node whenever **any** descendant (`.//o`, not just a direct child) is deeper than the threshold — one report per branch, no cascade.
- The object label uses `@name` when present, otherwise the last segment of `@base` (a clear short name like `m` instead of `Φ.m` or `ξ.φ`).

## Changes
- `src/main/resources/org/eolang/lints/design/too-deep-object.xsl` — new `eo:depth()` function, descendant-based guard, filtered depth and name-based label in the message.
- `src/test/resources/org/eolang/lints/packs/single/too-deep-object/catches-too-deep-object.yaml` — asserts now check a single defect and the exact message (`The object "m" is nested too deep: 13 levels...`).

## Tests
- `mvn clean install -Pqulice` (668 tests) — pass

### DIFF
--- a/src/main/resources/org/eolang/lints/design/too-deep-object.xsl
+++ b/src/main/resources/org/eolang/lints/design/too-deep-object.xsl
@@ -3,14 +3,24 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="too-deep-object">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0" id="too-deep-object">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/escape.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
+  <!--
+  Depth of the object in its nesting chain, counting only the ancestors that
+  represent real nesting: the wrapper objects ("Φ.tuple", "Φ.bytes",
+  "Φ.number", "Φ.string") do not count. The selector and the message use the
+  same metric.
+  -->
+  <xsl:function name="eo:depth" as="xs:integer">
+    <xsl:param name="o" as="element()"/>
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 
please review this PR

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2126](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1415, opened 2026-09-07
fix(#1369): drop the upper bound on the architect email TLD

### LINKED ISSUE #1369: `incorrect-architect` rejects any address whose TLD is longer than four characters
`incorrect-architect` caps the top-level domain of the address at four characters, so an address on any longer TLD is reported as malformed.

`src/main/resources/org/eolang/lints/metas/incorrect-architect.xsl:16`:

```xml
<xsl:if test="$meta-head='architect' and not(matches(upper-case($meta-tail),'^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$'))">
```

`[A-Z]{2,4}` is the whole story: a TLD of five characters or more never matches, and the meta is reported with

```
The format of the +architect meta is wrong: ... (an email expected instead)
```

Delegated TLDs longer than four characters have been ordinary for a decade — `.online`, `.software`, `.email`, `.company`, `.systems`, `.technology`, `.dev` is fine but `.developer` is not.

| `+architect` | verdict |
| --- | --- |
| `yegor256@gmail.com` | accepted |
| `yegor@example.info` | accepted |
| `yegor@example.online` | **reported** |
| `yegor@eolang.software` | **reported** |

Verified as two packs under `incorrect-architect`, each with `defects: 0`, both failing on master:

```
Tests run: 490, Failures: 2
```

while `good-architect.yaml` beside them keeps passing.

## Root cause

The pattern is one of the well-known "good enough" email regexes from the era when four characters was the longest TLD in use. Nothing else in the rule constrains the domain, so the length limit is doing all of the rejecting.

## Suggested fix

Drop the upper bound, which is the part that has gone stale:

```xml
'^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$'
```

That keeps every address the rule accepts today and stops refusing the ones above. If the intent is only to catch a value that is obviously not an address — a URL, a name, an empty tail — then requiring an `@` with something on each side of it says that much without a guess about the domain.

yegor256:
## Summary

- `incorrect-architect` rejected any `+architect` email whose top-level domain was longer than four characters (e.g. `.online`, `.software`, `.technology`), because the regex capped the TLD group at `{2,4}`.
- Dropped the upper bound (`{2,4}` → `{2,}`) in `src/main/resources/org/eolang/lints/metas/incorrect-architect.xsl`, so any TLD length is accepted, matching how the rule already behaves for two-to-four character TLDs.
- Extended `good-architect.yaml` with the addresses from the issue (`yegor@example.info`, `yegor@example.online`, `yegor@eolang.software`) to lock in the fix.

## Test plan

- [x] `mvn test -Dtest=org.eolang.lints.LtByXslTest` — 496 tests, 0 failures.

Closes #1369.

https://claude.ai/code/session_01DFQ7orNQq55Vcavky8UMfR

---
_Generated by [Claude Code](https://claude.ai/code/session_01DFQ7orNQq55Vcavky8UMfR)_

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/incorrect-architect.xsl
+++ b/src/main/resources/org/eolang/lints/metas/incorrect-architect.xsl
@@ -13,7 +13,7 @@
       <xsl:for-each select="/object/metas/meta">
         <xsl:variable name="meta-head" select="head"/>
         <xsl:variable name="meta-tail" select="tail"/>
-        <xsl:if test="$meta-head='architect' and not(matches(upper-case($meta-tail),'^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$'))">
+        <xsl:if test="$meta-head='architect' and not(matches(upper-case($meta-tail),'^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$'))">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
--- a/src/test/resources/org/eolang/lints/packs/single/incorrect-architect/good-architect.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/incorrect-architect/good-architect.yaml
@@ -6,6 +6,9 @@ sheets:
 defects: 0
 input: |
   +architect hello@gmail.com
+  +architect yegor@example.info
+  +architect yegor@example.online
+  +architect yegor@eolang.software
 
   [x] > foo
     42 > @


yegor256:

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2358](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1422, opened 2026-09-07
#1392: Do not report generated cactus names

### LINKED ISSUE #1392: `unoptimizable-recursion` falsely indicates autonamed identifiers as defects
`unoptimizable-recursion` names the offending formation with whatever is in `@name`, including the auto-generated cactus names the parser invents, so the defect points at an identifier the author never wrote and cannot search for.

Run over the parsed XMIR of `eo-runtime/src/main/eo/directory.eo` (`objectionary/eo` at `b747717325`):

```
directory.xmir:115  warning  experimental
The recursion in "a🌵100-6" cannot be turned into a loop, no self-call sits in a tail position
```

The name comes from the XMIR:

```xml
<o base="ξ.ρ.a🌵100-6"
   line="115"
   loc="Φ.directory.walk.a🌵99-4.a🌵100-6.earlier"
   name="earlier"
   pos="9">
```

`a🌵100-6` is what the parser generates for an object written with the `>>` auto-name suffix (§9.2) — in the source, line 115 of `directory.eo` is

```
        ^.stepped rest.tail > earlier
```

inside a `[^ accum prefix dir] >> walked` / `[^ rest] >> stepped` pair, neither of which has a name a reader could match against `a🌵100-6`.

`src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl:63` selects every named formation:

```xml
<xsl:for-each select="//o[eo:abstract(.) and @name and parent::o and o[@name='φ']]">
```

and line 82 prints the name as it stands:

```xml
<xsl:text>The recursion in </xsl:text>
<xsl:value-of select="eo:escape(@name)"/>
```

The rest of the codebase treats a cactus name as something not to talk about — `funcs/special-name.xsl:10` exists for exactly this:

```xml
<xsl:sequence select="$name = 'φ' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
```

and `anemic-getter`, `compound-name`, `invalid-name-notation`, `non-kebab-name` and `unsorted-named-attributes` all call `eo:special()` before reporting.

## Suggested fix

Either skip the auto-named formations, if the advice cannot be acted on without a name:

```xml
<xsl:for-each select="//o[eo:abstract(.) and @name and not(eo:special(@name)) and parent::o and o[@name='φ']]">
```

or keep reporting them and say where they are instead of what they are called — the `@loc` of the same node (`Φ.directory.walk.a🌵99-4.a🌵100-6`) is not much better, but the enclosing named ancestor is:

```xml
<xsl:value-of select="eo:escape((ancestor::o[@name and not(eo:special(@name))])[last()]/@name)"/>
```

which would say "the recursion inside `walk`", a name the author can find.

The choice matters beyond this rule: `>>` is the idiomatic way to write a helper in the runtime, so any rule that reports on formations will meet cactus names.

gemshrine:
Fixes #1392.

`unoptimizable-recursion` reported every named recursive formation, including
the cactus names generated for anonymous `>>` formations. A message such as
`a🌵100-6` does not exist in the EO source and gives the author no searchable
name to act on. The rule's advice therefore pointed at an internal parser
identifier rather than a source-level formation.

The rule now excludes generated cactus names from this diagnostic. This is the
same boundary used by the naming lints: a generated identifier is an internal
parser detail, not a user-facing declaration. Named source formations keep the
existing recursion analysis and message unchanged.

A new fixture creates an auto-named recursive helper and asserts that the rule
reports no defect. The complete XSL fixture suite passes, including all
existing positive recursion diagnostics.

Verification:

- `mvn -q -Dtest=LtByXslTest test` — 514 tests, 0 failures, 0 errors;
- the new fixture is parsed and included in the full single-scope lint pack.

### DIFF
--- a/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
+++ b/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
@@ -90,7 +90,7 @@
   -->
   <xsl:template match="/">
     <defects>
-      <xsl:for-each select="//o[eo:abstract(.) and @name and o[@name='φ']]">
+      <xsl:for-each select="//o[eo:abstract(.) and @name and not(contains(@name, '🌵')) and o[@name='φ']]">
         <xsl:variable name="name" select="string(@name)"/>
         <xsl:variable name="root" select="o[@name='φ']"/>
         <xsl:variable name="calls" select=".//o[eo:is-self-call(string(@base), $name)]"/>
--- a/src/test/resources/org/eolang/lints/packs/single/unoptimizable-recursion/allows-auto-named-formation.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/unoptimizable-recursion/allows-auto-named-formation.yaml
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/design/unoptimizable-recursion.xsl
+defects: 0
+input: |
+  [] > app
+    [n] >>
+      if. > @
+        n.lt 1
+        0
+        ^.app (n.minus 1)


yegor256:
Thanks, merged! Skipping the cactus-named formations is the right cut, and the new pack pins it.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3994](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1423, opened 2026-09-07
#1391: Compare each meta with its nearest predecessor

### LINKED ISSUE #1391: `unsorted-metas` fails to report out-of-order meta if it's alphabetically above first meta
`unsorted-metas` compares every meta with the **first** meta of the file rather than with the one immediately above it, so a meta that is out of order with respect to its neighbour, but still above the first one alphabetically, is never reported.

`src/main/resources/org/eolang/lints/metas/unsorted-metas.xsl:13-17`:

```xml
<xsl:for-each select="/object/metas/meta">
  <xsl:variable name="meta-text" select="concat(head, ' ', tail)"/>
  <xsl:variable name="previous" select="(preceding-sibling::meta)[1]"/>
  <xsl:variable name="previous-text" select="concat($previous/head/text(), ' ', $previous/tail/text())"/>
  <xsl:if test="$meta-text &lt; $previous-text">
```

`preceding-sibling` is a reverse axis, so `preceding-sibling::meta[1]` is the *nearest* preceding meta — but the parentheses in `(preceding-sibling::meta)[1]` turn the axis result into a sequence in **document order** first, and `[1]` then picks its first item: the file's first meta. Every meta is therefore checked against the same one.

This file is reported clean:

```
+architect yegor256@gmail.com
+version 0.0.0
+home https://github.com/objectionary/eo

[] > foo
  42 > @
```

`home` sits after `version`, which is the exact thing the rule is for. Verified as a pack under `unsorted-metas` with `defects: 0`, which passes on master:

```
Tests run: 489, Failures: 0
```

The two `catches-*` packs beside it keep passing because in both of them the out-of-order meta is out of order against the *first* meta as well — `catches-metas-unsorted.yaml` has `+alias stdout …` followed by `+alias stdin …`, and `catches-many-unsorted-metas.yaml` starts with `+z`, above which everything else compares low. So the suite never exercises the case the parentheses break.

## Suggested fix

Drop the parentheses, so the reverse axis keeps its meaning:

```xml
<xsl:variable name="previous" select="preceding-sibling::meta[1]"/>
```

With that, the file above is reported at the `+home` line, and the two existing packs keep their current defects.

A pack for the neighbour case is worth adding in the same change — three metas where the third is out of order with the second but above the first — since that is exactly what the suite cannot see today.

gemshrine:
Fixes #1391.

`unsorted-metas` used `(preceding-sibling::meta)[1]`. Parenthesizing a reverse
axis orders its result in document order before `[1]`, so every meta was
compared with the file's first meta. A neighbor violation that remained above
the first one alphabetically was silently missed.

The selector now uses `preceding-sibling::meta[1]`, which preserves the reverse
axis and selects the immediately preceding meta. Existing cases that are out
of order against the first remain reported; only the missing neighbor case is
added.

The regression fixture has `architect`, `version`, then `home` metadata.
`home` is out of order with `version` but greater than `architect`, so the old
rule reported zero while the corrected rule reports one warning for `home`.

Verification:

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/unsorted-metas.xsl
+++ b/src/main/resources/org/eolang/lints/metas/unsorted-metas.xsl
@@ -12,7 +12,7 @@
     <defects>
       <xsl:for-each select="/object/metas/meta">
         <xsl:variable name="meta-text" select="concat(head, ' ', tail)"/>
-        <xsl:variable name="previous" select="(preceding-sibling::meta)[1]"/>
+        <xsl:variable name="previous" select="preceding-sibling::meta[1]"/>
         <xsl:variable name="previous-text" select="concat($previous/head/text(), ' ', $previous/tail/text())"/>
         <xsl:if test="$meta-text &lt; $previous-text">
           <xsl:element name="defect">
--- a/src/test/resources/org/eolang/lints/canonical.eo
+++ b/src/test/resources/org/eolang/lints/canonical.eo
@@ -3,11 +3,11 @@
 +architect yegor256@gmail.com
 +home https://www.eolang.org
 +package canonical
-+version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 +unlint missed-obfuscation
 +unlint unit-test-missing
++version 0.0.0
 
 [start] > canonical
   malloc.for > @
--- a/src/test/resources/org/eolang/lints/main-with-test.eo
+++ b/src/test/resources/org/eolang/lints/main-with-test.eo
@@ -3,10 +3,10 @@
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package f
-+version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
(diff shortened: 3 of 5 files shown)


gemshrine:
Dropping the parentheses made three `SourceTest` cases go red, and they were right to: `canonical.eo`, `main-with-test.eo` and `valid-source.eo` all put `+version` above `+spdx`, which is out of order. Nothing had ever compared those two, because every meta was measured against `+architect` at the top.

The order these fixtures now use matches what the transpiled sources in `objectionary/eo` carry: `architect`, `home`, `package`, `spdx`, `unlint`, `version`.

yegor256:
Thanks! Merged.

Nice catch on the parenthesised predicate: `(preceding-sibling::meta)[1]` re-sorts the reverse axis into document order and always yields the first meta of the file, so every meta was compared against the same one. Dropping the parentheses restores the immediately-preceding sibling, and the reordered fixtures confirm the rule was silently passing unsorted files before.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4006](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1424, opened 2026-09-07
#1368: Accept complete Semantic Version strings

### LINKED ISSUE #1368: `incorrect-version` rejects `1.0.0-alpha.1` and `1.0.0+build`, which SemVer defines
`incorrect-version` reports two version strings that SemVer defines, while telling the author that SemVer is what it wants.

`src/main/resources/org/eolang/lints/metas/incorrect-version.xsl:18`:

```xml
<xsl:if test="$meta-head='version' and not(matches($meta-tail, '^\d+\.\d+\.\d+(-[a-zA-Z0-9-]+)?$|^\d+\.\d+-SNAPSHOT$'))">
```

and the message it produces:

```
The format of the +version meta is wrong: <tail> (SemVer expected instead)
```

Two shapes that [SemVer 2.0.0](https://semver.org/) spells out are refused by that pattern:

| `+version` | SemVer | the rule |
| --- | --- | --- |
| `0.0.1-alpha` | valid | accepted |
| `0.0.1-beta-1` | valid | accepted |
| `1.0.0-alpha.1` | valid — §9, dot-separated pre-release identifiers | **reported** |
| `1.0.0+20130313144700` | valid — §10, build metadata | **reported** |

The pre-release group is `[a-zA-Z0-9-]+`, which has no dot in it, so a numbered pre-release such as `-alpha.1` or `-rc.2` — the form SemVer itself uses in its examples — falls through. Build metadata is not in the pattern at all, in either alternative.

Verified as two packs under `incorrect-version`, each with `defects: 0`, both failing on master:

```
Tests run: 490, Failures: 2
  incorrect-version/probe-semver.yaml
  incorrect-version/probe-build.yaml
```

while `allows-good-versions.yaml` beside them keeps passing, so it is these two shapes specifically.

## Root cause

The pattern was written from the common cases rather than from the grammar. The second alternative, `^\d+\.\d+-SNAPSHOT$`, is a deliberate Maven exception, so the intent is "SemVer, plus the Maven two-part snapshot" — and the first alternative falls short of the SemVer half.

## Suggested fix

The official SemVer regular expression, with the Maven alternative kept:

```xml
<xsl:if test="$meta-head='version' and not(matches($meta-tail, '^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$|^\d+\.\d+-SNAPSHOT$'))">
```

That also starts rejecting a leading zero in a numeric component (`01.2.3`), which SemVer forbids and the current pattern allows; if that is more strictness than wanted, `\d+` in the three places keeps today's behaviour.

Either way the two shapes above deserve packs, so the rule and the standard it names cannot drift apart again.

gemshrine:
Fixes #1368.

`incorrect-version` claimed to enforce SemVer but accepted only a single
hyphenated pre-release token and no build metadata. Consequently valid versions
such as `1.0.0-alpha.1` and `1.0.0+20130313144700` were reported as malformed.
The old expression also allowed numeric components with leading zeroes, which
SemVer explicitly disallows.

The rule now uses the complete SemVer 2.0 grammar for numeric components,
dot-separated pre-release identifiers, and dot-separated build identifiers.
The existing Maven `X.Y-SNAPSHOT` compatibility form remains accepted. Invalid
versions and their diagnostics are unchanged apart from the newly valid SemVer
forms.

The regression coverage adds pre-release, build-only, and combined SemVer
examples. The existing invalid-version fixture was updated so its former
`alpha.beta` case is treated as valid and all remaining defects retain their
line-specific assertions.

Verification:

### DIFF
--- a/src/main/resources/org/eolang/lints/metas/incorrect-version.xsl
+++ b/src/main/resources/org/eolang/lints/metas/incorrect-version.xsl
@@ -13,7 +13,7 @@
       <xsl:for-each select="/object/metas/meta">
         <xsl:variable name="meta-head" select="head"/>
         <xsl:variable name="meta-tail" select="tail"/>
-        <xsl:if test="$meta-head='version' and not(matches($meta-tail, '^\d+\.\d+\.\d+(-[a-zA-Z0-9-]+)?$|^\d+\.\d+-SNAPSHOT$'))">
+        <xsl:if test="$meta-head='version' and not(matches($meta-tail, '^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$|^\d+\.\d+-SNAPSHOT$'))">
           <xsl:element name="defect">
             <xsl:variable name="line" select="eo:lineno(@line)"/>
             <xsl:attribute name="line">
--- a/src/test/resources/org/eolang/lints/packs/single/incorrect-version/allows-semver-prerelease-and-build.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/incorrect-version/allows-semver-prerelease-and-build.yaml
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/metas/incorrect-version.xsl
+defects: 0
+input: |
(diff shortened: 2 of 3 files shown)


yegor256:
Thanks, merged! The official SemVer pattern is the right thing to match, and dropping `1.0.0-alpha.beta` from the negative fixture follows from it.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4018](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1425, opened 2026-09-07
#1393: Do not flag public numeric constants as methods

### LINKED ISSUE #1393: `excessive-visibility` lint incorrectly flags public constants in `win32`/`posix` as over-exposed
`excessive-visibility` decides that an attribute is over-exposed by looking only at the tests written inside the same object, so every attribute a *library* object publishes for other files is reported — with advice that would break the callers.

Run over the parsed XMIR of `eo-runtime/src/main/eo/win32.eo` (`objectionary/eo` at `b747717325`):

```
win32.xmir:22  warning  experimental
The method "append" is public, but no unit test refers to it; obfuscate it with >> instead of >
win32.xmir:23  warning  experimental
The method "creat" is public, but no unit test refers to it; obfuscate it with >> instead of >
```

and the same for `posix.xmir`.

Those two are not methods and they are not unused. `win32.eo:22-23`:

```
  8 > append
  256 > creat
```

they are the flag constants of the syscall wrapper, and the runtime reads them from other files:

```
eo-runtime/src/main/eo/directory.eo:70   win32.rdonly.plus win32.creat
eo-runtime/src/main/eo/file/touched.eo:76  win32.creat.plus win32.exclusive
eo-runtime/src/main/eo/posix.eo:104      posix.wronly.plus posix.append
```

Following the advice — renaming `> creat` to `>> creat` — gives the attribute a generated cactus name (§9.2), which is unreachable from outside the file, so all three call sites above stop compiling.

## Root cause

`src/main/resources/org/eolang/lints/design/excessive-visibility.xsl:22` looks for the reference inside the object's own tests:

```xml
<xsl:variable name="used" select="$tests//o[@base and matches(@base, concat('^(ξ(\.ρ)+\.|\.)', $method, '(\.|$)'))]"/>
<xsl:if test="empty($used)">
```

A `.xmir` holds one file, so "no test refers to it" is all this rule can see; a caller in another file leaves no trace in this document. The same blindness is reported for `redundant-object` in #1276 — there the references are `Φ`-rooted and simply not indexed; here the references are in another document altogether, which no indexing inside one file can fix.

The message compounds it by calling every attribute a "method": `8 > append` binds a number.

## Suggested fix

The check is sound for a *program* and unsound for a *library*, so it needs to know which it is looking at. Two ways:

- Report only attributes of an object that is not published — one with a `+package` meta is a library object by definition, and its direct attributes are its API; the rule could skip the top-level object's own attributes and keep checking the nested formations, where "used only by tests" really does mean over-exposed.
- Or gate the rule on a whole-program run: `lints` also has the `Programs` entry point that sees every XMIR of a project at once, and cross-file references are decidable there.

Whichever is chosen, the wording is worth fixing in the same change: "The method X is public" should be "The attribute X", since the rule matches every named child, and the advice `obfuscate it with >>` should not be given for anything reachable from outside the file.

gemshrine:
Fixes #1393.

`excessive-visibility` treated every named public child as a method. Numeric
constants such as `8 > append` and `256 > creat` in the `win32` and `posix`
runtime wrappers were therefore reported when their names were not repeated
in local tests. The suggested `>>` replacement would make those constants
unreachable, although other runtime files use them as part of syscall flags.

The selector now excludes attributes whose value is an EO numeric literal
(`Φ.number`). Such values are constants, not callable methods whose visibility
can be reduced based on local test references. Existing analysis of named
object and method attributes remains unchanged, including its current
test-scoped boundary.

A regression fixture defines a public numeric constant in an object with a
unit test and asserts that no excessive-visibility defect is emitted. The
fixture also keeps the constant unused locally, matching the runtime case that
was incorrectly reported.

Verification:

### DIFF
--- a/src/main/resources/org/eolang/lints/design/excessive-visibility.xsl
+++ b/src/main/resources/org/eolang/lints/design/excessive-visibility.xsl
@@ -15,7 +15,7 @@
       <xsl:for-each select="//o[eo:abstract(.) and @name]">
         <xsl:variable name="tests" select="o[eo:test-name(@name)]"/>
         <xsl:if test="exists($tests)">
-          <xsl:for-each select="o[@name and @base and @base != '∅' and @name != 'φ' and not(@local) and not(eo:test-name(@name))]">
+          <xsl:for-each select="o[@name and @base and @base != '∅' and @base != 'Φ.number' and @name != 'φ' and not(@local) and not(eo:test-name(@name))]">
             <xsl:variable name="method" select="@name"/>
             <!--
             A test refers to a sibling method either by calling it on
--- a/src/test/resources/org/eolang/lints/packs/single/excessive-visibility/allows-public-constant.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/excessive-visibility/allows-public-constant.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets:
+  - /org/eolang/lints/design/excessive-visibility.xsl
+defects: 0
+input: |
+  [] > foo
+    8 > append
+
+    [] +> checks-foo
+      42 > @


yegor256:
CI is green and this does remove the false positive from #1393, but I would rather someone else decide on the shape of the exclusion before it lands:

`@base != 'Φ.number'` names one base by hand. A public constant bound to `Φ.string`, `Φ.bytes` or a tuple is still reported, so the rule keeps the same defect for every non-numeric constant. If the intent is "a constant is not a method", the test probably belongs on whether the attribute is abstract at all, rather than on a single base.

Assigning to @yegor256 to pick between the narrow fix and the general one.

---

yegor256:
CI is green and the added pack passes, but I would rather not merge this, because it does not fix what #1393 diagnoses.

The issue's root cause is that the rule reads one `.xmir` and so cannot see a caller in another file; it proposes either skipping the top-level attributes of a `+package` object or gating the rule on a whole-program `Programs` run. This PR instead excludes `@base = 'Φ.number'`, which silences the two attributes in the report and leaves `> name "..."` or a `+> ` bool constant published for other files still flagged with the same breaking advice. The issue also asks for the wording (`The method "X" is public`) to be corrected in the same change, and that is untouched.

Nothing here is wrong; it is narrower than the defect. Assigning to @yegor256 to say whether the narrow relief is worth landing now or whether this should wait for the package/`Programs` fix.

---

PR #1427, opened 2026-09-08
fix(#1366): decode formatter specifiers instead of matching raw hex in sprintf-constant-args

### LINKED ISSUE #1366: `sprintf-constant-args` selects the removed `Φ.txt.sprintf`, so the rule never fires on current EO
`sprintf-constant-args` selects an object current EO does not have, so the rule can never fire on real code.

`src/main/resources/org/eolang/lints/misc/sprintf-constant-args.xsl:20`:

```xml
<xsl:for-each select="//o[@base='Φ.txt.sprintf'][count(o)=2][o[2][@base='Φ.tuple']]">
```

There is no `txt` package in `objectionary/eo` any more, and no `sprintf` object anywhere in the runtime. Formatting is `eo-runtime/src/main/eo/string/printf.eo`, reached as a method on the format string, so a parsed XMIR carries a dispatch:

```
(nothing)

$ grep -o 'base="[^"]*printf[^"]*"' eo-runtime/target/eo/1-parse/bytes/as-i16.xmir
base=".printf"
```

Every one of the eight packs feeds the rule `txt.sprintf`, so the suite stays green while nothing in a real program is ever examined.

The check itself is worth keeping: a format template that is a literal, with arguments that are all literal strings, has a result known at parse time and should be written as one string.

## Root cause

The shape of the call changed with the object. `"fmt".printf (* args)` puts the template on the **receiver**, which a real XMIR emits as the child without an `as` attribute, and leaves exactly one argument:

```xml
<o base=".printf">
  <o base="Φ.string">…the format…</o>
  <o as="α0" base="Φ.tuple">…the arguments…</o>
</o>
```

(checked against `eo-runtime/target/eo/1-parse/bytes/as-i16.xmir`)

so `count(o)=2` still holds, but `o[1]` is the template rather than `o[2]` being the tuple, and `@base` is `.printf`.

## Suggested fix

```xml
<xsl:for-each select="//o[@base='.printf'][o[@as][1][@base='Φ.tuple']]">
  <xsl:variable name="text" select="o[not(@as)][1][@base='Φ.string']/o[1][@base='Φ.bytes']/o/text()"/>
  ...
  <xsl:apply-templates select="o[@as][1]" mode="constant-args"/>
```

with the message naming `.printf`, and the packs rewritten to the current call form so they exercise something the parser can actually produce.

Counting the formatters needs the same treatment as in #1355: the byte pairs `2573`, `2564`, `2566`, `2578`, `2562` miss every specifier that carries a flag, a width or a precision, so `"%08.2f"` counts as zero formatters and the rule stays silent on a call it should examine.

Two sibling rules have the same dead selector and are reported separately, since each needs its own adaptation: #1355 is about arity and tuple shape, #1357 about the template's content, and this one about arguments that are all literals.

yegor256:
## Problem

Closes #1366.

`sprintf-constant-args` selects the `.printf` dispatch correctly, but its formatter count is computed by searching the hex dump of the format string for the *literal* hex byte sequence of a bare specifier (`2573` for `%s`, `2564` for `%d`, `2566` for `%f`, `2578` for `%x`, `2562` for `%b`). A specifier that carries a flag, width, or precision — e.g. `%08.2f` — inserts extra bytes between `%` and the conversion letter, so its hex sequence never matches and the lint silently misses the defect (0 formatters counted instead of 1).

## Fix

Decode the hex-encoded bytes back into the actual characters (the same `eo:hex-to-placeholder` approach already used by the sibling lint `sprintf-without-formatters.xsl`), then count specifiers with a regex that understands `%[N$][flags][width][.precision]conversion`, consistent with how `sprintf-without-formatters` already detects formatters.

## Testing

Added a pack, `catches-printf-with-flagged-format-and-constant-args.yaml`, using a real `"%08.2f".printf (* "3.14")` call with a constant string argument. It fails on `master` (0 defects reported instead of 1) and passes with this change. Ran the full `LtByXslTest` suite (517 tests) — all green, including the pre-existing `sprintf-constant-args` and `sprintf-without-formatters` packs.

https://claude.ai/code/session_01TFrxLNmS5phrWQhwndWUCU

---
_Generated by [Claude Code](https://claude.ai/code/session_01TFrxLNmS5phrWQhwndWUCU)_

### DIFF
--- a/src/main/resources/org/eolang/lints/misc/sprintf-constant-args.xsl
+++ b/src/main/resources/org/eolang/lints/misc/sprintf-constant-args.xsl
@@ -3,7 +3,7 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" version="2.0" id="sprintf-constant-args">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:math="http://www.w3.org/2005/xpath-functions/math" xmlns:eo="https://www.eolang.org" version="2.0" id="sprintf-constant-args">
   <xsl:import href="/org/eolang/funcs/lineno.xsl"/>
   <xsl:import href="/org/eolang/funcs/defect-context.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Reusing the `eo:hex-to-placeholder` approach from the sibling lint keeps the two consistent.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2338](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1404, opened 2026-09-06
fix(#1388): warn on invalid +unlint format

VasilevNStas:
Fixes #1388.

## Problem
Malformed `+unlint` values such as `+unlint ascii-only:abc` or `+unlint ascii-only:` (empty line part) are silently ignored: neither suppression nor a warning happens, so a typo quietly disables an intended suppression. `LtIncorrectUnlint` only validated the *name* part (up to `:`), not the format after it.

## Solution
`LtIncorrectUnlint` now also validates the format of the tail. When the lint name exists but the tail is neither the bare name, `name:N` nor `name:N-M`, it reports an error explaining the expected formats.

## Changes
- `src/main/java/org/eolang/lints/LtIncorrectUnlint.java` — filter and message extended to flag invalid formats; extracted private helpers (`tail`, `nameOf`, `invalid`, `defect`).
- `src/test/java/org/eolang/lints/LtIncorrectUnlintTest.java` — new tests: `catchesUnlintWithInvalidLineFormat`, `catchesUnlintWithEmptyLinePart`, `allowsUnlintWithRange`.
- `src/test/resources/org/eolang/lints/packs/single/incorrect-unlint/catches-invalid-format.yaml` — new YAML pack.

## Tests
- `mvn clean install -Pqulice` (672 tests) — pass
- `mvn clean test -Pdeep` (15 tests) — pass

Issue #1388, opened 2026-09-05
`+unlint` fails to warn when given invalid values, leading to silent suppression failures


VasilevNStas:
## Problem

Malformed `+unlint` values such as `+unlint foo:abc` or `+unlint foo:` (empty line part) are silently ignored: the lint neither suppresses anything nor warns about the bad format, so a user typo quietly disables an intended suppression.

## Root cause

`src/main/java/org/eolang/lints/LtUnlint.java:62-72` parses the `+unlint` tail into three buckets (`global`, `single`/`range` via regex, and `granular`):

```java
final boolean global = !unlint.matches(UNLINT_LINE_REGEX);
...
final boolean granular = ... starts-with(tail, name + ":") ...
```

A tail like `foo:abc` or `foo:` matches the `granular` predicate but none of the three regex branches, so it falls through with no effect. `LtIncorrectUnlint` validates only the *name* part (up to `:`), not the format after it.

## Minimal example

```eo
+package x
+architect yegor@256.com
+version 0.0.1
+home https://github.com/objectionary
+spdx SPDX-FileCopyrightText: Copyright (c) 2026 Objectionary.com
+spdx SPDX-License-Identifier: MIT
+unlint ascii-only:abc

[] > main
  # Привет
  42 > @
```

The `ascii-only` defect is NOT suppressed (fine), but the user is never told that `ascii-only:abc` is an invalid format — the suppression silently does nothing.

## Expected behavior

A `+unlint` with an invalid line/range part should produce a warning (either from `LtUnlint` or from `LtIncorrectUnlint`), so the user learns about the typo.

## Related

- `LtIncorrectUnlint.java` — currently validates only the lint name.
- The `+unlint` format is parsed independently in three places (`LtUnlint`, `DefectMissing`, `UnlintInRange`) — a shared parser would make validation consistent.

0crat:
[...]

PR #1435, opened 2026-09-10
test(#1021): cover incorrect-unlint with a real lint name and line suffix

### LINKED ISSUE #1021: many-void-attributes:37 suppression not recognized by incorrect-unlint
The suppression of the lint` many-void-attributes` using a line-specific directive like: `many-void-attributes:37` at [win32.eo](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/org/eolang/sys/win32.eo) is not recognized by the `incorrect-unlint` lint checker.

This causes an error during `eo-maven-plugin:lint` execution when running tests for dataize in eoc project by command `npm test -- test/commands/test_dataize.js`

At the step `dataizes: lang Java, version 0.56.2, tag 0.56.2` The following error occurs:

**[ERROR] org.eolang.sys.win32:10 Suppressing "many-void-attributes:37" does not make sense, because there is no lint with that name (incorrect-unlint (Single))**

The generated `win32.xmir` file contains meta:

```
      <meta line="10">
         <head>unlint</head>
         <tail>many-void-attributes:37</tail>
         <part>many-void-attributes:37</part>
      </meta>
```

The [LtIncorrectUnlint.java](https://github.com/objectionary/lints/blob/master/src/main/java/org/eolang/lints/LtIncorrectUnlint.java) contains this code for split tail by ":"

```
xml.path("/object/metas/meta[head='unlint']")
    .filter(
        u -> !this.names.contains(
            u.element("tail").text().orElse("unknown").split(":", -1)[0]
        )
    )
```

### Expected behaviour
The lint name should be parsed correctly, ignoring the line number suffix during validation. That is:
`many-void-attributes:37` → should be recognized as "many-void-attributes"

yegor256:
## Summary

`LtIncorrectUnlint` reports "Suppressing ... does not make sense" for
`+unlint many-void-attributes:37`-style suppressions of a *real* lint,
because it looks up the lint name after stripping the `:line` suffix.

Looking at `src/main/java/org/eolang/lints/LtIncorrectUnlint.java` on
`master`, this is already correct: `nameOf()` splits the tail on `:`
before checking membership in the known-lints set, and
`LtIncorrectUnlintTest#understandsUnlintsWithLineNumber` already covers
the shape generically. That fix landed as a side effect of PR #1404
(closing #1388), which rewrote this class, but #1021 itself was never
closed and had no coverage using an actual registered lint name run
through the production pack pipeline (the exact way the originally
reported `eo-maven-plugin:lint` failure on `win32.eo` exercised it).

This PR adds that missing integration-level regression pack, using the
real `many-void-attributes` lint name with a line suffix, asserting
zero defects — so this exact scenario can never silently regress again.

## Test plan

- [x] `mvn test -Dtest=LtByXslTest,LtIncorrectUnlintTest` passes locally, including the new pack `packs/single/incorrect-unlint/allows-real-lint-with-line-suffix.yaml`
- [ ] CI green

Closes #1021

https://claude.ai/code/session_011JuFSRNk5QvWp5dnwsqPyW

---
_Generated by [Claude Code](https://claude.ai/code/session_011JuFSRNk5QvWp5dnwsqPyW)_

### DIFF
--- a/src/test/resources/org/eolang/lints/packs/single/incorrect-unlint/allows-real-lint-with-line-suffix.yaml
+++ b/src/test/resources/org/eolang/lints/packs/single/incorrect-unlint/allows-real-lint-with-line-suffix.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+lint: incorrect-unlint
+asserts:
+  - /defects[count(defect)=0]
+input: |
+  +unlint many-void-attributes:37
+  +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+  +spdx SPDX-License-Identifier: MIT
+
+  [] > foo


zaguzovmaksim0-hue (approved):
I checked the current implementation and the regression scope. `LtIncorrectUnlint.nameOf()` already strips the `:line` suffix before membership validation, and `invalid()` explicitly accepts both `name:N` and `name:N-M`, so the reported `many-void-attributes:37` shape is valid on current `master`.

The value of this PR is therefore the production-pack regression: it uses the real registered lint name instead of a synthetic name and runs through the same pack machinery that originally exposed the problem. That is the right way to pin the now-correct behavior without changing implementation code.

All current CI checks are green across Linux, macOS, and Windows. No correctness issue found. Approving.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (12) hits-of-code; +4 to give you at least something.  Your running score is [+148](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1436, opened 2026-09-10
fix(#1434): teach `eo:test-name` the `p🌵`/`n🌵` test markers

### LINKED ISSUE #1434: `eo:test-name` function incorrectly identifies new test attribute names `p🌵` and `n🌵` in `funcs/test-name.xsl`
The `eo:test-name` function in `src/main/resources/org/eolang/funcs/test-name.xsl:15` decides that an attribute belongs to a unit test by reading `starts-with($name, '+') or starts-with($name, '-')`. Those two markers are on their way out: `+` and `-` are not legal φ-calculus attribute names, so objectionary/eo#8566 renames them, and the parser now emits a truthy test attribute as `p🌵name` and a throwing one as `n🌵name`. Against XMIR produced that way `eo:test-name` returns false for every test, and every consumer of the marker changes behaviour at once.

Two of them get loud. `unit-test-missing.xsl:15` counts zero tests in a file that is full of them and fires once per file, and `sparse-decoration.xsl:18` stops excluding test formations and flags each one whose only child is `φ`. CI on that branch reports 2094 warnings across 173 XMIR files in `eo-runtime` where master reports none: 167 from `unit-test-missing` and 1927 from `sparse-decoration`.

Four go quiet, which is the worse half, and they need more than a membership fix. `bad-test-name.xsl`, `wrong-test-order.xsl`, `unit-test-without-phi.xsl` and `incorrect-test-object-name.xsl` all strip the marker with `substring(@name, 2)`, which assumes it is one character wide. Saxon counts 🌵 as a single codepoint, so `substring('p🌵can-do-x', 2)` is `🌵can-do-x` and the `^(can|accepts)-` match at `bad-test-name.xsl:22` fails. Teaching `eo:test-name` the new markers on its own would therefore not restore `bad-test-name`, it would make it fire on every test in the repository instead of on none. Its polarity check on line 20 is `starts-with(@name, '+')`, which is false for `p🌵`, so each of those reports would also call a positive test negative and demand `cannot-`, `rejects-` or `stops-on-`.

`LtTestNotVerb` never imports the function at all. Its XPath is `/object//o[@name and starts-with(@name, '+')]` written inline, so `unit-test-is-not-verb` goes dark whatever `test-name.xsl` says, and nothing in the log shows that it did.

`non-kebab-name.xsl:15` and `unsorted-named-attributes.xsl:15` are the two that survive by accident. Both exclude test attributes and special ones, and `eo:special` at `funcs/special-name.xsl:10` matches any name containing the cactus, so the exclusion still happens, for the wrong reason but with the right result.

The fix is a marker-aware pair rather than a single predicate: `eo:test-name` accepting `p🌵` and `n🌵` alongside `+` and `-`, a companion function that strips whichever marker is present instead of slicing one character, the four call sites above moved onto it, `bad-test-name` reading its polarity from the marker rather than from `+`, and `LtTestNotVerb` going through the same function as everything else. Accepting both the old and the new markers lets the two repositories move independently, after which eo can bump to the new release and `+`/`-` can be dropped for good.

@yegor256

yegor256:
Closes #1434.

The EO parser stops prefixing a test attribute `@name` with `+` or `-`, since neither is a legal φ-calculus attribute name (objectionary/eo#8566). A truthy test now arrives as `p🌵name` and a throwing one as `n🌵name`.

### `funcs/test-name.xsl`

A marker-aware set of functions rather than a single predicate:

- `eo:test-marker` — the marker in front of the name, or an empty string. Accepts `p🌵` and `n🌵` alongside the old `+` and `-`, so `lints` and `eo` may move independently, after which the old pair can be dropped for good.
- `eo:test-name` — unchanged signature, now true for either spelling.
- `eo:test-title` — the name with whichever marker is present removed. `substring(@name, 2)` cannot do this: Saxon counts 🌵 as one codepoint, so slicing one character off `p🌵can-do-x` leaves `🌵can-do-x`.
- `eo:positive-test` — polarity read from the marker.

### Call sites moved onto them

| File | Was | Now |
| --- | --- | --- |
| `tests/bad-test-name.xsl` | `starts-with(@name, '+')`, `substring(@name, 2)` | `eo:positive-test`, `eo:test-title` |
| `tests/wrong-test-order.xsl` | `substring(@name, 2)` | `eo:test-title` |
| `tests/unit-test-without-phi.xsl` | `substring(@name, 2)` | `eo:test-title` |
| `misc/incorrect-test-object-name.xsl` | `substring(@name, 2)` | `eo:test-title` |

The lints that only ask `eo:test-name` — `unit-test-missing`, `sparse-decoration`, `test-word`, `test-with-comment`, `excessive-visibility`, `non-kebab-name`, `unsorted-named-attributes`, `invalid-name-notation` — follow the function and need no edit.

### `LtTestNotVerb`

Its XPath had `starts-with(@name, '+')` written inline and stripped the marker with `replace("+", "")`, so `unit-test-is-not-verb` went dark whatever `test-name.xsl` said. It now goes through `MarkedName`, the Java twin of the same function, which holds the marker list in one place.

### Tests

New YAML packs feed XMIR carrying the new markers straight through `document:`, since the parser on the classpath still emits the old ones: `bad-test-name` (4), `wrong-test-order` (2), `unit-test-without-phi`, `incorrect-test-object-name` (2), `unit-test-missing`, `sparse-decoration`, and a new `unit-test-is-not-verb` pack for the Java lint. The `catches-` ones assert the defect text, which pins the stripped title and the polarity.

`mvn test` — 710 pass. `mvn -Pqulice verify` and `xcop` are clean.

https://claude.ai/code/session_01XWJEebtyWwtj6cuN9AoRiq

---
_Generated by [Claude Code](https://claude.ai/code/session_01XWJEebtyWwtj6cuN9AoRiq)_

### DIFF
--- a/src/main/java/org/eolang/lints/LtTestNotVerb.java
+++ b/src/main/java/org/eolang/lints/LtTestNotVerb.java
@@ -48,7 +48,7 @@ final class LtTestNotVerb implements Lint {
     @Override
     public Collection<Defect> defects(final XML xmir) throws IOException {
         return new Xnav(xmir.inner())
-            .path("/object//o[@name and starts-with(@name, '+')]")
+            .path(MarkedName.positives())
             .filter(object -> !this.isVerb(object))
             .map(LtTestNotVerb::verbDefect)
             .collect(Collectors.toList());
@@ -65,9 +65,7 @@ final class LtTestNotVerb implements Lint {
     }
 
     private boolean isVerb(final Xnav object) {
-        return this.vocabulary.isVerb(
-            object.attribute("name").text().get().replace("+", "")
-        );
+        return this.vocabulary.isVerb(LtTestNotVerb.title(object));
     }
 
     private static Defect verbDefect(final Xnav object) {
@@ -77,8 +75,12 @@ final class LtTestNotVerb implements Lint {
             new LineOf(object).value(),
             String.format(
                 "Test object name: \"%s\" doesn't start with verb in singular form",
-                object.attribute("name").text().get().replace("+", "")
+                LtTestNotVerb.title(object)
             )
         );
+    }
+
+    private static String title(final Xnav object) {
+        return new MarkedName(object.attribute("name").text().get()).title();
     }
 }
--- a/src/main/java/org/eolang/lints/MarkedName.java
+++ b/src/main/java/org/eolang/lints/MarkedName.java
@@ -1,0 +1,87 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.lints;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * The {@code @name} of a unit test attribute, as the EO parser spells it.
+ *
+ * <p>The parser marks a truthy test attribute with the {@code p🌵} prefix
+ * and a throwing one with {@code n🌵}. Older parsers used {@code +} and
+ * {@code -} for the same purpose, which are not legal φ-calculus attribute
+ * names. Both spellings are accepted, so that {@code eo:lints} and
+ * {@code eo} may move independently.</p>
+ *
+ * <p>This is the Java twin of {@code /org/eolang/funcs/test-name.xsl}, for
+ * the lints that are not XSL stylesheets. Keep the two in sync.</p>
+ *
+ * @since 0.2.12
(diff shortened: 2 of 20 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (358 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2367](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1445, opened 2026-09-11
fix(#1426): name the named-attribute self-call shape

### LINKED ISSUE #1426: unoptimizable-recursion.xsl lacks specific naming for shapes
The puzzle `1321-e3532044` from #1321 has to be resolved: 

https://github.com/objectionary/lints/blob/53f3c3a25495d01e5c89798c498ca1c70569c98f/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl#L45-L132

The puzzle was created by @evgeny on 07-Sep-26. 

Estimate: 60 minutes,  role: DEV. 

yegor256:
## What

Closes #1426.

`unoptimizable-recursion` used to fall back to the generic reason "no
self-call sits in a tail position" for a self-call that is bound to a
named attribute (other than `φ`) and later read from outside the
formation — the exact shape the puzzle called out with the `range`
example. It now reports a specific reason for that shape:

```
The recursion in "range" cannot be turned into a loop, the self-call is
bound to the attribute "next", read from outside the formation instead
of returned
```

The remaining `@todo` comment is narrowed down to the two shapes this
lint still doesn't name specifically: a self-call reached as a
decoratee, and mutual recursion between two formations.

## Why

The compiler leaves a self-recursive formation as plain recursion
(instead of turning it into a Java loop) for several syntactic shapes,
and the lint is meant to name each one at the call site so the author
can rewrite the object. Before this change, the "named attribute"
shape fell through to a generic, uninformative message.

## Test plan

- Added `packs/single/unoptimizable-recursion/catches-a-self-call-bound-to-a-named-attribute.yaml`
  covering the new message.
- `mvn test` passes locally (full suite, no failures).

https://claude.ai/code/session_016hX2q1mH8xVFQPPRZV7Dch

---
_Generated by [Claude Code](https://claude.ai/code/session_016hX2q1mH8xVFQPPRZV7Dch)_

### DIFF
--- a/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
+++ b/src/main/resources/org/eolang/lints/design/unoptimizable-recursion.xsl
@@ -42,14 +42,12 @@
     </xsl:choose>
   </xsl:function>
   <!--
-  @todo #1321:60min Name the shapes this lint still misses.
-  The compiler also leaves the recursion alone when the self-call is
-  spelled "Phi.name" instead of "^.name", when the formation is reached
-  other than by a plain application, as a decoratee or as the receiver of
-  a dispatch, when an attribute holding a self-call is read from outside
-  the formation, as in "range", and when two formations call each other.
-  Each of those is syntactic, so each can be found here and reported with
-  a reason of its own, the way the four below are.
+  @todo #1321:60min Name the two shapes this lint still misses.
+  The compiler also leaves the recursion alone when the formation is
+  reached as a decoratee instead of by a plain application, and when two
+  formations call each other instead of one calling itself. Both are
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/eager-planck-wxuhpe`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1444`". Your running score is [+2643](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1448, opened 2026-09-11
#1441: Accept can-prefixed positive tests

### LINKED ISSUE #1441: Conflict between `unit-test-is-not-verb` and `bad-test-name` lints over `can-` prefix in positive test names
Two lints in the same jar ask for opposite things about a positive test name. `bad-test-name` in `src/main/resources/org/eolang/lints/tests/bad-test-name.xsl` requires it to match `^(can|accepts)-`, and `LtTestNotVerb` hands the same title to `Vocabulary.isVerb`, which tags it with OpenNLP and accepts it only when the first word is `VBZ`. `can` is a modal, tagged `MD`, so every name the first lint allows in its `can-` form the second one rejects.

The collision only surfaced in 0.4.9, since `MonoLints` in 0.4.8 held no reference to `LtTestNotVerb` and the check never ran. In `objectionary/eo` it now fires 1972 times against `eo-runtime`, which is every positive test in the module, and the only spelling that satisfies both lints is `accepts-`, which reads wrong for most of them: `accepts-negate-false` is not what the test says.

Either `unit-test-is-not-verb` should accept the modal that its sibling lint mandates, or the two should agree on one prefix set.

@yegor256

zaguzovmaksim0-hue:
Closes #1441.

Aligns `unit-test-is-not-verb` with the sibling `bad-test-name` rule. Positive test names beginning with `can-` are now accepted explicitly; all other positive names continue through the existing OpenNLP `VBZ` verb check.

This keeps `accepts-*` and ordinary singular-verb names working as before while removing the contradiction where `bad-test-name` requires/permits `can-*` but OpenNLP tags `can` as a modal (`MD`) and the Java lint rejects it.

The existing positive-name parameterized test now includes `can-read-bytes` as a regression case. `git diff --check` is clean.

The commit was created through GitHub and is cryptographically marked `Verified`.

### DIFF
--- a/src/main/java/org/eolang/lints/LtTestNotVerb.java
+++ b/src/main/java/org/eolang/lints/LtTestNotVerb.java
@@ -68,7 +68,8 @@ final class LtTestNotVerb implements Lint {
     }
 
     private boolean isVerb(final Xnav object) {
-        return this.vocabulary.isVerb(LtTestNotVerb.title(object));
+        final String title = LtTestNotVerb.title(object);
+        return title.startsWith("can-") || this.vocabulary.isVerb(title);
     }
 
     private static Defect verbDefect(final Xnav object) {
--- a/src/test/java/org/eolang/lints/LtTestNotVerbTest.java
+++ b/src/test/java/org/eolang/lints/LtTestNotVerbTest.java
@@ -59,7 +59,7 @@ final class LtTestNotVerbTest {
     @ExtendWith(MayBeSlow.class)
     @Execution(ExecutionMode.CONCURRENT)
     @ParameterizedTest
-    @ValueSource(strings = {"generates-report", "runs", "parses-dom"})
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks — this does resolve the collision #1441 describes, and it is the direction the issue proposes. One concern before merging: the `can-` prefix is now a string literal inside `LtTestNotVerb.isVerb`, while the set of allowed prefixes is really owned by `bad-test-name.xsl` (`^(can|accepts)-`) and by `Vocabulary`. Two places now have to be kept in step, which is how the two lints drifted apart in the first place. Teaching `Vocabulary` that a modal followed by a verb is acceptable, or reading the prefix set from one shared place, would keep them from diverging again.

Assigning to @yegor256 to decide which of the two resolutions the issue offers he wants.

---

PR #1449, opened 2026-09-11
#1440: Treat rho as a special name

### LINKED ISSUE #1440: `invalid-name-notation` lint excessively reports `ρ` character in receiver names
`eo:special` in `src/main/resources/org/eolang/funcs/special-name.xsl:10` exempts `φ`, the lambda, `ξ` and the cactus, but not `ρ`. `invalid-name-notation` filters candidates through that function, so every formation that writes its receiver down as `[^ index opened bits]` has the resulting `<o name="ρ"/>` measured against `[a-z]+(-[a-z]+)*` and reported. In `objectionary/eo` at `eo-runtime`, that is 629 of the lint's 764 reports, all of them noise. `$eo:rho` is already defined in the `_specials.xsl` this file imports through `_funcs.xsl`, so `or $name = $eo:rho` alongside the `$eo:xi` term is the whole fix.

The remaining 135 reports are a different question, and they are the reason the lint cannot be switched on in `eo-runtime` even once `ρ` is exempt. The motive says "no digits", and the standard library names its own numeric types with them: `u8`, `i64`, `as-i16` through `as-u64`, `win32` and `stat64` all fail the expression, and none of them can be renamed without renaming the language's types. Either the expression should admit a digit that does not lead the name, or the lint needs to stay off wherever such names are the domain.

@yegor256

zaguzovmaksim0-hue:
Closes #1440.

Adds the receiver name `ρ` to the shared `eo:special()` predicate used by `invalid-name-notation` and other name lints. `ρ` is a reserved structural name just like `φ`, `λ`, and `ξ`; applying the ordinary `[a-z]+(-[a-z]+)*` naming rule to it produces false positives for every explicit receiver.

The existing `invalid-name-notation/allows-special-names.yaml` fixture now includes a `ρ`-named object alongside the other special forms and still expects zero defects.

This intentionally addresses only the receiver false positives from #1440. The separate question of digits in domain names such as `u8`/`i64` is unchanged.

Validation before commit: XSL is well-formed XML and `git diff --check` is clean. The commit is GitHub-signed and `Verified`.

### DIFF
--- a/src/main/resources/org/eolang/funcs/special-name.xsl
+++ b/src/main/resources/org/eolang/funcs/special-name.xsl
@@ -7,6 +7,6 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:function name="eo:special" as="xs:boolean">
     <xsl:param name="name"/>
-    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or $name = $eo:xi or contains($name, $eo:cactoos)"/>
+    <xsl:sequence select="$name = '&#x3C6;' or $name = $eo:lambda or $name = $eo:xi or $name = $eo:rho or contains($name, $eo:cactoos)"/>
   </xsl:function>
 </xsl:stylesheet>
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Treating `ρ` the way `φ` and `ξ` are treated is right for all four lints that ask `eo:special`.

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (3) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+303](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1456, opened 2026-09-11
#1442: Handle overflowing syntax versions

### LINKED ISSUE #1442: `Version.parsed()` throws `NumberFormatException` for numeric component exceeding `Integer.MAX_VALUE`
## What happens

`Version.parsed()` first accepts arbitrary-length decimal components with `\d+`, then stores each component in an `int` with `Integer.parseInt()`:

```java
private static final Pattern CORE = Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)");
...
new Version(
    Integer.parseInt(matcher.group(1)),
    Integer.parseInt(matcher.group(2)),
    Integer.parseInt(matcher.group(3))
)
```

On current master `91ff81b80f0918d1e01846f289ead22dc58ad3e0`, an otherwise numeric version whose component exceeds `Integer.MAX_VALUE` passes the regex and escapes the parser as a bare `NumberFormatException`.

I compiled the current `Version.java` together with a same-package harness and got:

```text
1.2.3 -> Optional[1.2.3]
999999999999999999999999.2.3 -> THREW java.lang.NumberFormatException: For input string: "999999999999999999999999"
1.999999999999999999999999.3 -> THREW java.lang.NumberFormatException: For input string: "999999999999999999999999"
```

## Why this is a problem

`Version.parsed(String)` returns `Optional<Version>` and its current callers treat an unparseable/non-SemVer value as `Optional.empty()`. `LtSyntaxVersion`, for example, explicitly documents that non-SemVer values are ignored rather than crashing the lint run. An oversized numeric `+syntax` value is syntactically matched as a version core but instead aborts linting with an implementation-level integer conversion exception.

This makes the behavior depend on the magnitude of the digits rather than on whether the version is parseable under the lint's contract.

## What should happen

An overflowing numeric component should be handled as an unparseable version (return `Optional.empty()`), or `Version` should use a representation that can safely compare the accepted numeric range. In either case, malformed/oversized metadata should not leak a bare `NumberFormatException` out of `Version.parsed()`.

I searched the open issues for `Version parsed NumberFormatException`, `syntax-version overflow`, `huge syntax version`, `integer overflow version`, and related wording and found no duplicate.

zaguzovmaksim0-hue:
Closes #1442.

Replaces #1447 after rebasing the fix onto the current `master` and adapting the regression to the repository's newer YAML-pack test structure.

`Version.parsed()` now treats an overflowing numeric `major.minor.patch` component the same way it treats malformed/non-SemVer input: it returns `Optional.empty()` rather than leaking `NumberFormatException` out of linting.

A new `syntax-version` pack feeds `+syntax 999999999999999999999999.2.3` through the real lint and expects zero defects/no crash.

The commit was created through GitHub and is cryptographically marked `Verified`.

### DIFF
--- a/src/main/java/org/eolang/lints/Version.java
+++ b/src/main/java/org/eolang/lints/Version.java
@@ -57,13 +57,19 @@ final class Version {
         final Matcher matcher = Version.CORE.matcher(text.trim());
         final Optional<Version> result;
         if (matcher.find()) {
-            result = Optional.of(
-                new Version(
-                    Integer.parseInt(matcher.group(1)),
-                    Integer.parseInt(matcher.group(2)),
-                    Integer.parseInt(matcher.group(3))
-                )
-            );
+            Optional<Version> parsed;
+            try {
+                parsed = Optional.of(
+                    new Version(
+                        Integer.parseInt(matcher.group(1)),
+                        Integer.parseInt(matcher.group(2)),
+                        Integer.parseInt(matcher.group(3))
+                    )
+                );
+            } catch (final NumberFormatException ignored) {
+                parsed = Optional.empty();
+            }
+            result = parsed;
         } else {
             result = Optional.empty();
         }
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+327](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1462, opened 2026-09-14
fix(#1451): catch overflow when parsing numeric @line values

### LINKED ISSUE #1451: `LineOf` allows numeric `@line` values exceeding `int` range resulting in `NumberFormatException`
## What happens

The fix for non-numeric `@line` values made `LineOf.value()` check `line.matches("\\d+")` before parsing, but a numeric value can still exceed Java's `int` range and escape as the same bare `NumberFormatException`.

Current master `91ff81b80f0918d1e01846f289ead22dc58ad3e0`:

```java
final String line = this.element.attribute("line").text().orElse("0");
final int result;
if (line.matches("\\d+")) {
    result = Integer.parseInt(line);
} else {
    result = 0;
}
```

A minimal Java reproduction of that exact branch:

```text
7 -> 7
abc -> 0
999999999999999999999999 -> THREW java.lang.NumberFormatException: For input string: "999999999999999999999999"
```

## Why it matters

`LineOf` explicitly promises line 0 when the attribute is "missing or not an integer". A decimal string outside the representable `int` range is not a valid integer for this API, yet it passes the regex and terminates lint analysis with an implementation-level conversion exception.

This is distinct from #1387: that report covered non-numeric values such as `line="abc"`, and current master now handles those. The remaining failure is numeric overflow after the new regex gate.

Broken/hand-written XMIR is precisely where defensive line parsing matters; the linter should report defects rather than crash while trying to read their location.

## What should happen

Catch overflow and fall back to 0, or parse through a safe helper that returns 0 for any value that cannot be represented as a line number.

A regression in `LineOfTest` should cover a numeric string larger than `Integer.MAX_VALUE` and assert that `value()` returns 0 without throwing.

I searched issues for `LineOf overflow`, `LineOf large line`, `LineOf Integer.MAX_VALUE`, `numeric @line NumberFormatException`, and `huge line NumberFormatException`; no duplicate was found.

yegor256:
## What

`LineOf.value()` gated the raw `@line` attribute with `line.matches("\\d+")` before calling `Integer.parseInt(line)`, which correctly turned non-numeric values like `"abc"` into `0`. But a decimal string that matches `\d+` can still be wider than `Integer.MAX_VALUE` (e.g. `"999999999999999999999999"`), and `Integer.parseInt` still throws a bare `NumberFormatException` for those, crashing lint analysis instead of falling back to line `0` as documented.

## Fix

`LineOf.value()` now parses through a small private helper that wraps `Integer.parseInt` in a `try`/`catch`, returning `0` on overflow just like it already does for a missing or non-numeric attribute.

## Tests

Added `LineOfTest#returnsZeroWhenLineExceedsIntRange`, asserting that a `line` attribute far outside the `int` range returns `0` instead of throwing.

```
mvn test -Dtest=LineOfTest,LtByXslTest
Tests run: 555, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Also ran `mvn -Pqulice install -DskipTests` locally, which passes clean.

Closes #1451

https://claude.ai/code/session_01LXmkW1iAS7vgk51MWW4wdm

---
_Generated by [Claude Code](https://claude.ai/code/session_01LXmkW1iAS7vgk51MWW4wdm)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3256](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1464, opened 2026-09-14
test(#1463): validate document usage justification in packs

### LINKED ISSUE #1463: Invalid usage of `document:` in test packs over `input:` without explicit justification
I keep getting PRs from bots that add new test packs under `src/test/resources/org/eolang/lints/packs` using the raw `document:` field (hand-written XMIR) when a plain EO `input:` snippet would work just as well. Out of ~534 pack files, 160 already use `document:`, and that number should legitimately be small: raw XMIR is only needed when a test targets an invariant that a valid EO parse can never produce (e.g. `object-has-data`'s "catches" cases, which construct a data node under an object with the wrong `@base` on purpose). For anything else, `input:` is more readable and closer to what a real user would write.

I'd like a check (in the pack-loading test infrastructure, or as a separate validation step) that fails when a new or changed pack yaml uses `document:` without some explicit justification, so this stays exceptional instead of becoming the default.

Relevant classes to investigate: `XtLint`, `XtYaml`.

volodya-lombrozo:
## Description

Enforce explicit justification for raw XMIR usage in test packs by introducing validation that requires an `xml-reason` key whenever the `document` field is used instead of `input`. This prevents accidental proliferation of hand-written XMIR in test cases where standard EO input would be more appropriate and readable.

## Changes

- New `YamlPack` fixture class to parse and validate YAML test pack files
- New test `checksDocumentUsageIsJustified()` that ensures all packs using `document` include an `xml-reason` explanation
- Updated all affected test pack YAML files to include the required `xml-reason` key with TODO comments for future conversion to EO input

Fixes #1463


volodya-lombrozo:
@rultor merge

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (277 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+418](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1465, opened 2026-09-14
Ignore self-references inside unit tests

### LINKED ISSUE #1460: `unoptimizable-recursion` incorrectly reports non-recursive objects as recursive due to self-references in `++>` tests
**What happens**

`unoptimizable-recursion` counts a reference from a unit test as a recursive self-call, so an object that isn't recursive at all is reported for unoptimizable recursion. In the new test layout a `++>` test is an attribute of the object it tests, and naming that object inside the test is exactly what a test does.

Run over `eo-runtime/src/main/eo/clock.eo` (`objectionary/eo` at `8655f8755e`, lints at `de102d68`):

```
--- as it is
    DEFECT line 35: The recursion in "clock" cannot be turned into a loop, the self-call is the receiver of a dispatch
--- with its tests cut off
    no unoptimizable-recursion defect
```

Line 35 of `clock.eo` is a test, not a body:

```
  clock.as-bytes.size.eq 8 ++> can-be-eight-bytes-long
```

`clock` is a plain formation whose `φ` is an `as-i64` of a syscall. There is no recursion in it. Cutting the tests off the file and leaving the body untouched makes the defect go away, so the test is the whole reason for it.

The same shape shows up on other objects of the runtime whose tests mention them by name:

```
eo-runtime/src/main/eo/mktemp.eo:63  | mktemp.tmpfile.exists ++> can-create-a-temp-file
eo-runtime/src/main/eo/nop.eo:24     | nop 42 ++> accepts-a-number-argument
eo-runtime/src/main/eo/false.eo:19   | false.if
eo-runtime/src/main/eo/getenv.eo:26  | getenv "EO_UNSET_VARIABLE_6142" "first"
```

**Why**

The self-calls are collected over the whole subtree of the formation, tests included, while the tail-position search looks only inside `φ` (`design/unoptimizable-recursion.xsl`):

```xml
<xsl:variable name="root" select="o[@name='φ']"/>
<xsl:variable name="calls" select=".//o[eo:is-self-call(string(@base), $name)]"/>
<xsl:variable name="loops" select="($root | $root//o)[eo:is-self-tail(string(@base), $name) and eo:tail(., $root)]"/>
<xsl:if test="exists($calls) and empty($loops)">
```

A call written in a test therefore lands in `$calls` but can never land in `$loops`, and the pair "some calls, no loops" is the defect condition. Every neighbouring lint that must not look at tests uses `eo:test-name()` for it, for example `misc/sparse-decoration.xsl` and `design/excessive-visibility.xsl`; this one doesn't.

**How to fix**

Leave the test attributes out of the self-call search, the way the neighbours do:

```xml
<xsl:variable name="calls"
  select=".//o[eo:is-self-call(string(@base), $name)][not(ancestor-or-self::o[eo:test-name(@name)])]"/>
```

A pack for it: a formation with a non-recursive body and a `++>` test that names the formation must give `defects: 0`, while the same formation with a real self-call in its body must still be reported.

gemshrine:
Fixes #1460

The `unoptimizable-recursion` lint treated a self-reference inside a `++>` unit test as a recursive call in the object body. The reference is part of the assertion and does not participate in the production recursion, so this produced a false positive for ordinary objects such as `clock`.

The call selection now excludes self-call nodes below attributes recognized as unit tests. The existing handling of self-calls in the actual object body is unchanged.

A regression fixture covers a non-recursive object whose only self-reference appears in a `++>` test and expects no defect.

Verification:

- `mvn -q -Dtest=LtByXslTest#checksAllLints test` (535 cases, 0 failures)
- `git diff --check`


yegor256:
Merged, thanks! Self-references inside unit tests no longer read as unoptimizable recursion.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3496](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1466, opened 2026-09-14
Ignore the receiver in missed obfuscation

### LINKED ISSUE #1461: `missed-obfuscation` incorrectly counts receiver `ρ` as a void leading to spurious defect reports
**What happens**

`missed-obfuscation` counts the receiver `ρ` among the void attributes a formation is told to obfuscate. A formation whose only void is its own receiver gets a defect it cannot act on, because `ρ` is the receiver and there is nothing to rename it to.

On lints `de102d68` with eo-parser `0.63.0`:

```eo
+package foo
+version 0.0.0

[^] > thing
  ^.other > @
```

```
void: name=[ρ]   (U+03C1)
DEFECT line 4: The formation "thing" must obfuscate its void attributes before decorating another object
```

A formation with a real void gets the same message, which is the intended case:

```eo
[x] > thing
  x > @
```

and one with no voids at all is silent, as it should be.

Over `eo-runtime/src/main/eo` (`objectionary/eo` at `8655f8755e`) the rule reports **706** defects, and in **92** of them the only void of the formation is `ρ`.

**Why**

The selector takes any void whose name is not prefixed `cant-` (`design/missed-obfuscation.xsl`):

```xml
<xsl:for-each select="//o[eo:abstract(.) and @name and o[@name='φ'] and o[@base='∅' and @name and not(starts-with(@name, 'cant-'))]]">
```

Writing the receiver down as `[^]` emits `<o base="∅" name="ρ"/>`, which matches. The motive talks only about attributes a caller provides:

> A formation that decorates another object should obfuscate its void attributes. Otherwise, an attribute of the formation may shadow an attribute with the same name in the decoratee

`ρ` shadows nothing of the kind: it is the receiver every formation has, and `eo-printer` writes it as `^` rather than as a name a reader could confuse with an attribute of the decoratee.

This is the same shape as #1299, where `many-void-attributes` counts `ρ` against its cap of five. The fix there is the fix here.

**How to fix**

Leave the receiver out of the selector:

```xml
<xsl:for-each select="//o[eo:abstract(.) and @name and o[@name='φ'] and o[@base='∅' and @name and @name != 'ρ' and not(starts-with(@name, 'cant-'))]]">
```

A pack for it: `[^] > thing` with `^.other > @` must give `defects: 0`, while `[^ x] > thing` must still be reported for `x`.

gemshrine:
Fixes #1461

The `missed-obfuscation` lint treated the implicit receiver `ρ` as a user-defined void attribute. A formation that only had `ρ` and then decorated another object was therefore reported even though the receiver cannot be obfuscated or renamed.

The lint now excludes the receiver from the voids that require obfuscation. Other non-`cant-` void attributes continue to be checked exactly as before.

A regression fixture models a formation whose only void is `ρ` and verifies that no defect is emitted.

Verification:

- `mvn -q -Dtest=LtByXslTest#checksAllLints test` (536 cases, 0 failures)
- `git diff --check`


yegor256:
Merged, thanks! Excluding the receiver from the void check removes the false positives reported in #1461.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3508](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1470, opened 2026-09-15
#1446: name the whole character, not half of a surrogate pair

### LINKED ISSUE #1446: `ascii-only` defect message misrepresents supplementary Unicode characters
## What happens

`LtAsciiOnly` identifies the first forbidden comment character with `String.chars()`, then narrows the returned UTF-16 code unit to a single `char`:

```java
final Optional<Character> abusive = comment.text().get().chars()
    .filter(chr -> (chr < 32 && chr != '\n') || chr > 127)
    .mapToObj(chr -> (char) chr)
    .findFirst();
...
final Character chr = abusive.get();
...
String.format(
    "Only ASCII characters are allowed in comments, while \"%s\" is used ...",
    chr,
    ...
)
```

On current master `91ff81b80f0918d1e01846f289ead22dc58ad3e0`, a supplementary Unicode character such as `😀` is represented by a surrogate pair. The first surrogate is selected as the “character”, detached from its pair, and the diagnostic no longer contains the offending character.

A minimal reproduction of the exact current algorithm:

```java
String text = "hello 😀 world";
Character bad = text.chars()
    .filter(chr -> (chr < 32 && chr != '\n') || chr > 127)
    .mapToObj(chr -> (char) chr)
    .findFirst()
    .orElseThrow();
System.out.println(Integer.toHexString(bad));
System.out.println(bad);
System.out.println(Integer.toHexString(text.codePointAt(text.indexOf("😀"))));
```

Actual output on Java 21:

```text
bad-code-unit=d83d
bad-rendered=?
actual-codepoint=1f600
```

So a comment containing `😀` is reported as containing `?`/an unpaired surrogate rather than U+1F600.

## Why this is wrong

The lint message explicitly tells the user *which character* is forbidden. For BMP characters (Cyrillic/Chinese), the existing tests pass because one UTF-16 code unit equals one Unicode code point. Supplementary characters are different: `String.chars()` enumerates code units, not code points, and narrowing one surrogate to `char` destroys the original character.

This also makes the implementation inconsistent with its semantic task: the rule is about characters/code points, not Java UTF-16 storage units.

## What should happen

Use `String.codePoints()` and retain the offending value as an `int` code point. Render it with `Character.toChars(codePoint)` (or an equivalent code-point-safe conversion), and compute the diagnostic position from the code point's UTF-16 offset without splitting the surrogate pair.

A regression fixture with an emoji in the top comment should assert that the defect message contains the actual emoji rather than a replacement/unpaired-surrogate glyph.

I searched existing issues for `ascii-only emoji`, `ascii-only surrogate`, `LtAsciiOnly supplementary`, and `ASCII comment emoji position` and found no duplicate.

opusstudiohq-max:
Closes #1446.

`LtAsciiOnly` found the first forbidden character with `chars()`, which yields UTF-16 code units, and then narrowed it to a `char`. A character outside the Basic Multilingual Plane is a surrogate pair, so the message named the leading surrogate on its own - not the character the author wrote, and not a character at all.

The defect was still reported, which is why nothing caught this. Only the wording was wrong, and it was wrong for exactly the input a reader most needs named.

`codePoints()` yields whole characters, so the code point survives and is turned back into text with `Character.toChars`. The filter is untouched: it already compared numeric values, and a code point outside ASCII still fails it. `indexOf` now takes that string rather than a `char` - for the Basic Multilingual Plane the position is unchanged; for a surrogate pair it is the start of the pair instead of a match on a lone surrogate.

### Negative control

The pack was run against **unfixed** master first, on a throwaway PR on my own fork. It failed on ubuntu, macos and windows, and the failure is precisely the one the issue describes:

```
1 XPath expression(s) failed:
  OK:   /defects[count(defect)=1]
  FAIL: /defects/defect[contains(text(), '😀')]
```

The lint fires; the message does not carry the character. With the source change, the same pack passes on all three platforms.

So the test isolates the wording and not the detection - if someone narrows the code point back to a `char`, this pack says so.

---


yegor256:
Merged, thanks! Walking code points instead of chars reports supplementary characters whole.

---

0crat:
@opusstudiohq-max Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+24](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1458, opened 2026-09-13
`sparse-decoration` motive incorrectly lists `if > @` as acceptable, conflicts with lint and pack rules


gemshrine:
**What happens**

The motive of `sparse-decoration` shows this as a **correct** program:

```eo
[] > decorates-app
  if > @
    true
    5
    five
```

The lint flags it anyway. Running `new LtByXsl("misc/sparse-decoration").defects(...)` on the example as parsed by `EoSyntax` (lints `de102d68`, eo-parser 0.63.0):

```
defects=1: Sparse decoration is prohibited (line 1)
```

The motive's other two correct examples (`[free] > decorates-with-free-args` and the unit test) produce no defect, as documented.

**Why it is the motive that is wrong**

The lint's own packs say an application decorated through `@` is sparse. `sparse-application.yaml` expects a warning for

```eo
[] > test
  if. > @
    a
    42
    88
```

and `sparse-formation-with-if.yaml` does the same. The XSL checks exactly that: an abstract object with a single child named `φ` that isn't void and has no `^` reference (`misc/sparse-decoration.xsl`). Nothing in it looks at whether `@` is applied to arguments. So the example in `motives/misc/sparse-decoration.md` teaches the opposite of what the rule enforces, and a user who reads the motive after getting the warning finds their code listed as correct.

**How to fix**

Replace the first correct example in `sparse-decoration.md` with a program that really isn't sparse, for example one with a second attribute or a void parameter, like the existing `[free] > decorates-with-free-args`. Optionally, add a sentence that decorating through an application with arguments is still sparse, matching `sparse-application.yaml`.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+3482](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1476, opened 2026-09-16
fix(#1459): make motive examples parse and match their own lint

### LINKED ISSUE #1459: Pinned parser rejects EO examples causing Incorrect examples to not trigger lint
**What happens**

Many EO examples in `src/main/resources/org/eolang/motives/**/*.md` are not valid EO for the parser that lints depends on (eo-parser `0.63.0`, `home.version` in `pom.xml`). In 22 motives the **Incorrect** example doesn't even trigger the lint it documents, because the parser rejects it before the lint gets a chance to run.

I ran every fenced `eo` and `xml` block of every motive through `EoSyntax(...).parsed()` and then through its own `LtByXsl` (lints `de102d68`). There are 199 blocks, and 85 of them, in 41 motives, come back with parser errors:

| parser error | blocks |
|---|---|
| `a blank line must separate the top comment block from the rest of the file` | 47 |
| `comment is allowed only on top of the file, before metas` | 20 |
| `QQ is not a valid object name, root Q must be followed by a dot` | 6 |
| `unexpected content after name suffix` | 3 |
| `test attribute legal only as direct child of top-level object` | 2 |
| `missing blank line before a +> test attribute (R-6.5.3)` | 2 |
| `object inside formation must have a name` | 1 |
| `line shape not yet implemented in spec parser` | 1 |

Most of them are the old style of putting a `# Comment.` right above an object, `QQ.io.stdout`, and tests written as `[] +> name` without the blank line.

The Incorrect example yields **no** defect of its own lint in these motives:

`atoms/atom-without-rt`, `comments/comment-is-too-wide`, `comments/comment-not-capitalized`, `comments/comment-starts-with-article`, `comments/comment-too-short`, `comments/comment-without-dot`, `critical/named-object-abstract-nested`, `design/no-attribute-formation`, `design/too-deep-object`, `misc/anemic-getter`, `misc/incorrect-test-object-name`, `misc/redundant-object`, `misc/sparse-seq`, `misc/sprintf-constant-args`, `misc/sprintf-without-formatters`, `misc/wrong-sprintf-arguments`, `names/anonymous-formation`, `names/compound-name`, `names/invalid-name-notation`, `names/phi-is-not-first`, `tests/test-word`, `tests/unit-test-without-phi`.

In `atoms/rt-without-atoms`, `metas/incorrect-jvm-rt-location`, `metas/incorrect-node-rt-location` and `tests/unit-test-missing`, it's the other way round: the **Correct** example doesn't parse, and its own lint reports a defect on it.

For example, `comments/comment-too-short.md`:

```eo
# This comment is short.
[] > foo
  42 > @
```

is rejected with `[2:0] a blank line must separate the top comment block from the rest of the file`, and `comment-too-short` stays silent on it.

**Why it matters**

The motive is what a user reads after getting a defect (`Lint.motive()`). An example that the compiler refuses, or one that the lint doesn't flag, doesn't explain the rule. It also hides real drift. For instance, the three `sprintf` motives still use `QQ.txt.sprintf`, which is the base #1371 says the modern parser never emits.

Nothing tests the motives. `LtByXslTest` and `PkMonoTest` only check that the file exists and isn't empty.

**How to fix**

1. Rewrite the examples in current EO syntax: comments only in the file header followed by a blank line, `Q.org.eolang...` or plain names instead of `QQ`, tests as `+>` attributes with exactly one blank line before them.
2. Add a test that walks every motive, parses each block under `Incorrect:` and `Correct:`, fails on parser errors, and checks that the lint reports at least one defect on the Incorrect blocks and none on the Correct ones. That's the check above. It took about 20 seconds for all 94 motives.

yegor256:
## Summary

Closes #1459.

Many EO examples in `src/main/resources/org/eolang/motives/**/*.md` no longer parse under the current `eo-parser` (0.63.0), or parse fine but don't actually trigger (or wrongly trigger) the lint they're supposed to demonstrate. A defect reported to a user links to one of these motives, so a broken example fails to explain the rule, or worse, tells the user their flagged code is "correct."

- Rewrote the broken examples in current EO syntax across 44 motive files: adding the blank line the modern grammar requires after a top-of-file comment, moving comments to appear only before metas (never after), dropping the legacy `QQ.` shorthand for plain unqualified names or `Q.org.eolang...`, adding the blank line required before `+>`/`->` test attributes, replacing the retired atom `?` marker with `/Q.type`, replacing the retired self-recursion `%` operator with `$`, and adding the `^.` prefix now required for references to an enclosing scope.
- Fixed a few examples that were simply wrong for their own lint: `mandatory-package`/`-version`/`-home`'s second "Incorrect" example didn't have zero of the meta (so it didn't trigger the "mandatory" check), `sparse-decoration`'s first "Correct" example was itself flagged (already tracked as #1458), `not-empty-atom`'s "Incorrect" atom had no non-void attribute besides `λ`, and `named-object-abstract-nested`'s "Incorrect" shape can't be produced by the EO parser at all (the parser hoists a nested named application argument up via `vars-float-up`), so it's now shown as XMIR only, matching how `atom-in-atom` already does this for a similarly unwritable shape.
- Added `Motives`/`Example` (`src/main/java/org/eolang/lints/`) and `MotivesTest`: for every lint, it parses `lint.motive()`'s `Incorrect:`/`Correct:` sections and asserts each example parses without errors and gets (or doesn't get) a defect from that lint, so this can't silently regress again.

## Test plan

- [x] `mvn test` (`MotivesTest` plus the full existing suite) passes
- [x] `mvn verify -Pqulice` (checkstyle/PMD/jtcop) passes clean
- [x] Manually diffed every changed motive to confirm the shown "Incorrect"/"Correct" examples still demonstrate what the prose claims

https://claude.ai/code/session_01BQN9EcUGELXyCETzVDsUfM

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQN9EcUGELXyCETzVDsUfM)_


yegor256:
The `mvn` matrix failed on `LtByXslTest.checksDocumentUsageIsJustified`, not because of this PR's diff: `missed-obfuscation/allows-rho-void.yaml` was added by #1466 (merged as be9cd3e2) without the `xml-reason` key that check requires, and master's own CI has been red on that commit and the next one since. I merged master into this branch and, rather than just adding an `xml-reason` string, converted that pack from a raw XMIR `document:` to real EO `input:` (`[^] > thing` / `foo > @` parses to the same shape and still gives `defects: 0`), which sidesteps the requirement entirely. Pushed as 4bc5f93.

---

PR #1487, opened 2026-09-18
test(#1475): convert void-attributes tests to EO input

### LINKED ISSUE #1475: Unjustified xml-reason test packs require conversion or legitimate justification
#1463 added a check (`LtByXslTest.checksDocumentUsageIsJustified`) that fails when a pack uses `document:` (raw XMIR) without an `xml-reason:` key explaining why. To land that check without blocking on a full audit, I put a standard placeholder, `xml-reason: "not yet converted to an EO input test"`, on the 156 packs I hadn't reviewed yet (only the 4 genuinely XML-only `object-has-data` packs got a real reason).

I need to go through those 156 packs one by one and, for each, either:
- rewrite it using `input:` (EO source) when the pack doesn't rely on an invariant a valid EO parse can never produce, or
- replace the placeholder with a real `xml-reason:` when raw XMIR is actually required.

The placeholder text is what makes the remaining work greppable: `grep -rl 'not yet converted to an EO input test' src/test/resources/org/eolang/lints/packs`.

Relevant classes to investigate: `LtByXslTest`, `YamlPack`.

volodya-lombrozo:
Convert test packs for `void-attributes-not-higher-than-other` rule from raw XMIR to EO input format, and provide legitimate justifications for those requiring XML-only representation.

Fixes #1475


volodya-lombrozo:
@rultor merge

